PR #174, opened 2023-06-16
Fix log in and sign up page UX issues

### LINKED ISSUE #151: Signup: email sign up form error doesn't load correct tab
To reproduce:

- Goto http://devthedocs.org/accounts/signup/
- Sign up with something that triggers a form error -- reuse an existing login/email, or mismatch passwords
- Click sign up
- You'll be brought back to the social account login tab

To fix:

- Use push state on the tabs. There is an example of this on the site-community pricing page I believe.

agjohnson:
This aims to resolve #151 by making the log in and sign up forms for email sign up target a URL that will cause the FUI tab module to make the email tab active on page load.

This adds a dependency for tab history, jquery-address, and a helper module for consolidating the tab module instantiation call from our semanticui data binding.

Technically, this extends the sign up form from the log in form to ensure the structure is the same between the two pages. A few fixes are in the log in form page, namely some work around the tab module usage so that we can link directly to the email log in and email sign up tabs when needed. This is required to load the correct tab page on form errors, and we instruct forms to hit the form URL with the tab history hash in the URL. This is the piece that makes the module detect the active tab page.

Closes #151

[Screencast from 2023-06-16 16-22-28.webm](https://github.com/readthedocs/ext-theme/assets/1140183/7488ce22-4b6b-4dde-9b64-b2f7f64fb57a)


humitos:
It would be good to add a small message like "We recommend signing up with a connected account instead of email" under the "Sign up with Email" tab, so users re-consider it before continuing with email 😄

agjohnson:
:+1: That's not hard, I'll add something to the log in and sign up pages

PR #178, opened 2023-06-20
Template: gold membership

humitos:
Close `div` properly. Otherwise, the footer looks weird.

### Before

Screenshot_2023-06-20_17-43-31

### After

Screenshot_2023-06-20_17-43-02


agjohnson (approved):
Good catch!

PR #179, opened 2023-06-22
Build additions and fixes for build detail page

### LINKED ISSUE #75: Add a rebuild this build button
We added a "Rebuild this build" button to the Build Detail's page. See https://github.com/readthedocs/readthedocs.org/pull/8227

Note: without looking I feel like I already implemented this before readthedocs.org did.

### LINKED ISSUE #145: Build details page: add button to cancel the build
Currently, it's not possible to cancel a running build.

agjohnson:
This adds a few missing features and fixes a couple UX and UI issues with
 the build detail page:

 - Adds build retry button for PR builds, ported from the production dashboard
 - Adds a build cancel button for active builds, also ported
 - Tunes the JS observables
  - Drop computed for static observables, to reduce dependency chain and put
    updates in one place
  - Use observables for can_cancel/can_retry, instead of abstracting the data
    binding logic into templates
 - Fixes a small display bug with loading icon
 - Adds a helper CSS for hiding elements

Fixes #75
Fixes #145

----

While running the cancel button will show:

After it disappears:

image

I can't easily give proof of the PR builder retry button, but it's in the same location.


src/js/build/detail.js
```diff
@@ -222,12 +222,23 @@ export class BuildDetailView {
...
+    this.can_cancel = ko.observable(false);
+    /** @observable {Boolean} Build can be retried */
+    this.can_retry = ko.observable(false);
+    /** @observable {Boolean} There was doc output in the build */
+    this.can_view_docs = ko.observable(false);
+
+    // Consolidate all of the observable updates that depend on build state
+    this.state.subscribe((state) => {
+      this.update_state(state);
+    });
```
humitos, line 241: Base multiple observable only in one attribute: interesting pattern 👍🏼

humitos (approved):

src/js/build/detail.js
agjohnson: This is effectively what computed observables do, but reversed just to make it clearer which observables are updated when `state` is updated.

PR #183, opened 2023-06-26
Notification about beta release

humitos:
I'd like to start promoting the new Beta dashboard. I think we are ready for it now. I've been using it during the last weeks and I haven't found big issues.

My idea here is to send a notification to all users. I wrote this code that we can use (after improving the copy):

```python
from django.utils.translation import gettext_lazy as_
from readthedocs.notifications.notification import SiteNotification
from messages_extends.constants import SUCCESS_PERSISTENT

class BetaDashboardNotification(SiteNotification):

    success_message = _(
        'We are launching our new <strong>Beta dashboard</strong>. '
        '<a href="https://beta.readthedocs.org/">Give it a try</a> and send us feedback.'  # noqa
    )
    success_level = SUCCESS_PERSISTENT


for u in User.objects.filter(profile__banned=False).iterator():
    BetaDashboardNotification(user=u, success=True).send()
```

This notification will be shown to users on the old dashboard. They can dismiss the notification by clicking on the link or in the cross:

Screenshot_2023-06-26_09-49-48

Besides, when they arrive to the new beta dashboard they will see another notification saying this instance is in beta and we appreciate them sending back feedback to us. This notification cannot be dismissed on purpose; that's why I put it on the HTML template directly.

Screenshot_2023-06-26_09-57-06

### DIFF
--- a/readthedocsext/theme/templates/base.html
+++ b/readthedocsext/theme/templates/base.html
@@ -49,6 +49,12 @@
 
     {% block content-wrapper %}
       <div class="ui very padded container">
+
+        {# Temporal notification while we are in Beta #}
+        <div class="ui message info">
+            This dashboard is currently in Beta. Please report any feedback you may have.
+        </div>
+
         {% block messages %}
           {% include "includes/utils/messages.html" %}
         {% endblock messages %}
--- a/readthedocsext/theme/templates/base.html
+++ b/readthedocsext/theme/templates/base.html
@@ -52,7 +52,9 @@
 
         {# Temporal notification while we are in Beta #}
         <div class="ui message info">
-            This dashboard is currently in Beta. Please report any feedback you may have.
+            This dashboard is currently in beta,
+            you can <a href="https://{{ PRODUCTION_DOMAIN }}">return to the legacy dashboard</a> if you encounter any problems.
+            Feel free to <a href="https://{{ PRODUCTION_DOMAIN }}/support">report any feedback</a> you may have.
         </div>
 
         {% block messages %}


agjohnson:
We talked about easing into notifications here, as we probably already wanted to gate a beta, and time frames in July will be tricky with time off.

Continuing to point users to the beta in issues is still a great idea. We'll pick up the pace on a beta when I'm back in July.

agjohnson (commented):
We definitely want this sort of UI element on the beta dashboard, especially as we start getting more general users trying it out.

But, it's probably going to be annoying to have this be a static block that can't be dismissed. This block will be on every page -- form confirmations, logout, etc.

If this was a dismissable notification I think that would be enough. Perhaps emit this on first log in to the beta instance?

humitos:
I opened this 8 months ago. I think we are more than ready to start promoting the beta templates. I would like to move forward with this idea, I don't see any blocker on the ext-theme side. What do you think?

agjohnson:
The issue I noted above originally is still a problem. As this is implemented, the block is not dismissable and will be on every page. The notification system we have implemented is a better option for this.

humitos:
Dismissing a notification is not implemented on the new dashboard yet, and it will also require https://github.com/readthedocs/ext-theme/issues/259 to do it properly. I'm fine adding this as a static block for now and improve it in the next iteration. My main goal here is to get more people using the new dashboard --it's objectively better than the old one, but nobody is taking advantage of it.

If we don't want to go this way, I'm also fine _not adding this notification in the ext-theme templates_; but I really want to add a notification in the old dashboard pointing people to the new one. We can use the new notification system there to attach a one-time and dismissable notification to all the users.

agjohnson:
Yeah, I think you are talking more about adding a link from the legacy dashboard here. A notification here is just an exit path back if things go wrong. We do need this and it should be easy to find.

I don't feel a static block at the top of all pages is the best approach though. It interferes with the design on all pages -- forms, 404 pages, etc. There are lots of downstream places this doesn't fit.

I would suggest a more limited approach in both cases, like only on the main dashboard template. We already discussed starting with the build detail page, which is a step before this all.

humitos:
OK. I'm closing this PR since we are not going to move in the static block direction --which is what this PR was about.

> I would suggest a more limited approach in both cases, like only on the main dashboard template

We can start there, then 👍🏼 . This can be achieved by creating a one-time-dismissable notification as I mentioned before.

PR #188, opened 2023-07-11
Add unified build list and build detail page title for builds

### LINKED ISSUE #158: Builds: list filtering and PR builds bugs
We can support filtering by active versions, it would be great to be able
to also support filtering by PR build. This is currently possible, with a
URL like:

https://beta.readthedocs.org/projects/mitiq/builds/?version=1663

So, there are several pieces here:

- [x] There is a partially working filter already, but it doesn't seem to work. Fixed in https://github.com/readthedocs/readthedocs.org/pull/10528
- [x] The build listing doesn't show PR builds different from version builds, it could be better UX
- [x] There could be a link to the filtered build list view from the build detail page
- [x] The pull request build status shown on the build detail page is not the same status shown on the listing page -- there is no mention of the version being a pull request version

We don't list these non-active versions on the version listing page, so there isn't another entry point to the filtered build listing view besides a direct link from the build detail view -- for now at least.

agjohnson:
This addresses a few issues from #158. The build name used on the build
listing and build detail page both differed prior, and didn't mention
the version name consistently.

- Build detail page just used the build pk
- Build listing page listed all versions as "Version {{ version_name }}",
  even pull request builds, which were "Version 1234" in the build
  listing

Additionally, this adds some hierarchy to the build detail page, showing
an additional layer of nesting for both the version (links to a filtered
version listing) and the build list (links to a filtered build listing).

Closes #158

Build listing:

image

Build detail:

image

UX:

[Screencast from 2023-07-11 12-43-30.webm](https://github.com/readthedocs/ext-theme/assets/1140183/0757c546-7d77-437b-bfcb-ab9addfccc7a)

### DIFF
--- a/readthedocsext/theme/templates/builds/includes/build_name.html
+++ b/readthedocsext/theme/templates/builds/includes/build_name.html
@@ -21,7 +21,7 @@
     <span class="section">
       {# When used a a page title, link the sections independently #}
       {% if is_page_title %}
-        <a href="{% url "project_version_list" build.version.project.slug %}?version={{ build.version.slug }}">
+        <a href="{% url "projects_detail" build.version.project.slug %}?version={{ build.version.slug }}">
       {% endif %}
 
       {% if build.is_external %}
--- a/readthedocsext/theme/templates/builds/includes/build_name.html
+++ b/readthedocsext/theme/templates/builds/includes/build_name.html
@@ -1,0 +1,63 @@
+{% comment %}
+
+  A common include for showing the build name in a heading or title 
+
+  This is a more complex heading than other elements. It uses a breadcrumb to
+  show the version the build was for, as well as the type of build in the case
+  that build is for a pull request version.
+
+  When used as a page title, on the build detail page, show some extra text and
+  nested breadcrumbs that we don't need to show in the build listing. This
+  gives some UI to link back to the build list or the version list. Normally,
+  the whole breadcrumb is linked as single link.
+
+{% endcomment %}
+
+{% load i18n %}
+
+{# When not a page title, treat the entire breadcrumb as a link to the build #}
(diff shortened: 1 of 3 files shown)


readthedocsext/theme/templates/builds/includes/build_name.html
```diff
@@ -0,0 +1,63 @@
...
+
+{% load i18n %}
+
+{# When not a page title, treat the entire breadcrumb as a link to the build #}
+{% if not is_page_title %}<a href="{{ build.get_absolute_url }}">{% endif %}
+  <span class="ui {% if is_page_title %}large{% endif %} breadcrumb">
+    <span class="section">
+      {# When used a a page title, link the sections independently #}
+      {% if is_page_title %}
+        <a href="{% url "project_version_list" build.version.project.slug %}?version={{ build.version.slug }}">
```
agjohnson, line 24: Note: this does not work yet for pull request versions as the version list filters these out currently. This will work for normal versions however.

readthedocsext/theme/templates/builds/includes/build_name.html
```diff
@@ -0,0 +1,63 @@
+{% comment %}
+
+  A common include for showing the build name in a heading or title
```
humitos, line 3: It would be good to include what are the parameters this includes expects in the docstring as well. It seems that it needs `build=` and `is_page_title=`

humitos (approved):

readthedocsext/theme/templates/builds/includes/build_name.html
agjohnson: Good call. I'm normally good about this, I think it's a great benefit to have docs on the templates

PR #213, opened 2023-09-13
Addons: allow users to opt-in into the beta addons

### LINKED ISSUE #212: Addons: allow users to opt-in into the new addons
We have deployed the new addons for projects using `build.commands` and it has been working fine. We haven't received any complain and I've already saw users jumping into `build.commands` _because_ they have the flyout there as well (see https://github.com/readthedocs/readthedocs.org/issues/10706)

Also, we've found a way to "migrate old projects/builds using the old integrations to the new one" dynamically using a Cloudflare Worker (see https://github.com/readthedocs/readthedocs-ops/pull/1402).

However, _we can't do this migration automatically for all the projects_ because it won't look exactly the same. We will be moving the flyout from the bottom left being integrated into the navbar, to the bottom right being fixed in a floating position. I created an issue in the theme to make this integration to look exactly the same at https://github.com/readthedocs/sphinx_rtd_theme/issues/1523 -- but it needs a non-trivial amount of work for that to define the exact addons API, work on JS/CSS, release a new version, wait for people to migrate to this version, etc.

We can follow a simpler path here: "allow users to opt-in into the new addons". When opt-in we can communicate these differences (look&feel changes and extra features) and they can disable if they don't like it. This will give users the features we already have in the new addons immediately to them, it will allow us to gradually roll this out and fix issues as we go and also give us feedback about that integration into the theme should be.

Related:
* https://github.com/readthedocs/ext-theme/issues/211
* https://github.com/readthedocs/readthedocs-ops/pull/1402
* https://github.com/readthedocs/sphinx_rtd_theme/issues/1523

humitos:
Show a small form to the user where they can enable/disable the new beta addons on their project.

Screenshot_2023-09-13_16-35-50

* Closes #212
* Requires https://github.com/readthedocs/readthedocs.org/pull/10733

### DIFF
--- a/readthedocsext/theme/templates/projects/addons_form.html
+++ b/readthedocsext/theme/templates/projects/addons_form.html
@@ -15,7 +15,7 @@
 {% block project_edit_content %}
   <div class="ui info message">
     {% blocktrans trimmed %}
-      Read the Docs addons are in beta. Give it a try now!
+      We are in beta! Give it a try now and share your feedback with us by <a href="https://github.com/readthedocs/addons" target="_blank">opening an issue</a>.
     {% endblocktrans %}
   </div>
 
--- a/readthedocsext/theme/templates/projects/addons_form.html
+++ b/readthedocsext/theme/templates/projects/addons_form.html
@@ -1,0 +1,30 @@
+{% extends "projects/project_edit_base.html" %}
+
+{% load i18n %}
+{% load crispy_forms_tags %}
+
+{% block title %}{{ project.name }} - {% trans "Addons" %}{% endblock %}
+
+{% block project_addons_active %}active{% endblock %}
+{% block project_edit_content_header %}{% trans "Addons" %}{% endblock %}
+
+{% block project_edit_sidebar_help_topics %}
+  {% include "includes/elements/link.html" with url="https://blog.readthedocs.com/addons-flyout-menu-beta/" text="Blog post: Addons flyout menu beta" is_external=True class="item" %}
+{% endblock project_edit_sidebar_help_topics %}
+
+{% block project_edit_content %}
+  <div class="ui info message">
+    {% blocktrans trimmed %}
+      Read the Docs addons are in beta. Give it a try now!
+    {% endblocktrans %}
+  </div>
+
+  <form class="ui form"
+      method="post"
+      action="{% url 'projects_addons' project_slug=project.slug %}">
+    {% csrf_token %}
+    {{ form|crispy }}
+
+    <input class="ui primary button" type="submit" value="{% trans "Save" %}">
+  </form>
+{% endblock %}
--- a/readthedocsext/theme/templates/projects/edit_base.html
+++ b/readthedocsext/theme/templates/projects/edit_base.html
@@ -17,6 +17,9 @@
         <a class="item {% block project_advanced_active %}{% endblock %}" href="{% url "projects_advanced" project.slug %}">
           {% trans "Advanced settings" %}
         </a>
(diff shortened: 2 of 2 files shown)


readthedocsext/theme/templates/projects/edit_base.html
```diff
@@ -17,6 +17,9 @@
         <a class="item {% block project_advanced_active %}{% endblock %}" href="{% url "projects_advanced" project.slug %}">
           {% trans "Advanced settings" %}
         </a>
+        <a class="item {% block project_addons_active %}{% endblock %}" href="{% url "projects_addons" project.slug %}">
+            {% trans "Addons" %} <i class="ui small grey fa-duotone fa-rocket"></i>
```
agjohnson, line 21: Instead of the icon here, I would try using a label -- I've wanted to add a marker for new/beta features this way. Menu elements will show the label floating to the right, you can see an example here:

https://fomantic-ui.com/collections/menu.html#vertical-menu

Untested, but:

```suggestion
            {% trans "Addons" %}
            <div class="ui basic primary label">
              {% trans "Beta" %}
            </div>
```

readthedocsext/theme/templates/projects/addons_form.html
```diff
@@ -0,0 +1,30 @@
...
+{% block project_edit_content_header %}{% trans "Addons" %}{% endblock %}
+
+{% block project_edit_sidebar_help_topics %}
+  {% include "includes/elements/link.html" with url="https://blog.readthedocs.com/addons-flyout-menu-beta/" text="Blog post: Addons flyout menu beta" is_external=True class="item" %}
+{% endblock project_edit_sidebar_help_topics %}
+
+{% block project_edit_content %}
+  <div class="ui info message">
+    {% blocktrans trimmed %}
+      Read the Docs addons are in beta. Give it a try now!
```
agjohnson, line 18: This might be an okay place to ask the user for feedback too, maybe even just noting that we want feedback. We could link somewhere like the issue tracker too though maybe? Just an idea either way though.

agjohnson (approved):
Looks great! I'd like to do something nicer with the menu, let me know if you want me to fiddle with that though.

readthedocsext/theme/templates/projects/edit_base.html
humitos: Beautiful!


Screenshot_2023-09-14_12-53-10


This is exactly what I wanted but didn't know how to do it. I originally tested 🆕 but that icon was not available so I put just a random icon to denote that _I wanted something to differentiate_ this entry.

readthedocsext/theme/templates/projects/addons_form.html
humitos: Yes, I added this note here as a placeholder for this but I wasn't sure what exactly to write. I updated it a little and added a link to the repository.

readthedocsext/theme/templates/projects/edit_base.html
agjohnson: Perfect, that looks great. I'll want to reuse this pattern for new admin tabs and new features too, so glad we have a reason to try it out.

readthedocsext/theme/templates/projects/addons_form.html
agjohnson: That works great, I don't think we need to worry too much about the copy here. Most likely, users will enable the beta and then not return to this form. But, if the user is returning to the form, it's probably to turn the feature off -- so a great time to get feedback from them.

agjohnson:
Updates look great! We can merge this once https://github.com/readthedocs/readthedocs.org/pull/10733 is merged in.

PR #214, opened 2023-09-21
Switch to web-test-runner

agjohnson:
Reusing some of what I went through with the addons testing work, I
tried to use web-test-runner here as well. Even though we aren't using
web components here yet, this pattern still works quite nicely for
testing some complex browser integration style tests.

This uses WTR, Rollup for bundling for tests, and I've created two test
cases: I translated old tests to Chai/WTR, and a new test that mixes
HTML and JS, using WTR HTML tests.


src/js/tests/plugins/knockout_init.test.html
agjohnson, line 1: @humitos This was the pattern I was looking forward to trying. It seems like a tidy way to write HTML heavy tests, without having to manually build up the DOM.

You might not find as much use for it in addons as the Lit HTML `fixture` looks fairly comparable.

package.json
```diff
@@ -3,6 +3,7 @@
   "version": "1.0.0",
   "description": "Read the Docs site theme",
   "main": "src/js/index.js",
+  "type": "commonjs",
```
humitos, line 6: Not sure what's the impact of this or even if what I'm about to ask makes sense 😅 , but isn't `commonjs` deprecated and `libraryUMD` should be used here? 🤔

src/js/gold/index.js
```diff
@@ -1,6 +1,6 @@
 import jquery from "jquery";
 import ko from "knockout";
-import { loadStripe } from "@stripe/stripe-js/pure";
+import { loadStripe } from "@stripe/stripe-js/dist/pure.esm.js";
```
humitos, line 3: Sounds weird that you are using `dist/` path here. I remember that I was doing this as a hack but I understand that this is not correct, right?

Also, the documentation uses it without `dist/`: https://www.npmjs.com/package/@stripe/stripe-js#usage

src/js/application/plugins.js
```diff
@@ -1,34 +1,35 @@
...
 
-window.jQuery = jquery;
-global.jQuery = jquery;
+// TODO this is only really needed while testing, it could be behind a
+// conditional
+import "./globals";
 
 // Required for FUI tab module
-require("jquery-address");
+import * as jqueryAddress from "jquery-address";
```
humitos, line 9: I'm happy we are using the same pattern than in addons here. I don't really understand the differences, but it's good to have consistency across them 😄

src/js/tests/application.test.js
```diff
@@ -1,44 +1,41 @@
...
-  expect(config.debug).toBeFalsy();
-  expect(window.__webpack_public_path).toBeUndefined();
-});
+  it("has registered view inside views as well", () => {
+    const app = new Application();
+    const view = new ApplicationView();
+    app.registry.attach(view);
+    expect(view.BuildDetailView).to.not.be.undefined;
+    expect(view.ProjectListView).to.not.be.undefined;
+  });
```
humitos, line 27: These two tests are the same.

src/js/tests/plugins/knockout_init.test.html
humitos: Interesting... So, this "runs a tests from inside an HTML file directly" loading the HTML into a browser and executing its JavaScript, right?

humitos (approved):
This looks good to me.

Note that I don't have all the knowledge required to fully understand this and I still have some doubts that I'm not even able to formulate questions from them 😓 . I understand the pattern and the big picture here, but I don't feel comfortable _writing_ actual test cases --or giving constructive feedback about "what could be changed to make it better".

It makes sense to me, tho. I suppose I would be able to come back and give better feedback once I start working with web-test-runner and writing more tests in the addons repository, tho.

We are still discovering/figuring out what's the best pattern in both repositories, so I'm happy to keep moving forward here and adapt things we find could be better in the future.

src/js/application/plugins.js
agjohnson: I originally struggled to find this solution in the past, so left this as a mixed es6/commonjs library. But, the answer was much easier than I was expecting apparently

src/js/tests/plugins/knockout_init.test.html
agjohnson: Yeah, I've found it's quite nice for some of the element and view tests in ext-theme. It might be something you find useful in writing some tests on addons, though I think the `html` fixture will do everything you need for most tests.

src/js/gold/index.js
agjohnson: It's actually recommended from their documentation for their ES module:
https://github.com/stripe/stripe-js#importing-loadstripe-without-side-effects

Stripe's JS actually has a problem very similar to the EA client. Their normal library loads a script from their website, automatically hooks into browser events on import and so is very difficult to test. I know the browser events are part of their fraud detection, but I feel okay skipping that amount of data collection.

So they did what I was suggesting with the EA client: one library automatically injects things at start up, and they make a module available (`dist/pure.esm.js`) that acts more like a ES module, and has no startup side effects.

package.json
agjohnson: This setting is indeed a bit confusing and it depends the most on what you need from your library. This was an attempt to fix an issue with web-test-runner, though in testing again, I don't actually think this change is needed. I'll follow up with a fix here too.

web-test-runner tries to push authors towards pure ES modules, so you would normally have `"type": "module"` here for a modern library. CommonJS is still a valid output type, though things are standardizing around ES modules still. UMD is a hack to support both at once, but UMD also did not yet supersede CommonJS.

Even more confusing, Webpack dictates our module output type, the package.json setting is more for Node internals and web-test-runner internals.

agjohnson:
There are two fixes I'll open up a follow up PR for:

- [ ] Package.json type = module
- [ ] Duplicated test case

I've chained a few PRs, so things will go quicker if merge history is straight.

src/js/gold/index.js
humitos: > It's actually recommended from their documentation for their ES module:
> [stripe/stripe-js#importing-loadstripe-without-side-effects](https://github.com/stripe/stripe-js#importing-loadstripe-without-side-effects)

The link you shared uses `import {loadStripe} from '@stripe/stripe-js/pure';`, not `import { loadStripe } from "@stripe/stripe-js/dist/pure.esm.js";`

humitos: > So they did what I was suggesting with the EA client: one library automatically injects things at start up, and they make a module available (`dist/pure.esm.js`) that acts more like a ES module, and has no startup side effects.

Cool 😎

agjohnson:
The package module type also helped push this towards Webpack 5, which was an outstanding upgrade already. Webpack 4 did not like the changes needed to switch to a full ES module, but the upgrade to Webpack 5 wasn't as bad as I thought it would be either.

PR is here: https://github.com/readthedocs/ext-theme/pull/220

PR #215, opened 2023-09-22
Add test for Knockout semanticui data binding

agjohnson:
This confirms that the custom binding for SUI modules works as expected
by inspecting the element and confirming it seems to have had our logic
and the jQuery SUI plugin logic applied to it.

This found a small bug with the binding code itself. I only tested
dropdown for now, I'm not sure if it's worth testing anything too deeply
yet.


src/js/application/index.js
```diff
@@ -50,21 +51,22 @@ export class Application {
...
-    console.debug("Loading site front end configuration from script tag");
-
-    const site_config_src = jquery("script#site-config").text() || "{}";
-    const site_config = JSON.parse(site_config_src);
-    if (site_config.webpack_public_path) {
-      __webpack_public_path__ = site_config.webpack_public_path;
+    if (this.config === undefined) {
+      console.debug("Loading site front end configuration from script tag");
+      const site_config_src = jquery("script#site-config").text() || "{}";
+      this.config = JSON.parse(site_config_src);
```
humitos, line 57: Should we handle exceptions here? What happens if `script#site-config` contains non-JSON text?

src/js/application/plugins.js
```diff
@@ -347,9 +347,9 @@ export const message = {
  */
 export const semanticui = {
   update: (element, value_accessor, all_bindings) => {
-    const value = ko.unwrap(value_accessor());
+    const binding_value = ko.unwrap(value_accessor());
     const jq_element = jquery(element);
-    for (const [key, value] of Object.entries(value)) {
+    for (const [key, value] of Object.entries(binding_value)) {
```
humitos, line 352: He, nice variable overridden finding 😉

src/js/tests/plugins/knockout_semanticui.test.html
```diff
@@ -0,0 +1,48 @@
...
+        <div class="item">Choice 1</div>
+        <div class="item">Choice 2</div>
+      </div>
+    </div>
+
+    <script type="module">
+      import { expect } from "@open-wc/testing";
+      import { runTests } from "@web/test-runner-mocha";
+      import { default as ko } from "knockout";
+      import { default as jquery } from "jquery";
```
humitos, line 16: I had to use the same `import { default as objectPath } from "object-path"` and I was confused about this. I don't really understand why it's called "default", I was expecting to use `* as objectPath` but that didn't work.

Is this because we are mixing CommonJS and UMD or something similar? Or like old and new javascript patterns? Using `default as module` is _the way_ to "modernize an old module"?

src/js/tests/plugins/knockout_semanticui.test.html
```diff
@@ -0,0 +1,48 @@
...
+
+      runTests(async () => {
+        describe("Knockout semanticui plugin", () => {
+          it("SUI module initialized successfully", () => {
+            const elem = jquery(".ui.dropdown");
+
+            // The HTML attribute we add. Note `data()` is a jQuery method that
+            // can convert HTML data attributes to camelCase attribute names.
+            const data = elem.data();
+            expect(data.semanticuiDropdown).to.be.true;
```
humitos, line 32: Are you talking `data()` converting this `data-bind="semanticui: { dropdown: {}}` attribute? In that case, why `semanticuiDropdown` would be `true` if it's `{}`? 🤔

src/js/tests/plugins/knockout_semanticui.test.html
```diff
@@ -0,0 +1,48 @@
...
+
+            // Inspect internal attributes added to the element by the FUI
+            // module plugin. For example, the module is attached to the element
+            // here:
+            // https://github.com/fomantic/Fomantic-UI/blob/b19851433cef2feba58b036cc19dcc72da176f1b/src/definitions/modules/dropdown.js#L143C26-L149
+            const moduleData = elem.data("module-dropdown");
+            expect(moduleData.can.show()).to.be.true;
+            expect(moduleData.is.active()).to.be.false;
+            expect(moduleData.is.disabled()).to.be.false;
+            expect(moduleData.is.empty()).to.be.false;
```
humitos, line 42: Wouldn't this test be testing Fomantic-UI library instead of our own thing?

humitos (approved):
I understand the pattern we are following here and it looks fine to me. However, I'm not 100% sure to understand the exact tests written. I'm confused a little about how SemanticUI works here and if we are testing our own code or FomanticUI code.

I'm approving so you can move forward with this, but I'd like to have some answers to my questions to clarify this and have more context about how things work here because I feel confused 🙃

src/js/tests/plugins/knockout_semanticui.test.html
agjohnson: The name `default` comes from the imported libraries exports, namely the `default export`:

```
export default const foo = 123;
```

This is the value that will be returned when no specific import was specified:

```
> import something from "library";
> something === 123;
True
```

In the test suite, Rollup is performing some translation on CommonJS modules to make them ES modules. Rollup does something special for the default exports, and I was having a hard time with this myself. I noticed there was a difference between `import *` and `import default` though I think that both should work for a properly written CommonJS library.

I think using `default` might actually be the best, but don't have strong proof for this. It looks nicer and is more explicit, though might not also work for all libraries in both Rollup and Webpack.

src/js/tests/plugins/knockout_semanticui.test.html
agjohnson: `data()` seems to have two modes, only one of which I've used.

The [jQuery.data() method](https://api.jquery.com/data/) can pull data attributes from the element (this is what the tests below do) and it can pull data from the element, stored as HTML `data-*` attributes (this is what I've used it for). The latter is what happens without passing any arguments.

From an element like this:

```html
<div data-some-var="foo"></div>
```

It will return:

```json
{"someVar": "foo"}
```

So, this actually isn't doing anything with the `data-bind` attribute at all. This test is testing a value that is added to the element after initialization:

```
<div data-semanticui-dropdown="true"></div>
```

This is used for some CSS selectors and to avoid duplicating instantiation.

src/js/application/index.js
agjohnson: This test would be a good addition to the general handling of the `jsonInit` plugin, I'll come back to add a test for this next as well. The site configuration only happens in one place in the templates, so should be safe, but it's using the same plugin underneath.

agjohnson:
I'll treat the test for jsonInit handling as a separate change, that's a great test to have.

- [ ] Add test for exception handling on bad JSON data to the jsonInit (and friends) bindings

agjohnson:
Test and error handling added here:

- https://github.com/readthedocs/ext-theme/pull/221

PR #217, opened 2023-09-22
Fix up pre-commit config to be useful, add djhtml

agjohnson:
This ports some of the changes from the addons/etc repo here to match:

- Check assets are current using a script
- Run prettier via pre-commit
- Run black via pre-commit

In addition:

- Run djhtml linter on templates to correct spacing
- Drops common submodule for now

I'd view the commits separately to reduce noise:

- cac1603 - Fix up pre-commit config to be useful
- 2be2336 - Run djhtml on all template files to correct spacing


humitos:
> Run djhtml linter on templates to correct spacing

This is great! I'm always unsure how to indent them! 👍🏼

humitos (approved):
Looks good to me! I'm happy we are moving forward with more linters in our code. That makes our development work less stressful 😉

agjohnson:
> This is great! I'm always unsure how to indent them! 👍🏼

Hah, honestly me too :upside_down_face: I usually just use what vim gives me, but it's way easier with something applying rules. There are many ways to indent HTML, it's definitely not as clear a science like Python is.

I've noticed djhtml is not perfect, and some syntax throws it way off, but it's better than some of the other tools I've tried. This one came recommended, I'm surprised I haven't come across it yet.

PR #218, opened 2023-09-26
Add invite listing to organization owner listing

agjohnson:
This also adds a invite form in a modal. I noted some areas where this 
pattern is not ideal, but it's okay for now I feel.

The tests are not currently passing, I might remove them.

Note, this is chained on 3 PRs I'm waiting to merge, so history is extensive
here. Changes start at 6e12bf0. If the UX seems reasonable, I'll continue on with the corresponding team listing changes.

[Screencast from 2023-09-25 17-28-19.webm](https://github.com/readthedocs/ext-theme/assets/1140183/a9de8983-0c18-4c3f-9e78-365ff3de5d5c)


humitos:
> This also adds a invite form in a modal. I noted some areas where this
> pattern is not ideal, but it's okay for now I feel.

Expanding on some background here would be helpful. Why you decided to use a modal on this particular case? Why not using the same pattern we are following on all the other places (e.g. environment variables, redirects, domains, etc) where "a new _thing_" is added? On  environment variable: we have list and create in separate views/pages:

* https://beta.readthedocs.org/dashboard/docs/environmentvariables/
* https://beta.readthedocs.org/dashboard/docs/environmentvariables/create/


> If the UX seems reasonable, I'll continue on with the corresponding team listing changes.

What I'm seeing in the video looks good but I'd prefer if we only have one pattern to add new things (see previous questions). If we are moving forward with the modal approach, I'd like to validate the form before closing the modal, tho.

I haven't used this UI/UX too much, honestly. I think that @stsewd is a better fit here for reviewing this since he was the one that implemented the full logic on invites and revokes. I suppose he will have more valuable feedback here than myself.

agjohnson:
> Why not using the same pattern we are following on all the other places

I've wanted to put this in a modal in other places too, but didn't spend time to work out the changes here. I'd probably reuse this modal pattern there too.

But I also have opened issues to stop using this pattern of the hybrid ListView/CreateView too. It creates problems in the UI and hides the form under important content, which can sometimes be enough content that the form is pushed out of view entirely. I lean towards _always_ creating CreateViews for every interface so we don't have these mixed patterns. I already don't like that most UIs have a "Add *" button in the top right of the listing, but we are breaking that on some views by inlining the CreateView form in the ListView page.

This view also has *two* listings, so the form would be even more crowding in the UI here, and would be that much more hidden when the owners and invite lists are long.

So, I don't think the modal is super important, but until we have CreateViews for all the missing UIs, it's better than showing the form inline to the rest of the ListView content.

> What I'm seeing in the video looks good but I'd prefer if we only have one pattern to add new things

Yeah, agree 100%. I think this should be using CreateView to separate the form from the ListView, and where that isn't yet possible, a modal at least keeps the UI appearing the same (there is always a button in the top right of the listing view that brings the user to a form to add the new instance).

> If we are moving forward with the modal approach, I'd like to validate the form before closing the modal, tho.

Validation is why I'd prefer to have standard CreateView instead. To validate forms before submitting, we'd need to get into API validation of forms, which I don't feel we benefit much from.

For now, I think the modal works fine, but I'd want to follow this up eventually by making actual CreateViews/urls/etc.

agjohnson:
I've opened up the following issue to continue discussion on UI/UX patterns here:

- https://github.com/readthedocs/readthedocs.org/issues/10777

If this change has some momentum, I won't worry about extending the modal pattern here at all. However, for now, I'm going to leave the pattern to start testing it, as I don't know how soon we'd make a change like adding new views for the models that are missing them. I do very much agree that we should have one pattern here though, and feel this is the most correct version of fixing our UX discrepancies. The modal is better than nothing, but dedicated views unifies the UI/UX the most.

agjohnson:
With the underlying PRs merged, I think this is ready to merge. I'm going to continue the modal pattern on the team member listing/create view, but won't use the modal pattern anywhere else yet. I think the issue above will be the best place to focus energy unifying patterns.

I'll put the team member view changes in a separate PR most likely, but if this is still in review at that point, I might merge them up here.

readthedocsext/theme/templates/organizations/partials/invitation_list.html
stsewd, line 1: Maybe name this `invitation_owners_list.html` or similar? We also have invitations for team members to organizations.

stsewd (approved):

readthedocsext/theme/templates/organizations/partials/invitation_list.html
agjohnson: Aye, I was planning on reusing this template there, which I think should be fine. If I need I'll separate the two though

PR #225, opened 2023-10-11
Fix subprojects list template

stsewd:
`subproject` is a Project object, not a relationship object.

Can be tested with https://beta.readthedocs.org/dashboard/stsewd-demo/subprojects/


agjohnson (approved):
Good catch! I had to look pretty deeply to see where this comes from (it's from the Project.superproject property defined from `projects/views/base.py`)

PR #231, opened 2023-11-16
Integrations: always show secret and show warning if secret is not set

stsewd:
Screenshots

github-manual-no-seret
github-manual
generic-int

ref:

- https://github.com/readthedocs/readthedocs.org/issues/10906
- https://github.com/readthedocs/readthedocs.org/commit/55511b40ddcd6414183759db58eccc2907a7a3a8


readthedocsext/theme/templates/projects/integration_generic_webhook_detail.html
stsewd, line 1: With https://github.com/readthedocs/readthedocs.org/issues/10904, we should be able to just use one template instead of having to override it for generic integrations.

readthedocsext/theme/templates/projects/integration_webhook_detail.html
```diff
@@ -41,31 +37,62 @@
...
-    <div class="ui segment">
-      {% url 'api_webhook' project_slug=project.slug integration_pk=integration.pk as webhook_url %}
-      {% with PUBLIC_API_URL|add:webhook_url as integration_url %}
-        <code>{{ integration_url }}</code>
-        <span class="ui compact icon basic button" data-clipboard-text="{{ integration_url }}" data-content="{% trans "Copied!" %}">
-          <i class="fa-duotone fa-copy icon"></i>
-        </span>
-      {% endwith %}
-    </div>
+  <div class="ui segment">
```
agjohnson, line 49: I think all of the webhook attributes can go into a single `.ui.segment`, they don't need to be in separate segments.

I'd extend what you did below and make the URL a `div.ui.list > .item` too, with a header "Webhook URL" or similar. This would be a good update as the previous copy/UI didn't really describe this well.

readthedocsext/theme/templates/projects/integration_webhook_detail.html
```diff
@@ -41,31 +37,62 @@
...
+        <i class="fa-duotone fa-copy icon"></i>
+      </span>
+    {% endwith %}
+  </div>
 
-    {% block integration_details %}{% endblock %}
+  {% block integration_details %}
+    {% if integration.secret %}
+      <div class="ui segment">
+        <ul class="ui list">
```
agjohnson, line 62: I might use a `div.ui.list` and `div.ui.list > div.item` here, so that the list display is plain, instead of a bullet list.

readthedocsext/theme/templates/projects/integration_webhook_detail.html
```diff
@@ -41,31 +37,62 @@
...
+            </div>
+          </li>
+        </ul>
+      </div>
+    {% else %}
+      <div class="ui warning message">
+          {% blocktrans trimmed %}
+          This integration does not have a secret,
+          the authenticity of the incoming webhook cannot be verified.
+          <strong>Click on "Resync webhook" to generate a secret.</strong>
```
agjohnson, line 83: This copy here might need a conditional check on whether the intergration can be resynced at all. If `integration.can_sync == False`, the block below showing the form/button won't be there.

agjohnson (approved):
This is looking great! Thanks for jumping in on the change here. I noted a few visual improvements, but what you have fits really well already.

readthedocsext/theme/templates/projects/integration_webhook_detail.html
stsewd: This is under the `has_sync` check, same as the button.

stsewd:
Updated screenshots

Screenshot 2023-11-22 at 11-14-52 strong translation _strong - Integrations - Read the Docs
Screenshot 2023-11-22 at 11-14-47 strong translation _strong - Integrations - Read the Docs
Screenshot 2023-11-22 at 11-14-39 strong translation _strong - Integrations - Read the Docs

agjohnson (approved):
The updates look great! :muscle:

PR #239, opened 2023-12-07
Add SSH key admin views

agjohnson:
Sort of requires a patch to the view code, to avoid passing the
templates and bad corporate profile objects.


agjohnson:
Also not too much important to review here, this was a pretty straight forward port of the ``keys_list.html`` and ``keys_detail.html`` templates.

stsewd (approved):

PR #240, opened 2023-12-07
Add a listener for html files to web pack

agjohnson:
This is super handy, it will reload dev server connected pages when you
save an HTML template. No more switching windows and reloading the tab!


agjohnson:
@humitos I don't know if this is useful outside this repo, but I always miss the web-dev-server reload while working away from JS and LESS sources. This is super handy for working on the application.

humitos:
On addons, I think the reload is automatic because all the files are served by web-dev-server and it uses a websocket to track changes and communicate with the browser. So, I don't think we need anything extra there --at least for now.

agjohnson:
Yup, this uses that same mechanism, but includes HTML files in the inotify/listener, essentially. I can't think of any auxiliary files that addons has where you might benefit from something like this though.

PR #243, opened 2023-12-12
Improve performance of table listing for large querysets

### LINKED ISSUE #209: Organizations: security log page doesn't load
I think this is specific to our organization, because of the number of log lines. But this UI does load on the production instance.

The URL I am having trouble with is:
https://beta.readthedocs.com/organizations/read-the-docs/security-log/

This produces a 502 response, timing out as the beta instance tries to load the page.

agjohnson:
Previously, a large queryset would cause a view to timeout, as the
conditional `if objects` ends up inspecting the entire queryset for
some version of truthiness. Instead, a more explicit check for both
empty arrays and empty querysets avoids this overhead.

- Closes #209

### DIFF
--- a/readthedocsext/theme/templates/includes/crud/table_list.html
+++ b/readthedocsext/theme/templates/includes/crud/table_list.html
@@ -101,10 +101,17 @@
   {% endblock top_menu %}
 
   {% comment %}
-    Testing for boolean here is not efficient, but we support both querysets and arrays here #}
-    {% if not objects.exists %}
+    We usually pass in a queryset here, but some views emit arrays of objects
+    instead, so we support these as well. Therefore, the check here needs to be
+    for an empty array or an empty queryset. Because Django templates aim to
+    be as difficult as possible to use, this conditional looks pretty silly.
+
+    This cannot be a simple `if not objects` as this performs poorly when
+    there are many objects returned in the queryset.
+
+    Order on the conditional matters here, `and` has higher precedence.
   {% endcomment %}
-  {% if not objects %}
+  {% if objects.exists is None and objects|length == 0 or objects.exists == False %}
     {% comment "rst" %}
 
       .. _api-template-list-placeholder:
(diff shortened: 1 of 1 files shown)


readthedocsext/theme/templates/includes/crud/table_list.html
```diff
@@ -101,10 +101,17 @@
...
   {% comment %}
-    Testing for boolean here is not efficient, but we support both querysets and arrays here #}
-    {% if not objects.exists %}
+    We usually pass in a queryset here, but some views emit arrays of objects
+    instead, so we support these as well. Therefore, the check here needs to be
+    for an empty array or an empty queryset. Because Django templates aim to
+    be as difficult as possible to use, this conditional looks pretty silly.
+
+    This cannot be a simple `if not objects` as this performs poorly when
+    there are many objects returned in the queryset.
```
humitos, line 110: Here, you can a a link to the Django docs if you want that explains this behavior a little better: https://github.com/readthedocs/ext-theme/issues/209#issuecomment-1853518101

readthedocsext/theme/templates/includes/crud/table_list.html
```diff
@@ -101,10 +101,17 @@
...
+
+    Order on the conditional matters here, `and` has higher precedence.
```
humitos, line 112: It took me a while to add imaginary parenthesis 😄 . This particular case is explained in Django docs at https://docs.djangoproject.com/en/5.0/ref/templates/builtins/#boolean-operators -- we could add this link here.

humitos (approved):
Sounds good! Left 2 comments to add links to do the docs only.

readthedocsext/theme/templates/includes/crud/table_list.html
agjohnson: Yeah, I'm constantly annoyed at Django templates. This one didn't make sense to me either

Issue #128, opened 2023-03-01
Page responses times


humitos:
I've been having _a feeling_ that some pages were behaving slow. In particular project dashboard and build listing. Today, I wanted to remove doubts about this and I checked these times on New Relic:

* project dashboard: **3.5s** (https://onenr.io/0Zw0Dl5v2wv)
* build listing: **1s** (https://onenr.io/0OQMEqYNvQG)

I haven't explored too much yet because I don't have good connection from the train, but I'm almost sure this is because we are doing many queries to the db to show all the data it's currently showing. With a simple look, I can tell they are the `builds_build` queries the ones that are consuming most of the time in these pages to render.

agjohnson:
Think I might have found a clue here. I set up pyinstrument on the beta dashboard in production and did some profiling. I was trying to resolve #209 but having trouble even getting the page to load.

But I gave a try to the project listing and saw this:

![image](https://github.com/readthedocs/ext-theme/assets/1140183/1c8c73e3-5696-444c-bdbb-1f5ca38c7780)

So that already seems like a lot of time for `is_admin`.

I commented out the conditional block calling `is_admin` (this is on `projects/partials/project_list.html`) in production and page load times look great!

So, this can be resolved through some different template logic, hopefully.

agjohnson:
Ayup. Same on build listing page too:

![image](https://github.com/readthedocs/ext-theme/assets/1140183/319d2aae-9b2f-4d4b-a14c-386693a8d22e)

humitos:
> @humitos are you familiar with the correct query to use here? I could expose a `is_project_admin:project` style filter, but I'm not so sure.

I posted my proposed solution in https://github.com/readthedocs/ext-theme/issues/49#issuecomment-1851617977. I think the correct pattern is keep using `is_admin:project` everywhere and add a small ~5s cache on that filter tag.

agjohnson:
Maybe I'm not understanding what you're suggesting correctly, but wouldn't a 5s cache invalidate before it would be useful?

The problem on at least the project listing page is that the `is_admin` conditional runs for each project in the listing page, not a single project. A cache on these checks would almost certainly invalidate before the next time the project listing page was loaded. Maybe a long cache timeout could work here, but I would start to worry about parity between our database and our cache.

The build listing page can likely just use the context variable approach, I'm guessing that view is duplicating the same check against a single version.

humitos:
[...]

Issue #49, opened 2021-11-17
Replace usage of `is_admin:project` filter in object lists


agjohnson:
I've updated this issue to be more specific. Originally, this was meant as a note from an previous change:

- https://github.com/readthedocs/readthedocs.org/pull/7277/

The original change was to update use of `request.user|is_admin:project` with a context variable `is_project_admin` -- which is the same call and return value, but cached instead of repeated in the template.

However, neither approach is usable when listing multiple projects or similar, where we need the original logic (`request.user|is_admin:project`). We can't use the context variable approach, as this still incurs the same performance issue as doing this in template.

We need another option for iterating over a list of objects and using `is_admin`.

agjohnson:
Note this is not usable in a listing view yet, where I have used `is_admin` the most so far. Noted in #128 but this code is responsible for most of the loading lag on some listing UIs.

humitos:
I think the `is_project_admin` is the wrong approach. It needs to be defined on the context for each view. Besides, it implies rewriting all the templates that are currently using `is_admin:project` to use the new variable from the context.

I'd say we should implement a ~5s cache on `is_admin` filter tag instead (https://github.com/readthedocs/readthedocs.org/blob/5a5bf41db92cca4d14c9657da2e2a6067a8b9b53/readthedocs/core/templatetags/privacy_tags.py#L13-L15). That doesn't need to rewrite anything and all the templates will be benefit immediately without any other change.

I mentioned something similar on the PR linked at (https://github.com/readthedocs/readthedocs.org/pull/7277#discussion_r452072549)

agjohnson:
Mentioned on #128, this approach doesn't quite solve the problem there. I'm not sure a cached admin check is the final solution we want either, but I do agree we should do something other than replace these checks with `is_project_admin` -- which also doesn't work to solve #128.

Ideally:

- Templates and views all use the same approach
- This approach performs well when iterating over multiple objects (project listing view), not just a single object (project detail view)

I'll repurpose this issue to be more clear that perhaps `is_project_admin` is not the best target, but we probably want to do something more than just caching.

agjohnson:
Removed assignment for now, it's looking like this needs to be backend change.

agjohnson:
Just noting that this is mostly to resolve:

#128

PR #244, opened 2023-12-12
Improve performance of build list view

agjohnson:
This consolidates all of the build list checks for `is_admin:project`
into a single template variable for now. This replicates what we're
doing at the view level in other views.

Ongoing discussion for a long term approach is at #49

- Refs #128


humitos (approved):
Sounds good as an intermediate solution until we fix it globally.

PR #252, opened 2023-12-16
Rework FUI integration for 2.9.3 support

agjohnson:
This version implemented a new theme loading layer, `overrides.less`,
which I was not able to convince Webpack to override. So instead, I
tuned up how we were injecting our themes. The result is a bit more
standard now I feel, and one step closer to using the FUI source files
directly.

A few changes that 2.9.3 brought in: slightly different font loading,
icon loading, and duotone is now supported natively if we want to switch
back at some point.

The majority of the changes are on the common theme, in the
sui-common-theme repo.


agjohnson:
This change is all very much internal fiddling, it's not important to review.

Issue #259, opened 2024-01-04
Notification: support new notifications through API


agjohnson:
Edit: in discussion below we narrowed the focus of this for right now. Currently, everything is template driven, but the build detail page needs to be API driven still. We'll punt on the rest of the items as there isn't a huge amount of value in API driven user/organization notifications yet.

- [x] Remove page reload to view errors on build detail view and render errors from the API response
- [x] Ensure cold storage builds show errors as well, these should come from cold storage API data. (This is assuming the modeling has these new errors too, I'm not sure if we touched on this one at the modeling level yet)

----

### Decisions

- Will project/organization/user notifications be rendered by templates? Is there a need to make these API driven?

### Work

- Create a common Notification listing include template
  - Include this template on project base template, organization base template, and profile base template
  - If API driven, pass the notification API URL to the relevant Knockout view. There will be distinct URLs for project notifications, organization notifications, and user notifications (?).
  - Iterate over Notification instances and render. Reuse and tune current messages template for this
- Create a Notification listing for build detail page
  - Most of the logic above is still applicable, but this will be rendered slightly differently and in a different location than other notifications
- We did talk about how to use the bell icon menu and what notifications might go there. I feel we still aren't settled here, and maybe need to try out some options like news messages and high priority messages.

cc @humitos, you might have some input on what is needed here

humitos:
> Will project/organization/user notifications be rendered by templates? Is there a need to make these API driven?

Regarding template rendering vs. API driven, there are a few things to consider IMO:

- 👍🏼 template rendering is _a lot easier_ to me
- 😐 API driven is useful if we want more dynamic UX (similar to build detail's page), but I think it's not required for the notifications in the other resources at this point
- 👍🏼 using API driven will give us parity with the build detail's page
- 👎🏼 API driven will require writing and maintaining extra endpoints only for this use case

With that in mind, I would start with template rendering because it's easier to implement. We can keep thinking about this and change the approach if we find a more dynamic UX is required.

We need to create the PATCH endpoint to mark them as read / dismissed either way (see https://github.com/readthedocs/readthedocs.org/issues/10983)

> Create a Notification listing for build detail page

This is already implemented and it's working.

* Endpoint lives at http://devthedocs.org/api/v3/projects/test-builds/builds/3357/notifications/
[...]

PR #261, opened 2024-01-11
Notifications: show attached to `User`, `Project` and `Organization`

### LINKED ISSUE #260: Add missing notification template blocks for user, project, organization notifications
From the deploy today, it seems these were not ported to the new dashboard.

humitos:
Render notifications attached to these objects.
This is the first pass of this work and there are some things we need to make decisions and improve.

- Where (what pages) these notifications should render?
- How "global" they should be considered?

Note we are rendering these notifications in the template for now, but in the future they will be rendered using the APIv3: #259

#### User

Screenshot_2024-01-11_11-01-33

#### Project

Screenshot_2024-01-11_11-01-40

#### Organization

Screenshot_2024-01-11_11-01-16

Closes #260

### DIFF
--- a/readthedocsext/theme/templates/organizations/partials/header.html
+++ b/readthedocsext/theme/templates/organizations/partials/header.html
@@ -23,7 +23,7 @@ is_collapsed (default: True)
   {# Render all the notifications attached to the organization.#}
   {% for notification in organization.notifications.all %}
     <div class="ui message">
-      <i class="fa-solid {{ notification.get_message.get_display_icon_classes }}"></i>
+      <i class="{{ notification.get_message.get_display_icon_classes }} icon"></i>
       <span>{{ notification.get_message.get_rendered_body|safe }}</span>
     </div>
   {% endfor %}
--- a/readthedocsext/theme/templates/organizations/partials/header.html
+++ b/readthedocsext/theme/templates/organizations/partials/header.html
@@ -126,4 +126,13 @@ is_collapsed (default: True)
     </div>
   {% endblock organization_header_navigation %}
 
+{# Render all the notifications attached to the organization.#}
+  {% for notification in organization.notifications.all %}
+    <div class="ui message">
+      <i class="fa-solid {{ notification.get_message.get_display_icon_classes }}"></i>
+      <span>{{ notification.get_message.get_rendered_body|safe }}</span>
+    </div>
+  {% endfor %}
+
+
 {% endblock organization_header %}
--- a/readthedocsext/theme/templates/organizations/partials/header.html
+++ b/readthedocsext/theme/templates/organizations/partials/header.html
@@ -20,6 +20,16 @@ is_collapsed (default: True)
 {% endcomment %}
 
 {% block organization_header %}
+  {# Render all the notifications attached to the organization.#}
+  {% for notification in organization.notifications.all %}
+    <div class="ui message">
+      <i class="fa-solid {{ notification.get_message.get_display_icon_classes }}"></i>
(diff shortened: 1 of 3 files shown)


humitos:
I'm showing _all the notifications_ attached to `Project` and `Organization`. However, the endpoint we are creating will only show these notifications to admins/owners. @agjohnson should I do the same here by returning a more specific queryset via the `get_context()` in the view?

readthedocsext/theme/templates/organizations/partials/header.html
```diff
@@ -126,4 +126,13 @@
     </div>
   {% endblock organization_header_navigation %}
 
+{# Render all the notifications attached to the organization.#}
+  {% for notification in organization.notifications.all %}
+    <div class="ui message">
+      <i class="fa-solid {{ notification.get_message.get_display_icon_classes }}"></i>
```
agjohnson, line 132: If we are still adding `fa-solid` in `get_display_icon_classes`, the class should probably be removed here. Otherwise, this is safe for now.

agjohnson (commented):
I'd say we should keep all of the notifications in the same location on the page for now, except for the special location in the build detail page. It will be a little more confusing if there are user and project notifications, as these would display at two different points in the page.

I would put the project/organization notifications at the top of the content page, above the project/organization header. I could see us revisiting this later though.

readthedocsext/theme/templates/organizations/partials/header.html
humitos: I think you told me to remove `fa-solid` as a default class from that method. We only have `fas` now: https://github.com/readthedocs/readthedocs.org/blob/919d10aca54650c4a9af5f597f8a0f08646d3206/readthedocs/notifications/messages.py#L43-L46

agjohnson: Yeah, so for some context, `fas` is the shorthand version of `fa-solid` (and `fad` is the shorthand of `fa-duotone`). I think what I noted was you were always prepending `fa-solid` to the list, regardless of what classes where defined for the notification.

I think where we landed is still correct. If you don't specify classes, you will get `fas`(`fa-solid`) class icons. If you define the classes, you *have* to specify the `fas`/`fad` class as well as the other icon classes.

readthedocsext/theme/templates/organizations/partials/header.html
```diff
@@ -20,6 +20,16 @@
 {% endcomment %}
 
 {% block organization_header %}
+  {# Render all the notifications attached to the organization.#}
+  {% for notification in organization.notifications.all %}
+    <div class="ui message">
+      <i class="fa-solid {{ notification.get_message.get_display_icon_classes }}"></i>
```
agjohnson, line 26: So, two changes required here:

```suggestion
      <i class="{{ notification.get_message.get_display_icon_classes }} icon"></i>
```

Drop the `fa-solid` class that is already defined in the icon class method, and `icon` is a required class here, from FUI.

readthedocsext/theme/templates/projects/partials/header.html
```diff
@@ -20,6 +20,14 @@
 {% endcomment %}
 
 {% block project_header %}
+  {# Render all the notifications attached to the project.#}
+  {% for notification in project.notifications.all %}
+    <div class="ui message">
+      <i class="fa-solid {{ notification.get_message.get_display_icon_classes }}"></i>
```
agjohnson, line 26: And same changes here

```suggestion
      <i class="{{ notification.get_message.get_display_icon_classes }} icon"></i>
```

agjohnson (approved):

Issue #211, opened 2023-08-31
Addons: create project admin page to configure addons


humitos:
The new addons project will give users a lot of flexibility: we will eventually allow them to use the JSON response to build anything they want on the frontend (see https://github.com/readthedocs/addons/issues/13) and also because they could also inject the web-components by hand.

However, for the most common use case (no "development configuration") we still want to provide a way to customize some of the addons like "Default filter" on search, the `root_selector` for docdiff and others. I'm thinking all these things could be a page (or Addons subpages -- like one per addon: `Addons -> Search`) with a simple form on the project Admin for now.

Things we want to expose to configure currently (there will be more, sure):

* Docdiff:
  * enabled/disabled
  * `root_selector`: tells Docdiff the selector to perform the diff against the two versions
  * additions and/or deletions (see https://github.com/readthedocs/addons/issues/101)
* External version warning:
  * enabled/disabled
* Non latest version warning:
  * enabled/disabled
  * Javascript logic to make the comparison (see https://github.com/readthedocs/addons/issues/71)
* Search
  * enabled/disabled
  * Default filter (e.g. `project: test-builds`)
  * Filters (e.g. "Search only on this project", "Include subprojects", etc)
  * Fuzzy (see https://github.com/readthedocs/addons/issues/84)

> It may be too soon to start working on this development, but I want to start the conversation to understand if we all are in the same page.

I'd like to start collecting opinions here because there are different ideas around how to do this and I'd like to think more deeply on them. We can always say: "These are the defaults and there is no way to change them unless you inject the web-components by yourself and pass attributes", like: `<readthedocs-search default-filter="project: test-builds"></readthedocs-search>`, which is totally valid as well (in particular at this stage)

agjohnson:
> we still want to provide a way to customize some of the addons like "Default filter" on search

+1 here, for very common options at least.

> These are the defaults and there is no way to change them unless you inject the web-components by yourself and pass attributes"

This is definitely an option too, though I might save this option for more advanced customization though? I'd say for basic configuration adding UI is indeed the way to go.

> I'm thinking all these things could be a page (or Addons subpages -- like one per addon: Addons -> Search) with a simple form on the project Admin for now.

Hrm, I'm in between options here myself. I might lean towards the tabbed interface with multiple form instances. It's a little bit harder to get initial page load correct, but the overall UX is quite nice. It saves us from redundant views/templates too.

- Single form instance with multiple field sets
[...]

PR #262, opened 2024-01-15
Addons: leave Crispy forms to handle the rendering of the form

humitos:
This is the correct way to render a Crispy form.
It's also required for https://github.com/readthedocs/ext-theme/issues/211

### DIFF
--- a/readthedocsext/theme/templates/projects/addons_form.html
+++ b/readthedocsext/theme/templates/projects/addons_form.html
@@ -19,12 +19,8 @@
     {% endblocktrans %}
   </div>
 
-  <form class="ui form"
-        method="post"
-        action="{% url 'projects_addons' project_slug=project.slug %}">
-    {% csrf_token %}
-    {{ form|crispy }}
+  <div class="ui form">
+    {% crispy form %}
+  </div>
 
-    <input class="ui primary button" type="submit" value="{% trans "Save" %}">
-  </form>
 {% endblock %}


agjohnson:
What does the form look like when rendered this way? The `form|crispy` was most likely intentional, so we can control the additional elements like the submit button.

humitos:
Oh, I saw that we were using `{% crispy form %}` for a form that has a custom `Layout`: https://github.com/readthedocs/readthedocs.org/blob/919d10aca54650c4a9af5f597f8a0f08646d3206/readthedocs/projects/forms.py#L280-L292 and https://github.com/readthedocs/ext-theme/blob/a39cec9063a82d55e1a549f914d14c90dbb2e506/readthedocsext/theme/templates/projects/settings_advanced_form.html#L10-L15.

I followed that pattern because it does not render properly otherwise. The following screenshot is using `origin/main` branch:


Screenshot_2024-01-16_09-23-15

Using this PR, it renders the layout properly as shown in https://github.com/readthedocs/readthedocs.org/pull/11031


The documentation of crispy form says that using the filter is fine, but the tag gives a lot more customization power: https://django-crispy-forms.readthedocs.io/en/latest/filters.html

> The best way to make your forms crisp is using the [{% crispy %} tag with forms](https://django-crispy-forms.readthedocs.io/en/latest/crispy_tag_forms.html#crispy-tag-forms).

and it's the required way to use layouts as I want to do here:

> django-crispy-forms implements a class called FormHelper that defines the form rendering behavior. Helpers give you a way to control form attributes and its layout, doing this in a programmatic way using Python

_(from https://django-crispy-forms.readthedocs.io/en/latest/crispy_tag_forms.html#crispy-tag-forms)_

> so we can control the additional elements like the submit button.

We can control this as well, but instead of doing it at the template level, it's done in Python: https://github.com/readthedocs/readthedocs.org/blob/919d10aca54650c4a9af5f597f8a0f08646d3206/readthedocs/projects/forms.py#L292

agjohnson:
Also, it's probably worth doing this in the simplest way possible for now. I am going to want to split this up into a tabbed interface or multiple subpages, so we most likely will need multiple forms/views eventually (instead of fieldsets). I would plan for this change in the future with the code you are writing now, probably by avoiding planning out the UI too much to start.

Also also, this is rendering a little extra bad right now as I recently fixed a bug in checkbox display. I was wondering if this affected any of the checkboxes but didn't see any worse off. But it does look like multiple checkboxes are rendering inline instead of as block elements. I'll need a fix for this, which would help render this at least as a single list of checkboxes.

humitos:
OK. I don't want to change the pattern you were following (or the one you _want to follow_). I will check your review from that PR, but I'm happy to adapt this work to keep it simple and move forward with it. Currently, I just want to start giving users a way to enable/disable _each_ addon and be able to promote/discover it more.

agjohnson:
And let's continue discussion around forms and the Crispy layout API at https://github.com/readthedocs/readthedocs.org/issues/10777. I find it's easier to author, manipulate, and style forms inside templates, but if there are merits for using the layout API instead, I'd like that input too.

agjohnson:
And let's continue discussion around forms and the Crispy layout API at https://github.com/readthedocs/readthedocs.org/issues/11042. I find it's easier to author, manipulate, and style forms inside templates, but if there are merits for using the layout API instead, I'd like that input too.

- https://github.com/readthedocs/readthedocs.org/issues/11042

Issue #264, opened 2024-01-16
Remove "sponsored" footer on Commercial


ericholscher:
Realized we have the "sponsored by" content in the footer on the commercial side. Would be good to have a flag where we can only show this on community.

humitos:
I'm not sure how we are differentiating community from commercial templates. I didn't find these differentiation in the repository, but I assume we could just remove the `footer_sponsors` block in commercial ones:

https://github.com/readthedocs/ext-theme/blob/8889fae8459927ca924f24b160e6d8f0d54788b5/readthedocsext/theme/templates/includes/footer.html#L133-L136

ericholscher:
Yea, @agjohnson is there a pattern here for this?

agjohnson:
In other contexts, `USE_ORGANIZATIONS` makes sense, though less so here. I would just reuse that context variable for now, ie:

https://github.com/readthedocs/ext-theme/blob/d866730bfad4636921c6970bed3931ae4a3441a7/readthedocsext/theme/templates/includes/header.html#L69

ericholscher:
Cool, I used `USE_PROMOS`, since hopefully we'll get ORGANIZATIONS on .com before we ship ads on it 😆

humitos:
At this point, I think we should have a `ON_READTHEDOCS_COMMUNITY` and/or `ON_READTHEDOCS_BUSINESS` instead of using settings that were meant for different purposes 😄

ericholscher:
@humitos Yea. I think we have `PRODUCTION_DOMAIN == 'readthedocs.org'`, which is probably what we actually want? Though we could always just have another value in our context processor that precomputed that :D

PR #266, opened 2024-01-17
Hide "Sponsored by" section on commercial

ericholscher:
Fixes https://github.com/readthedocs/ext-theme/issues/264

### DIFF
--- a/readthedocsext/theme/templates/includes/footer.html
+++ b/readthedocsext/theme/templates/includes/footer.html
@@ -131,69 +131,70 @@
     {% endblock footer_copyright %}
 
     {% block footer_sponsors %}
+      {% if USE_PROMOS %}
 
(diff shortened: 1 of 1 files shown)


readthedocsext/theme/templates/includes/footer.html
```diff
@@ -131,6 +131,8 @@ <h4 class="ui inverted sub header">Tools</h4>
     {% endblock footer_copyright %}
 
     {% block footer_sponsors %}
+
+    {% if USE_PROMOS %}
```
agjohnson, line 135: You might need to run the HTML linter locally -- it's djhtml, which runs through pre-commit. I think the pre-commit is not running in CI for some reason

agjohnson (approved):

PR #271, opened 2024-01-30
Several form field fixes

agjohnson:
- Wrap checkbox with correct surrounding element for a field
- Add search class to multi select dropdown to make it keyboard searchable
- Multi select label icon was not showing from missing codepoint

image

[Screencast from 2024-01-29 23-55-21.webm](https://github.com/readthedocs/ext-theme/assets/1140183/8cbc1654-4966-4e56-97b5-d49247af2c80)


humitos (approved):
Looks great!

> Wrap checkbox with correct surrounding element for a field

Would this also solve the issue on the addons admin page?

humitos:
I'm merging this so it goes out in today's deploy.

agjohnson:
> Would this also solve the issue on the addons admin page?

Yup, this was the fix for that particular page:

image

PR #272, opened 2024-01-31
Use build.version_name first, in case build.version is None

agjohnson:
When versions are removed, the builds remain with a `build.version = None`,
along with some additional attributes on the Build instance (for example
`build.version_name`).

This bug affected any display that listed projects/versions that listed
the most recent build.

- Fixes readthedocs/readthedocs.org#11061

### DIFF
--- a/readthedocsext/theme/templates/includes/elements/chips/build.html
+++ b/readthedocsext/theme/templates/includes/elements/chips/build.html
@@ -13,8 +13,13 @@
 {% endblock chip_icon %}
 
 {% block chip_text %}
-  {# Allow for override of the text for version listing, where version is already stated #}
-  {% firstof text build.version_name build.version.verbose_name %}
+  {% comment %}
+    Allow for override of the text for version listing, where version is
+    already stated. Also use build method to get version name, in case version
+    instance was deleted at some point. The version name is stored on the build
+    in this scenario.
+  {% endcomment %}
+  {% firstof text build.get_version_name %}
 {% endblock chip_text %}
 
 {% block chip_detail_text %}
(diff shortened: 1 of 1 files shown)


humitos (approved):
Looks good to me.

> When versions are removed, the builds remain with a build.version = None,
along with some additional attributes on the Build instance (for example
build.version_name).

I'd suggest to put this comment in the template's code. 2 of us didn't know that we were saving these values in the build object itself to be able to delete the versions while keeping the build for history purposes.

readthedocsext/theme/templates/includes/elements/chips/build.html
```diff
@@ -14,15 +14,15 @@
 
 {% block chip_text %}
   {# Allow for override of the text for version listing, where version is already stated #}
-  {{ text|default:build.version.verbose_name }}
+  {% firstof text build.version_name build.version.verbose_name %}
```
stsewd, line 17: We have some helper methods

https://github.com/readthedocs/readthedocs.org/blob/505ee4c3030f00bac16068ddec6b817adc8e7a54/readthedocs/builds/models.py#L978-L991

humitos: It seems this method has a bug, actually. https://github.com/readthedocs/readthedocs.org/blob/505ee4c3030f00bac16068ddec6b817adc8e7a54/readthedocs/builds/models.py#L983, it should access `.slug` instead of `.verbose_name`

agjohnson: I won't do anything with the bug there, I am only using the version verbose name, so the first method works for this case.

Updated in b428bb6

PR #277, opened 2024-02-16
Add template and model view changes for organization authorization settings

### LINKED ISSUE #219: SSO: Complete Google Workspace form
It's always marked as disabled, and doesn't show the domain that will be used.

Screenshot 2023-09-27 at 11-34-20 santosorg - Authorization - Read the Docs

And after https://github.com/readthedocs/readthedocs-corporate/pull/1595 there is no need for the user to set the domain, but the domain from their Google account is used automatically instead.

Which means that users will need to connect their own account to Google (linked to a domain) before setting up the integration, that's why I put this note in case they don't have it set up.

https://github.com/readthedocs/readthedocs-corporate/blob/9e31a22ff99cbe7c26c5d2002ad619ad1f20ee36/readthedocsinc/corporate/templates/organizations/admin/sso_edit.html#L52-L62

agjohnson:
This relies on https://github.com/readthedocs/readthedocs-corporate/pull/1730

This drops the very custom form for a more native form, but retains some
of the styles of the authorization provider list. This is now just a
informational block describing the providers.

This also adds the KO view logic back to the view, so that we can
conditionally show/require the domain field.

Also raised in the PR above, a warning block is added when the provider
value is changing, so there is some information that
authentication/authorization might fail.

[Screencast from 2024-02-16 11-34-38.webm](https://github.com/readthedocs/ext-theme/assets/1140183/252d8622-38a2-4154-81bb-5e02baee53ec)

- Fixes #219


agjohnson:
Test failure is https://github.com/CircleCI-Public/browser-tools-orb/issues/102

Tests pass locally:

```

Chrome: |██████████████████████████████| 4/4 test files | 11 passed, 0 failed

Finished running tests in 2.8s, all tests passed! 🎉

```

agjohnson:
@stsewd based on your comment in #219, it wasn't clear to me what we want to do with the domain field yet. If this is read only, we probably want to handle this field differently at the Form level I suppose?

humitos:
I didn't review the code, but the video looks great! 💯 

> This drops the very custom form for a more native form, but retains some of the styles of the authorization provider list. This is now just a informational block describing the providers.

I like how it looks. However, I'm always hesitant to manipulate forms directly from the template because it gets hard to maintain the HTML code pretty easily, but I don't have a better suggestion here, so 🤷🏼 

> This also adds the KO view logic back to the view, so that we can conditionally show/require the domain field.

If I understand correctly, this domain field should be removed since the user shouldn't write anything here. The user enabling Google SSO should have their Google account already connected and the domain should be populated from there only --for security reasons.

> Also raised in the PR above, a warning block is added when the provider value is changing, so there is some information that authentication/authorization might fail.

This is 💯

stsewd:
> But on the domain field, at what point do we expect a value in this field? After the user submits this form I assume? If so, it might make sense to drop domain as a form field and instead show some extra content with the value above the form field -- similar to how the shareable link for authentication shows up for Allauth SSO on that form.

I think it makes sense before, so the user knows the action he's about to take. Like showing "You are about to connect to your @readthedocs.com Google Workspace" or something, also, we should probably allow connecting to just one Google account.

humitos:
@agjohnson @stsewd 
> But on the domain field, at what point do we expect a value in this field? After the user submits this form I assume?
> I think it makes sense before, so the user knows the action he's about to take

Isn't this a good case for the pre-validation mixin you created in the other PR? 😉

agjohnson:
I was thinking about that, there might be a way. The tricky part is that if we send a prevalidation form error about google being an invalid option, that also affects the user selecting github SSO as an option.

There might be a general error up front for both though: "You need to connect a service account to use single sign-on"

stsewd:
> True, this would be nicer UX. Is the domain field value already set before the user uses this form? I don't have Google SSO set up for testing yet so can't test it.

yeah, https://github.com/readthedocs/readthedocs-corporate/blob/a1dc088a2ad5476110c5fac0f938c135e551653d/readthedocsinc/acl/sso/forms.py#L62-L62

readthedocsext/theme/templates/organizations/settings/sso_edit.html
```diff
@@ -41,104 +42,87 @@
...
-            <strong>{% trans "Read the Docs teams" %}</strong>
-            <p>
-              {% blocktrans trimmed %}
-                Use built-in organization teams to manage access to your projects and documentation.
-              {% endblocktrans %}
-            </p>
-          </label>
-          <input type="radio" name="provider" tabindex="0" class="hidden" value checked />
+  <div data-bind="using: new OrganizationSettingsAuthorizationView()">
+    <form class="ui form" method="POST" action=".", data-bind="css: { warning: show_warning }">
```
humitos, line 46:
```suggestion
    <form class="ui form" method="POST" action="." data-bind="css: { warning: show_warning }">
```

readthedocsext/theme/templates/organizations/settings/sso_edit.html
```diff
@@ -41,104 +42,87 @@
...
+            </div>
+          </div>
+
+          <div class="item">
+            <i class="fab fa-github medium middle aligned icon"></i>
+            <div class="content">
+              <div class="header">{{ form.PROVIDER_ALLAUTH }}</div>
+              <div class="description">
+                {% blocktrans trimmed %}
+                  Control all access to your projects by reusing permissions from your project's version control provider.
```
humitos, line 76:
```suggestion
                  Control all access to your projects by reusing permissions from your repository's version control provider.
```

readthedocsext/theme/templates/organizations/settings/sso_edit.html
```diff
@@ -41,104 +42,87 @@
...
+          <div class="item">
+            <i class="fab fa-google middle aligned icon"></i>
+            <div class="content">
+              <div class="header">
+                {{ form.PROVIDER_EMAIL }}
+                {% if not form.has_google_sso_feature %}
+                  <a class="ui secondary horizontal label" href="{% url 'subscription_detail' organization.slug %}">
+                    Pro plan
+                  </a>
+                {% endif %}
```
humitos, line 92: Why not keeping the "Pro plan" badge here always? That way, users with the Pro plan will know the feature they are using is because they are paying for the Pro plan. What do you think?

readthedocsext/theme/templates/organizations/settings/sso_edit.html
```diff
@@ -41,104 +42,87 @@
...
+              </div>
+              <div class="description">
+                {% blocktrans trimmed %}
+                  Users with a verified Google Workspace email account matching
+                  your domain will auto-join your organization. This can be
+                  used on top of Read the Docs teams.
```
humitos, line 98: "This can be used on top of Read the Docs teams" is not 100% true. In particular now that we are enforcing users to have a Google account connected if the organization has Google SSO enabled.

I think we should mention that "All users require to have a Google account connected" and "Permissions will be managed by Read the Docs teams", but written with _nice English words_ 😄

readthedocsext/theme/templates/organizations/settings/sso_edit.html
```diff
@@ -41,104 +42,87 @@
...
-              {% blocktrans trimmed %}
-                Control all access to your projects by reusing permissions from your project's version control provider.
-                This replaces Read the Docs managed teams.
-              {% endblocktrans %}
-            </p>
-          </label>
-          <input type="radio" name="provider" tabindex="0" class="hidden" value="allauth" />
+      <div class="ui basic fieldset segment">
+        {% alter_field form.provider data_bind="valueInit: provider, value: provider" %}
+        {{ form.provider | as_crispy_field }}
```
humitos, line 109: Why this field doesn't need the `<div class="ui field` as the following one?

readthedocsext/theme/templates/organizations/settings/sso_edit.html
```diff
@@ -41,104 +42,87 @@
...
-              {% endblocktrans %}
-            </p>
-          </label>
-          <input type="radio" name="provider" tabindex="0" class="hidden" value="email" />
+        <div class="ui warning message ko hidden" data-bind="css: { hidden: !show_warning() }">
+          <i class="fad fa-circle-exclamation icon"></i>
+          {% blocktrans trimmed %}
+            When switching authentication providers, user permissions granted
+            through your current provider may be discarded. Ensure your new
+            authentication provider is configured correctly before making this change.
```
humitos, line 120: "... and all the users have a connected Google / VCS account" or similar.

humitos (approved):
🚀

readthedocsext/theme/templates/organizations/settings/sso_edit.html
agjohnson: Nice catch. I wish we could run the html-validator on Django templates :pray:

readthedocsext/theme/templates/organizations/settings/sso_edit.html
agjohnson: Great point. This used to say "Upgrade to the Pro plan", so it made more sense to hide after upgrading. But I suppose I would still say it is a little confusing to keep a badge/upgrade link after the user upgrades. I would wonder if my plan is correct.

I checked on Front, which I feel does a good job of communicating what plan is required for their features. The automation rules feature is included in our plan and there is no longer a call out for the required plan.

Also also though, reminding the user what they are paying for seems like a good idea.

For now, I think I'll drop the conditional here. But next up, I instead think this could be a reproducible pattern (template includes, or maybe even a small web component?). There are a number of places we could use a pro plan/etc badge that links to more information on upgrading, and it should probably communicate plan status. I'll open an issue to discuss more.

readthedocsext/theme/templates/organizations/settings/sso_edit.html
agjohnson: I should note this as it is custom.

`as_crispy_field` adds this by default, so `provider` does have this already.

Below, I wrapped in an additional `field` strictly to give KO the ability to set this element as `ui disabled field`. This works because `.ui.disabled.field > .ui.field` still renders correctly as a disabled field.

I am now realizing that `ui field` is not required at all anymore though, as there is no binding adding the `disabled` field here.

readthedocsext/theme/templates/organizations/settings/sso_edit.html
humitos: IIRC, there was a plugin for this... It's not this one in particular what I remember, but it may be useful: https://www.djlint.com/

readthedocsext/theme/templates/organizations/settings/sso_edit.html
humitos: > Also also though, reminding the user what they are paying for seems like a good idea.

Yeah, this was my thinking 👍🏼 

> For now, I think I'll drop the conditional here. But next up, I instead think this could be a reproducible pattern (template includes, or maybe even a small web component?). There are a number of places we could use a pro plan/etc badge that links to more information on upgrading, and it should probably communicate plan status. I'll open an issue to discuss more.

💯

readthedocsext/theme/templates/organizations/settings/sso_edit.html
agjohnson: Yeah, I had tried that in the past and it either didn't work or I disliked the output choices it made. It might be worth another try though.

readthedocsext/theme/templates/organizations/settings/sso_edit.html
agjohnson: Moved to #281

readthedocsext/theme/templates/organizations/settings/sso_edit.html
```diff
@@ -41,104 +42,87 @@
...
+                {{ form.PROVIDER_EMAIL }}
+                {% if not form.has_google_sso_feature %}
+                  <a class="ui secondary horizontal label" href="{% url 'subscription_detail' organization.slug %}">
+                    Pro plan
+                  </a>
+                {% endif %}
+              </div>
+              <div class="description">
+                {% blocktrans trimmed %}
+                  Users with a verified Google Workspace email account matching
```
agjohnson, line 96: @stsewd @humitos Is `Google Workspace` correct here? Are there other supported products here? Workspace is the old `Google Apps for Domains` or whatever nonsense it was called before.

humitos: Yes, Google Workspace is correct 👍🏼

PR #280, opened 2024-02-26
Version form: warn about deactivating a version

stsewd:
Screenshot from 2024-02-26 16-53-21

Closes https://github.com/readthedocs/ext-theme/issues/275


agjohnson (approved):
This looks good! I would like to maybe make this dynamic next, there isn't currently a KO view here though. Ideally, we only need to show this when the active field is changing.

PR #283, opened 2024-02-27
Fix project create wizard form step

agjohnson:
Originally raised here:

- https://github.com/readthedocs/readthedocs.org/issues/10842

I wonder if the API data source changed at some point? Either way, the
effect was that the project create form step would submit and not work
(though not raise a validation error either).

I finally noticed this locally today, and in the deploy while testing. I
included a fix for the default branch name, that wasn't auto populating
either.

- Fixes https://github.com/readthedocs/readthedocs.org/issues/10842


humitos:
> I wonder if the API data source changed at some point? Either way, the
> effect was that the project create form step would submit and not work
> (though not raise a validation error either).

I'm not sure if the API has changed, but I guess it hasn't. I think there have been some confusion here since @stsewd opened an issue some time ago about this: https://github.com/readthedocs/readthedocs.org/issues/10943

humitos (approved):

PR #292, opened 2024-03-01
Update org/project headers and make them always collapsed by default

### LINKED ISSUE #229: Organizations: minor issues from initial pass at templates
Some additional notes, leftovers from #207, moved from #208 and pushed up to public beta timeframe:

- [x] Add a proper element for website link in organization header
- [x] ~~Organization owners list avatars (and project maintainers list) likely looks weird with following text "... and X more". Probably replace the last avatar with a number label with the count?~~ https://github.com/readthedocs/ext-theme/issues/137
- [x] Make sure organization delete modal works
- [x] #291
- [x] Make the custom SSO edit form work, it's not taking into consideration the initial form value currently
- [x] #290
- [x] Add right floating labels for organization on collapsed header
- [x] Add JS view for SSO edit page

### LINKED ISSUE #289: Project header can be collapsed by default
I'm finding that I want the project/organization headers collapsed almost all of the time. Initially, they were expanded on just the main project/organization detail view. This leads to some UI lag issues, and the information is rarely useful.

I want to experiment with collapsing this and trying to make the dropdown more obvious.

agjohnson:
- These headers are now collapsed by default
- The expansion button is more obvious in the header now
- Adds project/org icons in the header
- Tune some of the content that didn't quite make sense in the headers
- Remove all of the subtemplate `is_collapsed` configuration

Updates from:

- Fixes #229
- Fixes #289

image

image

image


agjohnson:
I want to play with this pattern more, and I'm not _super_ happy with this yet, but I think it's at least worth a test. I'll merge this for tomorrow, but if we find any issue with this, I would either roll this back or continue making the dropdown more obvious (or more obvious there are project details hidden).

Issue #296, opened 2024-03-06
Pages: add missing support view


agjohnson:
The views at /support/ are currently not styled and need to be ported here. We are actively pointing to the support view and have a notification that calls this URL out directly.

PR #297, opened 2024-03-06
Support: define SemanticUI `field_file.html`

humitos:
It seems that crispy forms semanticui uses a file that doesn't define. I'm following the pattern we have here to "override" this file in our own repository.

This is required for https://github.com/readthedocs/readthedocs.org/pull/11193
Related https://github.com/readthedocs/ext-theme/issues/296


humitos:
Yeah, I'm not sure either. I copied it from one of the others you put in this repository as well and deleted the non-related-to-this-field-type things 😄 . It works fine on https://github.com/readthedocs/readthedocs.org/pull/11193 at least --but I suppose there may be some edge cases that don't work.

agjohnson:
Sounds like it is probably good then, but if we hit issues that would be the next step.

PR #298, opened 2024-03-06
Support: bring `support/index.html` template here

humitos:
We want to keep track of this file in the new ext-theme. See https://github.com/readthedocs/readthedocs.org/pull/11193#discussion_r1514769919


agjohnson (approved):
I can't tell, but it looks like djhtml might have some updates to this page. This is not running in CI currently, you'll have to run locally.

humitos:
Yeah, it re-formatted it completely via pre-commit before I pushed the changes.

Issue #276, opened 2024-02-13
Auth: port CAS login template changes


agjohnson:
The authentication templates were altered in the following PR:

- https://github.com/readthedocs/readthedocs.org/pull/11101

We should replicate some of the changes to our login form.

There is also a larger project to make a stand alone or more customizable login view for CAS. The current login view intends to look like part of the application. For now, I'd start here. We can return to this at a later stage to provide a better login view for CAS or in-doc login.

agjohnson:
Another approach perhaps worth considering would be whether this makes more sense as a docs addon:

- https://github.com/readthedocs/addons/issues/248

We can always start with this issue and work on an addons implementation later if that makes the most sense.

PR #307, opened 2024-03-19
Port CAS related changes

stsewd:
Normal login

Screenshot 2024-03-19 at 13-15-21 Log in - Read the Docs

Login restricted to one provider

Screenshot 2024-03-19 at 13-15-14 Log in - Read the Docs

Login when trying to access a documentation page

Screenshot 2024-03-19 at 12-52-21 Log in - Read the Docs

Screenshot 2024-03-19 at 13-09-19 Log in - Read the Docs
Screenshot 2024-03-19 at 13-09-15 Log in - Read the Docs

Closes https://github.com/readthedocs/ext-theme/issues/276

### DIFF
--- a/readthedocsext/theme/templates/account/login.html
+++ b/readthedocsext/theme/templates/account/login.html
@@ -36,10 +36,10 @@
         </a>
 
         {# If allowed providers is given, disable the username/password form #}
+        {% blocktrans asvar text_email_disabled %}Log in using email is disabled for this organization{% endblocktrans %}
         <a class="item {% if allowed_providers %}disabled{% endif %}"
            data-tab="email"
            {% if allowed_providers %}
-           {% trans "Log in using email is disabled for this organization" asvar text_email_disabled %}
            data-content="{{ text_email_disabled }}"
            aria-label="{{ text_email_disabled }}"
            {% endif %}>
--- a/readthedocsext/theme/templates/account/login.html
+++ b/readthedocsext/theme/templates/account/login.html
@@ -39,8 +39,9 @@
         <a class="item {% if allowed_providers %}disabled{% endif %}"
            data-tab="email"
            {% if allowed_providers %}
-           title="Log in using email is not enabled for this organization"
-           aria-label="Log in using email is not enabled for this organization"
+           {% trans "Log in using email is disabled for this organization" asvar text_email_disabled %}
+           data-content="{{ text_email_disabled }}"
+           aria-label="{{ text_email_disabled }}"
            {% endif %}>
           <i class="fad fa-envelope icon"></i>
           {% block authentication_email_text %}{% trans "Log in with email" %}{% endblock %}
--- a/readthedocsext/theme/templates/account/login.html
+++ b/readthedocsext/theme/templates/account/login.html
@@ -24,7 +24,10 @@
         {% endblock authentication_header %}
       </div>
 
-      <div class="ui small stackable secondary two item menu" data-bind="semanticui: {tabs: {history: true, autoTabActivation: false}}">
+      <div
+          {# The password URL is listed as an additional tab if present #}
+          class="ui small stackable secondary item menu {% if project_password_url %}three{% else %}two{% endif %}"
+          data-bind="semanticui: {tabs: {history: true, autoTabActivation: false}}">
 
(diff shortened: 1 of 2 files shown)


agjohnson:
This looks great! I'll jump into the code next, but a couple things just looking at the screenshots:

- It's probably better to show the email login tab as disabled rather than hide it completely. You should be able to add the disabled class to the tab. Future improvements here would be giving the user feedback on why email is disabled.
- The block on "Do you have a password" feels a bit misplaced. It probably makes more sense as another tab at the top of the section ("Log in with password") rather than prose copy on all tabs. This does start to crowd the tab bar, but we can adjust this afterwards. This tab might be better to hide on the dashboard login, as opposed to using the disabled class.

humitos:
> Showing the email login tab as disabled rather than hide it completely gives the user a bit more information

👍🏼

humitos:
This is looking good 💯 . I haven't check the code, but I have some feedback from the screenshots:

- When accessing a documentation, can we check if the organization has SSO enabled and only show _that_ provider? That way, there will be one and only one way to login.
- When accessing the documentation, can we show the "Do you have a password?" option _only_ if the organization has at least one non-expired password?
- Can we make the "Forgot your password?" text less prominent? Maybe small and under the "Login" button?

stsewd:
> When accessing a documentation, can we check if the organization has SSO enabled and only show that provider? That way, there will be one and only one way to login.

I think that's already done

https://github.com/readthedocs/readthedocs-corporate/blob/7dc57dea6cbf9b9a258ff9165650e613b4522c1d/readthedocsinc/proxito/views.py#L57-L64

> When accessing the documentation, can we show the "Do you have a password?" option only if the organization has at least one non-expired password?

That feels leaky from our part

stsewd:
@agjohnson no idea how to fix the width

Screenshot 2024-03-21 at 11-03-23 Log in - Read the Docs
Screenshot 2024-03-21 at 11-03-00 Log in - Read the Docs

Also, I put a `title` as a help text when the email tab is disabled, but it doesn't show as the tab is disabled :/ if we are going to disable that tab, I think we should put some context as why.

agjohnson:
> no idea how to fix the width

The menu bar is a `.ui.two.item.menu`. If you conditionally change that to `.ui.three.item.menu` it should keep the width correct. I can help adjust the form visually afterwards, as this will compact all the buttons.

> if we are going to disable that tab, I think we should put some context as why.

Ideally we would, yeah. Let's put this on the follow up list, I can help polish these things afterwards. I feel at least with the option disabled that's some amount of visual information. Maybe we end up disabling the tab form instead though.

> Can we make the "Forgot your password?" text less prominent?

Let's put this on the follow up list with other visual tweaks too.

stsewd:
Screenshot 2024-03-21 at 12-31-41 Log in - Read the Docs
Screenshot 2024-03-21 at 12-31-16 Log in - Read the Docs
Screenshot 2024-03-21 at 12-31-03 Log in - Read the Docs

agjohnson:
Looking great, I'll take a look at the code and have opened up an issue to address all of the visual fiddling next:

- #312

readthedocsext/theme/templates/account/login.html
```diff
@@ -16,23 +16,42 @@
...
           {% block authentication_vcs_text %}{% trans "Connected account" %}{% endblock %}
         </a>
 
-        <a class="item" data-tab="email">
+        {# If allowed providers is given, disable the username/password form #}
+        <a class="item {% if allowed_providers %}disabled{% endif %}"
+           data-tab="email"
+           {% if allowed_providers %}
+           title="Log in using email is not enabled for this organization"
+           aria-label="Log in using email is not enabled for this organization"
```
agjohnson, line 43: :+1: for catching the aria label here. We'll probably tweak this pattern, but here are a few general notes for these patterns:

- I've been using `trans .. asvar ...` to localize text once. (Later we should probably just add a template filter or something).
- FomanticUI has a tooltip module that uses the attribute `data-content`. It's probably still not going to show for disabled items though.


```suggestion
           {% trans "Log in using email is disabled for this organization" asvar text_email_disabled %}
           data-content="{{ text_email_disabled }}"
           aria-label="{{ text_email_disabled }}"
```

agjohnson (approved):

Issue #268, opened 2024-01-18
CI: add pre-commit checks in CI


agjohnson:
I thought this was added, but it isn't currently running. The pieces to make this work are all in other repositories already, but I know pre-commit at least needs to be configured to run against a commit instead of the staged work.

Namely, we seem to be missing the linting passes with djhtml, but there are others that could be triggered through pre-commit too, like prettier.

cc @humitos you might already know exactly what needs to be added here, this is probably a quick addition for you if you get a chance before I do.

agjohnson:
@humitos you've mentioned issues with NPM and pre-commit might block us here. What issues are you describing?

This repo is currently using Prettier (Node.js package) through a hook, so I don't think we'll have trouble directly with Node.js. I know we've done similar custom hooks for `npm run build` that rely on NPM:

- https://github.com/readthedocs/addons/tree/main/.pre-commit

Is this the piece you had trouble with on the website?

For this issue, I just put up and merged

- #311 

This does not supersede this issue, it was just to get something working while this issue is still open. The circleci configuration can be reduced back later though.

PR #311, opened 2024-03-21
Add prettier configuration for FUI and rerun pre-commit on all files

agjohnson:
- Update Prettier configuration to run again FUI sources
- Update djhtml configuration to actuall fix tab spacing issues
- Updates pre-commit configuration to same djhtml configuration
- Installs djhtml at CI to at least start linting templates

----

- This does not replace #268

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,23 +2,19 @@ version: 2.1
 
 orbs:
   browser-tools: circleci/browser-tools@1.4.7
-  python: circleci/python@2.1.1
 
 commands:
-  run-lint:
-    description: "Run code linter"
+  run-lint-node:
+    description: "Run Node linting"
     steps:
       - checkout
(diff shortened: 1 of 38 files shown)


humitos (approved):
Ideally CircleCI should run all the checks via pre-commit and pre-commit should be the one that executes prettier, djhtml and all the other checks. However, I found myself using pre-commit with npm wasn't as easy as I thought, but it would be _ideal_ if we figure it out how to do it eventually.

agjohnson:
Yup, I noted above this does not replace the need for pre-commit, this is just a solution until someone picks up that issue.

I was going to attempt bringing in pre-commit, but I looked at the application repo and the circleci configuration seemed rather complex.

You've mentioned the NPM issue previously, and I'm still a little confused about what issues there are to running that. I'll discuss this more in

- #268

PR #317, opened 2024-03-26
Port website Plausible jquery plugin

### LINKED ISSUE #301: Add plausible to dashboard
It would be helpful to monitor usage in both the users starting to use the beta and the users that are opting out of the beta. This can probably mostly be inferred with just adding Plausible on the beta dashboard and settings a few specific goals. A few notes here

- Track referrer when getting a new user signup on beta. What application view (not URL) did the user come from?
- Track users that are clicking a link to go back to the legacy dashboard. What view did this occur on?
- It would be most helpful to see the split in traffic between the beta dashboard and the legacy dashboard. This would take using the same analytics on both sites.

----

- [x] Set up Plausible for beta.readthedocs.org and beta.readthedocs.com
- [x] Copy Plausible inection from website

agjohnson:
This moved over easily, we're doing much of the same things on the website. The event tracking might interfere with some default SUI/Lit JS event handling, it hasn't been tested thoroughly here yet.

- Fixes #301


src/js/application/plugins.js
```diff
@@ -395,6 +398,81 @@ export const semanticui = {
   },
 };
 
+/**
```
agjohnson, line 401: This is just copied pasted from:

https://github.com/readthedocs/website/blob/c9360b660753b913189777e317fea005e0ebefe8/src/js/site.js#L123

humitos (approved):
I don't know how Plausible tracker works and how we are using it here, but I suppose this is correct. On the application we are using adding just the `<script>` in the head, but I suppose this tracks specific events also, not just page views.

agjohnson:
Using an external request `<script>` is the easiest method, but on our website and application we use a more thorough method and install it as a package. On both properties, the JS we use is installed as a package (giving us dependency pinning), and included in our bundle, ensuring there aren't any outside requests to fail our JS. The noted additions about tracking individual events on both the website/application are important too -- though we aren't yet using events on the application.

PR #319, opened 2024-03-26
CI: Run pre-commit

stsewd:
I did a local run with `pre-commit run --all-files` (the changes on the .rst file come from that), but on CI I'm just checking the last diff as the pre-commit doc recommend.

pre-commit makes use of node internally, so node is needed in that step, also merging the node and python lint jobs into one.

Closes https://github.com/readthedocs/ext-theme/issues/268

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -36,8 +36,7 @@ jobs:
     steps:
       - checkout
       - node/install:
-          node-version: '20.12'
-      - run: node --version
+          node-version: "20.12"
       - run: npm ci
       - run: npm run lint
       - run: pip install pre-commit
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -41,16 +30,17 @@ commands:
...
       - image: "cimg/python:3.11.8"
     steps:
-      - run-lint-python: {}
+      - checkout
+      - node/install:
+          node-version: "20.12"
+      - run: npm ci
+      - run: npm run lint
+      - run: pip install pre-commit
+      - run: pre-commit run --from-ref origin/HEAD --to-ref HEAD
```
agjohnson, line 43: I wasn't sure exactly what to run here for pre-commit, but this makes sense to me. This looks comparable to what the application tox pre-commit env does, so seems safe :+1:

agjohnson (approved):

agjohnson:
This looks good to merge to me. I had split up the linting in #311 but that was a temporary hack, so combining the jobs again is fine.

Thanks for jumping in here @stsewd!

Issue #304, opened 2024-03-12
Pages: add documentation/user error pages


agjohnson:
From #295, I'm considering the doc/user error pages separate work to help isolate the work more. Relevant conversation here:

- There are multiple versions of the 404 page, and some features we might want to bring forward into the new dashboard.
- The new dashboard loads a number of assets and one of the things I expect to hit is the bug where the URLs and assets aren't defined on the instance that is serving the 404 page. We have a few bugs here in the past, and it's one of the reasons our error pages are more empty now.
- But a large looming issue is that the error pages we show are going to be more heavily branded like our application, which is an odd fit for user documentation sites. Our CAS/etc login for documentation views should be super minimal. This is a later problem, but defining it for now.

----

- [x] Make Proxito use the ext-theme templates: https://github.com/readthedocs/common/pull/220
- [x] Documentation error templates should be isolated from application error templates and should not extend from the application templates at all -- to avoid contaminating the template rendering with application URLs etc
- [x] Proxito error page features could use tuning
- [x] Proxito error page copy could also use tuning and reduction
- [x] Consolidate error templates under single path and pattern
  - [x] `errors/404/*.html`
  - [x] `core/teapot.html`
  - [x] `401.html`
  - [x] `/spam.html` and proxito view
- [x] Proxito should have tuned 500/401 templates, as the dashboard error templates raise exceptions about missing URLs/etc.
- [x] Proxito error templates shouldn't use icons, as we can't whitelist every single hosted domain with our Font Awesome kit.

agjohnson:
I'm running into a lot of complexity with the proxito errors that are making this work fairly complicated.

I believe I will be breaking this work up into a few chunks:

- Split out Proxito templates for base errors on the application -- 401/500/etc, as the dashboard templates include URLs that can't resolve in the header/footer/etc. Proxito can't render the existing 500/etc error templates because they include URLs in header/footer/etc that proxito doesn't have.
- Next step, create a base style for proxito error messages using the new template paths. I mentioned these won't look the same, but found there are lower level issues that prevent porting the dashboard error templates to proxito [^1][^2]
- Do something with contextualized error messages. I am going to probably remove most of the features and avoid this complexity of these pages to start. More discussion in https://github.com/readthedocs/readthedocs.org/issues/11486

[^1]: We can't use icons. The hosted icon kit limits the kit usage by domain name, so custom domains and project sub domains aren't supported.
[...]

PR #326, opened 2024-04-04
Add dashboard error pages

### LINKED ISSUE #295: Pages: add missing 4xx/5xx error pages from the website
A few things happening here:

- The website has updated 4xx and 5xx pages we can copy. I would also like to redo some pages with more personality at some point though, but this is a later problem.
- There are multiple versions of the 404 page, and some features we might want to bring forward into the new dashboard.
- The new dashboard loads a number of assets and one of the things I expect to hit is the bug where the URLs and assets aren't defined on the instance that is serving the 404 page. We have a few bugs here in the past, and it's one of the reasons our error pages are more empty now.
- *But a large looming issue* is that the error pages we show are going to be more heavily branded like our application, which is an odd fit for user documentation sites. Our CAS/etc login for documentation views should be super minimal. This is a later problem, but defining it for now.
- 🆕 Show a log in form on the commercial 404 page. See:
  - https://github.com/readthedocs/readthedocs-corporate/issues/1760

# Work here

- [x] Application error pages
  - [x] 404 page
  - [x] 404 page login conditional for commercial
  - [x] 500 page
  - [x] 403 page
- [x] #304
  - Calling this separate work to deal with later.

agjohnson:
This ports in the initial error pages from our website so that the dashboard
error pages look polished. This does not yet update the documentation error
pages, which I am considering the next batch of work afterwards.

Changes here:

- Organizes dashboard errors into an isolated path. There will be a separate
  path for less branded error pages for Proxito.
- Error pages are consolidated to all match styles. For now the styles are
  fairly minor, but I'd want to return to these later. Our error pages have
  always been goofball pages, and I appreciate that.
- I used fallback error templates in case I missed anything, it wouldn't be hard
  to eventually drop the fallback templates if we want.
- The 404 page has some slight additional logic to point to the login view for
  now, with `?next`. This mostly satisfies my complaint about the 404 page being
  misleading for users.

----

- Requires readthedocs/readthedocs.org#11263
- Closes #295
- Refs #304

### DIFF
--- a/readthedocsext/theme/templates/errors/dashboard/401.html
+++ b/readthedocsext/theme/templates/errors/dashboard/401.html
@@ -1,0 +1,36 @@
+{% extends "errors/dashboard/base.html" %}
+
+{% load core_tags %}
+{% load i18n %}
+
+{% block title %}{% trans "Unauthorized request" %}{% endblock %}
+
+{% block error_status %}401{% endblock %}
+{% block error_title %}
+  {{ block.super }}
+  {% trans "Unauthorized request" %}
+{% endblock error_title %}
+
+{% block error_icon_computer_name %}fa-lock-keyhole{% endblock %}
+{% block error_icon_mobile_name %}fa-lock-keyhole{% endblock %}
+{% block error_icon_computer_style %}--fa-primary-color: orange;{% endblock %}
+{% block error_icon_mobile_style %}--fa-primary-color: orange;{% endblock %}
+
+{% block error_content %}
+  <p>
+    {% blocktrans trimmed %}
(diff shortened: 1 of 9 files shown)


agjohnson:
400/4xx fallback

humitos (approved):
Looks great! 💯  This is ready to merge.

Note there are maybe two extra pages here we will eventually want templates:

1. 410 Gone -- for Spam project
2. 418 I'm a teapot -- for dummy URLs defined in El Proxito

agjohnson:
Great point, I'll add 410 here too, that's easy enough.

The dummy response from proxito will use a different set of proxito error templates, that one is on the list at #304

Issue #299, opened 2024-03-06
Organizations: header navigation should link directly to the only organization


agjohnson:
The top navigation "Organizations" menu could use some tuning here:

- When there is only one organization, the menu item should be "Organization" and lead to the detail view for the only org
- When there are more than one organization, go to the listing view
- If the user does not have an organization, perhaps also go to the listing view? This might be the best place to show organization creation onboarding?

Only a few users should have more than one organization either way

agjohnson:
Moved the later parts of this to #332 and can close this.

PR #327, opened 2024-04-04
Fix organization nav link

agjohnson:
Don't show if the user is not authenticated. I made the other changes in #299 to
improve the UX of this link, but it hurt performance. It's worth revisiting
later though, so I'll leave this open.

- Refs #299

### DIFF
--- a/readthedocsext/theme/templates/includes/header.html
+++ b/readthedocsext/theme/templates/includes/header.html
@@ -1,7 +1,6 @@
 {% load i18n %}
 {% load gravatar %}
 {% load static %}
-{% load organizations %}
 
 {% comment %}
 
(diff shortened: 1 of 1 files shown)


readthedocsext/theme/templates/includes/header.html
```diff
@@ -1,6 +1,7 @@
 {% load i18n %}
 {% load gravatar %}
 {% load static %}
+{% load organizations %}
```
stsewd, line 4: Is this needed?

agjohnson: Ah nope! I was trying to use these filters to get a list of organizations for the user, but it really slowed down every page load.

agjohnson:
```suggestion
```

PR #328, opened 2024-04-04
Fix sign up comment

agjohnson:
We didn't catch this in #325, but the comment is a single line comment,
not a multiline comment.

image


humitos (approved):

stsewd (approved):
oops :/

agjohnson:
I missed this one in review, I was working in Jinja all morning, which is more reasonable about comments :upside_down_face: 

Maybe djlint would pick up on this? djhtml didn't of course. I did try djlint a few weeks ago and the changes all seemed pretty reasonable. If it helps folks, we could try switching. Both reformat HTML to consistent spacing.

PR #331, opened 2024-04-05
Fix a number of display issues with login view

### LINKED ISSUE #312: Login: view style fixes
- [x] From #307 we probably want a wider view on the form now
- [x] Support three log in options better, it's a bit cramped
- [x] Fix floating on remember me checkbox
- [x] Fix casing on remember me checkbox
- [x] Fix button floating, should be right floated
- [x] Make tabs more obvious, or make tab copy more obvious that these are all log in methods
- [x] #330
- [x] Reduce visual importance of 'forgot password'
- [x] Errors on login email form stopped showing (bad password/etc)
- [x] Increase distinction between a regular user/password from shared by password (see https://github.com/readthedocs/readthedocs-corporate/issues/647)

agjohnson:
- Fixes #312

image

image

image

image

### DIFF
--- a/readthedocsext/theme/templates/account/login.html
+++ b/readthedocsext/theme/templates/account/login.html
@@ -9,9 +9,7 @@
 {% block content %}
   <div class="ui stackable padded grid">
 
-    <div class="violet four wide computer only column"></div>
-
-    <div class="twelve wide computer sixteen wide tablet sixteen wide mobile column">
+    <div class="sixteen wide computer sixteen wide tablet sixteen wide mobile column">
       {% get_providers as socialaccount_providers %}
 
       <div class="ui medium header">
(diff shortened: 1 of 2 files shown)


humitos:
In the first image, I think it should say "Email" instead of "Login" that form field, since the option selected is "Email" under "Log in using".

Also, in the first image, there is a `|` at the end of the error message. Not sure where that comes from.

readthedocsext/theme/templates/semantic-ui/errors.html
```diff
@@ -2,19 +2,32 @@
...
       <div class="header">
         {% trans "Error" as default_header %}
         {% firstof error.header default_header %}
       </div>
-      {% if error.messages|length == 1%}
-        {{ error.messages|first|safe }}
+      {% if error.messages %}
+        {% comment %}
+          Some form errors are raised with multiple error messages, using the
+          `nessages` attribute. This can either be a single message or multiple
```
humitos, line 13:
```suggestion
          `messages` attribute. This can either be a single message or multiple
```

humitos (approved):

agjohnson:
> Also, in the first image, there is a | at the end of the error message. Not sure where that comes from.

Hah, this is my Firefox/Wayland/maybe an extension? I can't get rid of the damn things, they're everywhere!

readthedocsext/theme/templates/account/login.html
```diff
@@ -25,32 +23,29 @@
...
       <div
-          {# The password URL is listed as an additional tab if present #}
-        class="ui small stackable secondary item menu {% if project_password_url %}three{% else %}two{% endif %}"
+        class="ui small stackable teal secondary menu"
         data-bind="semanticui: {tabs: {history: true, autoTabActivation: false}}">
 
-        {# This is purposely minimal, as the verbiage for "Log in with connected account" is a mountful #}
+        <div class="header item">
+          {% trans "Log in using:" %}
+        </div>
```
stsewd, line 31: Maybe put this above the list of tabs? Feels like it uneven things, specially when shown here

image

agjohnson: I did try this and it was uneven in a different way, with lots of wasted space on the first line. I'd consider this again if translations blow out the width of these buttons.

But the current layout does feel natural to me, it uses up the full width of the segment

Issue #204, opened 2023-08-22
Templates: normalize template naming


agjohnson:
You'll find in the templates path that a lot of templates are replaced with symlinks to alternate paths. This is a temporary change to avoid making application changes to fix historical accumulation of varying patterns for template paths.

Eventually, we should come back and make application changes to normalize all of the template names, so we're not using competing standards across applications.

The standards used here aim to keep templates grouped by function and *where* in the application the templates are used. Some loose rules that template names are aiming for:

- Template symlinks are added for templates that aren't using standard Django template names. The standard form of `{model}_list.html` and `{model}_detail.html`, etc are preferred
- Additional nested paths are helpful for some applications, such as `templates/projects/edit/webhook_list.html` for project setting views
- In cases like organizations and projects, templates related to the project and organization UI are grouped under `templates/projects` and `templates/organizations` instead of applications like `build` and `subscriptions`. It's confusing to have some project UI templates outside the `projects` application

### Work

Go through the template directory, find template symlinks, and update the necessary application view code so we can remove the symlink.

Issue #82, opened 2021-11-18
Add redirect template highlighting on create form


agjohnson:
There is nice formatting in the redirect listing page, but not on the form page:

![Image](https://github.com/readthedocs/ext-theme/assets/1140183/61f7c2a4-59cd-4646-b73b-eb7098caba87)

This should be a fairly minor change and could just reuse some of the template logic/KO view from the listing page.

agjohnson:
Also note that the deletion modal doesn't show the highlighted version of this string. It might make sense to make this into a method/utility function.

- https://github.com/readthedocs/ext-theme/pull/339#discussion_r1575938413

PR #339, opened 2024-04-22
Updates to project/organization create views

### LINKED ISSUE #89: Add project users create view
Make project users a proper create view and add a button linking to the create on the top of the list view of project users/maintainers/admin

Currently this is an inline form

### LINKED ISSUE #323: Views: add new create views
Implement the views added in the related PR below, replacing list/form combination views

- [x] Translation create
- [x] User maintainer create
- [x] User invitations listing
- [x] User invitations detail
- [x] Email notification create
- [x] #338
- [x] Consolidate patterns for listing view create buttons
- [ ] ~~Organization owner create~~ This view is missing, see https://github.com/readthedocs/readthedocs.org/issues/11295
- [x] Test organization owner/team invites

----

- Refs https://github.com/readthedocs/readthedocs.org/pull/11246
- Refs https://github.com/readthedocs/readthedocs.org/issues/10777

agjohnson:
- Adds missing create views
- Unifies the UX/UI on create views
- Removes forms from listing views
- Adds missing invitation detail view

----

- Fixes #323
- Fixes #89
- Organization owner view still missing, see https://github.com/readthedocs/readthedocs.org/issues/11295

### DIFF
--- a/readthedocsext/theme/templates/invitations/invitation_detail.html
+++ b/readthedocsext/theme/templates/invitations/invitation_detail.html
@@ -61,7 +61,7 @@
         <div class="extra content">
           <form class="ui two buttons" action="." method="post">
             {% csrf_token %}
-            <input class="ui positive basic button" type="submit" name="accept" value="{% trans "Accept" %}" {% if invitation.expired%}disabled{% endif %}>
+            <input class="ui positive basic button" type="submit" name="accept" value="{% trans "Accept" %}" {% if invitation.expired %}disabled{% endif %}>
             <input class="ui negative basic button" type="submit" name="decline" value="{% trans "Decline" %}">
           </form>
         </div>
--- a/readthedocsext/theme/templates/invitations/invitation_detail.html
+++ b/readthedocsext/theme/templates/invitations/invitation_detail.html
@@ -1,0 +1,72 @@
+{% extends "base.html" %}
+
+{% load i18n %}
+
+{% block title %}
+  {% trans "Invitation request" %}
+{% endblock %}
+
+{% block content %}
+  <div class="ui very padded text container">
+    <div class="ui fluid card">
+      <div class="content">
+
+        <div class="header">
+          {% blocktrans trimmed with target_name=invitation.object.name|default:target_name target_type=target_type %}
+            Invitation to join the {{ target_name }} {{ target_type }}
+          {% endblocktrans %}
+        </div>
+
+        <div class="meta">
+          {% if invitation.expired %}
+            <span class="ui red text">
+              {% trans "Invitation expired" %}
+            </span>
+          {% else %}
+            {% blocktrans trimmed with valid_until=invitation.expiration_date|timeuntil expiration_date=invitation.expiration_date|date %}
+              Expires in <time datetime="{{ expiration_date }}">{{ valid_until }}</time>
+            {% endblocktrans %}
+          {% endif %}
+        </div>
+
+        <div class="center aligned description">
+          {% block invitation_content %}
+            <p>
+              {% url "profiles_profile_detail" invitation.from_user.username as profile_url %}
+              {% with from_name=invitation.from_user.get_full_name|default:invitation.from_user.username %}
+                {% blocktrans trimmed with requestor_name=requestor_name requestor_url=requestor_url target_name=invitation.object.name|default:target_name target_type=target_type %}
+                  <a href="{{ requestor_url }}">{{ requestor_name }}</a>
+                  has invited you to join the {{ target_name }} {{ target_type }}.
+                {% endblocktrans %}
+              {% endwith %}
+            </p>
+          {% endblock invitation_content %}
+
+          {% if invitation.expired %}
+            {% block invitation_expired %}
+              <p class="ui error message">
+                {% trans "Unfortunately, this invitation has expired." %}
+                {% if invitation.object_type == "organization" or invitation.object_type == "team" %}
+                  {% trans "Ask an admin of the organization to invite you again." %}
+                {% else %}
+                  {% trans "Ask one of the maintainers of the project to invite you again." %}
+                {% endif %}
+              </p>
+            {% endblock invitation_expired %}
+          {% endif %}
+        </div>
+      </div>
+
+      {% block invitation_actions %}
+        <div class="extra content">
+          <form class="ui two buttons" action="." method="post">
+            {% csrf_token %}
+            <input class="ui positive basic button" type="submit" name="accept" value="{% trans "Accept" %}" {% if invitation.expired%}disabled{% endif %}>
+            <input class="ui negative basic button" type="submit" name="decline" value="{% trans "Decline" %}">
+          </form>
+        </div>
+      {% endblock invitation_actions %}
+
+    </div>
+  </div>
+{% endblock content %}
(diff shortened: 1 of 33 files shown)


readthedocsext/theme/templates/invitations/invitation_detail.html
```diff
@@ -0,0 +1,72 @@
...
+            {% endblock invitation_expired %}
+          {% endif %}
+        </div>
+      </div>
+
+      {% block invitation_actions %}
+        <div class="extra content">
+          <form class="ui two buttons" action="." method="post">
+            {% csrf_token %}
+            <input class="ui positive basic button" type="submit" name="accept" value="{% trans "Accept" %}" {% if invitation.expired%}disabled{% endif %}>
```
humitos, line 64:
```suggestion
            <input class="ui positive basic button" type="submit" name="accept" value="{% trans "Accept" %}" {% if invitation.expired %}disabled{% endif %}>
```

readthedocsext/theme/templates/invitations/invitation_detail.html
```diff
@@ -0,0 +1,72 @@
...
+            <input class="ui negative basic button" type="submit" name="decline" value="{% trans "Decline" %}">
```
humitos, line 65: Shouldn't this button be disabled as well if the invitation is expired?

readthedocsext/theme/templates/invitations/partials/invitation_list.html
```diff
@@ -23,11 +25,12 @@
 {% endblock list_placeholder %}
 
 {% block list_item_right_buttons %}
-  {% if request.user|is_admin:organization %}
+  {# Invitation has a more direct method ``can_revoke_invitation`` but Django templates can't call methods with arguments #}
```
humitos, line 28: Yeah, we need to create a template tag/filter for this --probably makes sense as `{% if invitation|can_be_revoked %}` or similar

readthedocsext/theme/templates/projects/notification_form.html
```diff
@@ -0,0 +1,21 @@
...
+    {% trans "Add notification" %}
+  {% endif %}
+{% endblock %}
+
+{% block project_edit_content %}
+  <form class="ui form" method="post" action=".">
+    {% csrf_token %}
+    {{ form|crispy }}
+    <input class="ui primary button" type="submit" value="{% trans "Save" %}">
+  </form>
```
humitos, line 20: Does it make sense to move this chunk of code to a `includes/partials/crud/form.html` or similar as you are doing with create/remove buttons?

It seems this form it's pretty generic and we can probably re-use it everywhere.

readthedocsext/theme/templates/projects/partials/edit/keys_list.html
```diff
@@ -17,13 +17,9 @@
...
-  <div class="disabled item">
-    <a class="ui disabled icon button"
-       href="{# {% url "..." project.slug %} #}">
-      <i class="fa-solid fa-plus icon"></i>
-      {% trans "Add SSH key" %}
-    </a>
-  </div>
+  {% url "..." project.slug as create_url %}
+  {% trans "Add SSH key" as create_text %}
+  {% include "includes/crud/create_button.html" with url=create_url text=create_text %}
```
humitos, line 22: What `...` means here?

readthedocsext/theme/templates/projects/partials/edit/redirect_list.html
```diff
@@ -71,26 +68,12 @@
...
-          {% csrf_token %}
-          <input class="ui negative button" type="submit" value="{% trans "Remove" %}">
-        </form>
-      </div>
-    </div>
-  </div>
+  {% url "projects_redirects_delete" project_slug=project.slug redirect_pk=object.pk as form_url %}
+  {% trans "Remove redirect" as action_text %}
+  {% blocktrans trimmed asvar content_text with redirect=object %}
+    Remove redirect {{redirect}}?
```
humitos, line 74:
```suggestion
    Remove redirect {{ redirect }}?
```

humitos: I think we should show the `redirect` in a highlighted way to differentiate the user's data from the regular text. Also, my redirect has a `Description` that we should probably show here as well, since it's a better way to differentiate from the others.

Screenshot_2024-04-23_11-22-07


Also also, shouldn't the "Remove" button be red in some way? (maybe on hover at least ?)

readthedocsext/theme/templates/projects/partials/edit/translation_list.html
```diff
@@ -48,6 +37,15 @@
 {% endblock list_placeholder_text %}
 
 {% block list_item_right_buttons %}
+  {% comment %}
+    {# TODO there is no edit view for this model, only create and delete #}
```
humitos, line 41: Do we have an issue for this?

readthedocsext/theme/templates/projects/project_notifications_form.html
humitos, line 1: What are these symlinks for?

humitos (approved):
Looks good to me! 💯 . I left a few questions to understand the pattern a little better on those small details.

readthedocsext/theme/templates/invitations/invitation_detail.html
agjohnson: I was mostly just updating the underlying template, which did this same thing. I do feel "Decline" doesn't quite fit in this case, but only having the decline button enabled makes sense for now too. The "Decline" button should be enabled and the user should have some action available if the invitation is expired.

Ideally though, these buttons would be replaced with a single "Dismiss" button. We could come back to this if we want, but might need a third option at the view? I didn't look deeply

readthedocsext/theme/templates/invitations/partials/invitation_list.html
agjohnson: Yeah, it's a silly limitation of Django templates. Making a method into a duplicate template tag does not feel like a better or more maintainable solution, it's just a hoop to jump through. I keep running up against these limitations, so I'm a bit annoyed with Django at the moment.

readthedocsext/theme/templates/projects/notification_form.html
agjohnson: Very true. For most forms it should indeed be enough.

readthedocsext/theme/templates/projects/partials/edit/keys_list.html
agjohnson: Just a placeholder, this URL doesn't exist yet

readthedocsext/theme/templates/projects/project_notifications_form.html
agjohnson: I symlink templates that we misnamed so that we can start to organize them in a less chaotic way. In particular, this template (and most of the templates in `projects/`) follow a naming pattern of `{application}/{application}_{model}_{type}.html` instead of `{application}/{model}_{type}.html`.

For more, see

- #204

readthedocsext/theme/templates/projects/partials/edit/translation_list.html
agjohnson: We don't. Maybe it's not needed here, as there are no other options to edit in the form? That is, the project is either connected or not, there are no other fields like alias to edit.

readthedocsext/theme/templates/projects/partials/edit/redirect_list.html
agjohnson: This is a good addition for later, it's unrelated to the changes here so I will hold off on making any changes here. I have noted similar at:

- https://github.com/readthedocs/ext-theme/issues/82

PR #348, opened 2024-05-01
SAML: port template

stsewd:
This is just a basic form to show information about the integration.

Screenshot 2024-05-01 at 12-28-39 org - SAML integration - Read the Docs

Screenshot 2024-05-01 at 12-28-27 neworg - SAML integration - Read the Docs

Ref https://github.com/readthedocs/readthedocs-corporate/pull/1770


readthedocsext/theme/templates/organizations/settings/saml.html
```diff
@@ -0,0 +1,141 @@
...
+{% block organization_edit_content_header %}{% trans "SAML integration" %}{% endblock %}
+
+{% block organization_edit_content %}
+
+  <div class="ui padded segment">
+    <h4>
+      {% trans "Organization SAML settings" %}
+    </h4>
+
+    <div class="ui message">
```
agjohnson, line 20: This doesn't need to be a message, a simple paragraph is fine.

We should probably either provide the user with more help here, or we could just drop this paragraph too. The directions in this paragraph probably won't provide much additional context for the user.

readthedocsext/theme/templates/organizations/settings/saml.html
```diff
@@ -0,0 +1,141 @@
...
+      <p>
+        {% blocktrans %}
+          Use the following information to configure your SAML integration.
+        {% endblocktrans %}
+      </p>
+    </div>
+
+    <div class="ui form">
+      <div class="field">
+        <label>{% trans "Single sign on URL" %}</label>
```
agjohnson, line 30:
```suggestion
        <label>{% trans "Single sign-on URL" %}</label>
```

readthedocsext/theme/templates/organizations/settings/saml.html
```diff
@@ -0,0 +1,141 @@
...
+      </div>
+      <div class="field">
+        <label>{% trans "Name ID format" %}</label>
+        <div class="ui fluid action input">
+          <input type="text" value="EmailAddress" id="saml-name-id-format" readonly=true>
+          <button class="ui teal right icon button" data-clipboard-target="input#saml-name-id-format">
+            <i class="fa-solid fa-copy icon"></i>
+          </button>
+        </div>
+      </div>
```
agjohnson, line 55: These all look great, I like using the pattern of the clipboard buttons.

readthedocsext/theme/templates/organizations/settings/saml.html
```diff
@@ -0,0 +1,141 @@
...
+    </div>
+  </div>
+
+  {% if sso_integration %}
+    <div class="ui padded segment">
+      <h4>
+        {% trans "Read the Docs login URL" %}
+      </h4>
```
agjohnson, line 63: Instead of making this a separate segment, what about including this with the rest of the read only attributes in the segment above?

readthedocsext/theme/templates/organizations/settings/saml.html
```diff
@@ -0,0 +1,141 @@
...
+        <div class="field">
+          <label>{% trans "Single log out URL" %}</label>
+          <div class="ui fluid input">
+            <input type="text" value="{{ slo_url }}" readonly>
+          </div>
+        </div>
+        <div class="field">
+          <label>{% trans "Certificate" %}</label>
+          <div class="ui fluid input">
+        <!--<input type="text" value="{{ certificate }}" readonly>-->
```
agjohnson, line 119:
```suggestion
```

readthedocsext/theme/templates/organizations/settings/saml.html
```diff
@@ -0,0 +1,141 @@
...
+          <button class="ui teal right icon button" data-clipboard-target="input#rtd-saml-login-url">
+            <i class="fas fa-copy icon"></i>
+          </button>
+        </div>
+      </div>
+    </div>
+
+    <div class="ui padded segment">
+      <h4>
+        {% trans "SAML integration details" %}
```
agjohnson, line 82: It feels like these sections are swapped. These feel like "settings" instead of details. The read only attributes above, labeled "settings" feel like they should be labeled "integration details".

readthedocsext/theme/templates/organizations/settings/saml.html
```diff
@@ -0,0 +1,141 @@
...
+      </h4>
+
+      <div class="ui form">
+        <div class="field">
+          <label>{% trans "Domain" %}</label>
+          <div class="ui fluid input">
+            <input type="text" value="{{ domain }}" readonly>
```
agjohnson, line 89: Is there not a `form` instance that we can use here instead of making the form manually?

readthedocsext/theme/templates/organizations/settings/saml.html
```diff
@@ -0,0 +1,141 @@
...
+          </div>
+        </div>
+        <div class="field">
+          <label>{% trans "Entity ID" %}</label>
+          <div class="ui fluid input">
+            <input type="text" value="{{ entity_id }}" readonly>
+          </div>
+        </div>
+        <div class="field">
+          <label>{% trans "Single sign on URL" %}</label>
```
agjohnson, line 105:
```suggestion
          <label>{% trans "Single sign-on URL" %}</label>
```

readthedocsext/theme/templates/organizations/settings/saml.html
```diff
@@ -0,0 +1,141 @@
...
+          <div class="ui fluid input">
+            <input type="text" value="{{ sso_url }}" readonly>
+          </div>
+        </div>
+        <div class="field">
+          <label>{% trans "Single log out URL" %}</label>
```
agjohnson, line 111:
```suggestion
          <label>{% trans "Single logout URL" %}</label>
```

readthedocsext/theme/templates/organizations/settings/saml.html
stsewd: The above attributes are used to set up the SAML integration from the identity provider, including those in that section may lead to confusions about where to use those values. After the allauth upgrade, it should also be possible to link an existing account to SAML, for that we need another URL.

readthedocsext/theme/templates/organizations/settings/saml.html
stsewd: Nope, we fetch all the attributes from a single URL, we don't have a form for all attributes.

readthedocsext/theme/templates/organizations/settings/saml.html
agjohnson: I would use the exact UI pattern from the other SSO view then:

image

From https://beta.readthedocs.com/organizations/read-the-docs-sso-qa/authorization/

Namely, the callout should use the same element structure and appear at the top of the page, and not nested inside another segment. The title should be more descriptive, ie "Invite users into your organization" or whatever we are asking users to do there. This link is something we're expecting user to reuse so it should be easier to find.

readthedocsext/theme/templates/organizations/settings/saml.html
agjohnson: The read only attributes that users use on their own SAML provider are not settings to our users, the user cannot *set* them. I would try to avoid ambiguity here and be more descriptive with what these values actually are or do for the user.

- For the top section: "Details for your SAML provider" or "Configure your SAML provider"
- For the form section where the user does enter settings, all other forms in admin views use "Add ..." and "Update ..." as titles -- so "Add SAML integration" and "Update SAML integration" based on whether or not the form is bound.

readthedocsext/theme/templates/organizations/settings/saml.html
agjohnson: I'd like to do a bit more with the form here. This UI is making new patterns where I'd like to instead keep things consistent across all the views.

It's okay to show the read only values as a supplemental to a form instance, but the form should always be treated as a form at very least. That is, the two user editable fields (domain, metadata URL) should always be part of a form display through Crispy. Right now, after submitting, there doesn't seem to be a way to change or delete these fields.

With those fields displaying as a form, I would then display the readonly attributes separately, as to not confuse them with the user editable fields (domain and metadata URL).

readthedocsext/theme/templates/organizations/settings/saml.html
```diff
@@ -0,0 +1,138 @@
...
+        <div class="ui fluid action input">
+          <input type="text" readonly id="rtd-saml-login-url" value="{{ rtd_saml_login_url }}" />
+          <button class="ui teal right icon button" data-clipboard-target="input#rtd-saml-login-url">
+            <i class="fas fa-copy icon"></i>
+          </button>
+        </div>
+      </div>
+    </div>
+
+    <div class="ui padded segment">
```
agjohnson, line 78: Other views do no use segments to wrap the forms, the segment here can be removed so the form shows with no border.

readthedocsext/theme/templates/organizations/settings/saml.html
stsewd: We aren't allowing updating the integration yet https://github.com/readthedocs/readthedocs-corporate/issues/1781.

readthedocsext/theme/templates/organizations/settings/saml.html
```diff
@@ -0,0 +1,138 @@
+{% extends "organizations/admin/base.html" %}
```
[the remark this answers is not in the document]
stsewd: Do you mean this one https://github.com/readthedocs/ext-theme/blob/main/readthedocsext/theme/templates/organizations/settings/sso_edit.html? I don't see those have much in common.

readthedocsext/theme/templates/organizations/settings/saml.html
stsewd: Getting this now

Screenshot 2024-05-22 at 17-02-52 org - SAML integration - Read the Docs

Already added the `padded` class to the div.

readthedocsext/theme/templates/organizations/settings/saml.html
agjohnson: I noted on that issue, we already allow full control of other auth mechanisms, so it seems worth keeping these consistent instead of special casing SAML auth. We should improve the UX around auth changes in the future more broadly, but SAML is not particularly unique here.

readthedocsext/theme/templates/organizations/settings/saml.html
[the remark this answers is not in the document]
agjohnson: Yup that one. The existing templates reuse base templates frequently to ensure navigation, page headings, and recommended documentation are all the same across subviews.

For example, the environment variable form extends the list template to keep boilerplate identical:

https://github.com/readthedocs/ext-theme/blob/75ccbc474263003ac5cd0e5dcf710a8a85d1d3e0/readthedocsext/theme/templates/projects/environmentvariable_form.html#L1

stsewd:
Still have the problem from https://github.com/readthedocs/ext-theme/pull/348#discussion_r1610725000.

Screenshot 2024-05-23 at 09-50-51 org - Authorization - Read the Docs
Screenshot 2024-05-23 at 09-50-44 two - Authorization - Read the Docs

readthedocsext/theme/templates/organizations/settings/saml.html
agjohnson: You can use a `.ui.basic.segment` around the form if needed. The content border from `.ui.segment` is what is different from all the other forms.

readthedocsext/theme/templates/organizations/settings/saml.html
agjohnson: This feedback still stands for me. I really would like all of our UIs to function in a single, predictable way, and to only maintain one consistent pattern for our UI and UX.

Namely, this form should always just be `{{ form | crispy }}`. Extra fields can be read only using the hardcoded values still.

agjohnson (commented):
This looks close! My earlier feedback on consistent form usage still stands, the primary focus on the new dashboard was to make all of admin views consistent UI/UX, and consistent for maintenance.

agjohnson:
Its fine to start with this, but I'll note that I think we need to come back to this pattern of magic links -- for this and for SSO. I don't think the user is going to know the difference between the two links, and I'm sure if we look at the data users don't use these links. If we are telling owners to invite users, they should probably use the team invite UI.

image

humitos:
> If we are telling owners to invite users, they should probably use the team invite UI.

Agreed. I'd say that once an organization enables SAML, the team invite UI should expose this link somehow instead of being hidden under this specific UI. Ideally, there should be only one UI to invite members and that UI should point the users to the correct place depending on what authorization method they are using.

stsewd:
@agjohnson I'm having a hard time figuring out what exactly is expected here... Something like this?

Screenshot 2024-06-18 at 13-59-41 org - Authorization - Read the Docs

I really don't think we should allow changing the metadata URL yet or delete the integration without a better logic to handle "deactivation", since it's a very destructive change, it affects all users connected to that organization.

Also, this form isn't a model form, so fields are being manually initiated.

agjohnson:
> I really don't think we should allow changing the metadata URL yet or delete the integration without a better logic to handle "deactivation", since it's a very destructive change, it affects all users connected to that organization (and isn't easy to revert).

This is the new pattern I'm saying we skip. We already allow users to change, and potentially break, their project's authentication. This UI does not need to do anything different yet.

I noted this here, as well as a few places above:

- https://github.com/readthedocs/readthedocs-corporate/issues/1781#issuecomment-2125857203

Long term, I agree there is better UX we can add on these forms. But right now, all auth mechanisms, templates, and forms should use the same patterns, and currently the forms for other authentication changes warn the users as they are making changes but do not block the user from making these changes.

If your change is using a single `{{ form | crispy }}`, that is what all the other forms do and is correct. There should be no conditional read only rendering of a form that does not use `{{ form | crispy }}`. We can add warnings in later, that is not a change necessary now either.

> Also, this form isn't a model form, so fields are being manually initiated.

Yup, that should be fine, as long as it is a Form instance, we should always use Crispy to display the form.

stsewd:
> If your change is using a single {{ form | crispy }}, that is what all the other forms do and is correct. There should be no conditional read only rendering of a form that does not use {{ form | crispy }}. We can add warnings in later, that is not a change necessary now either.

There are only two fields in that form (metadata URL and domain), the other fields are just extracted from the settings and aren't part of the Django form. Do you mean creating a form with those fields as read only and manually initiate them?

agjohnson:
No, the read-only fields are okay for now as one-off, read only fields. We should eventually move this to application logic where we can though, this almost certainly doesn't need one-off template logic.

Currently, I am only talking about changing the two fields in the existing form. This form should always be displayed with `{{ form | crispy }}`, and should not use conditional logic to iterate over the form fields and manually display them.

stsewd:
> This form should always be displayed with {{ form | crispy }}, and should not use conditional logic to iterate over the form fields and manually display them.

I was already using form|crispy. I just added the update form, but it still has a conditional under to check if the integration is being created for the first time or not. And the other fields are manually displayed (since they aren't part of the form). Again, I'm not really sure if this is what you are expecting.

Screenshot 2024-07-01 at 16-20-17 new - Authorization - Read the Docs

Screenshot 2024-07-01 at 16-18-57 org - Authorization - Read the Docs

agjohnson:
> I was already using form|crispy

The issue was that the form was only displayed with crispy forms on creation. On update/detail view the form fields were manually generated. What you have here is close to what I was requesting.

We should track a follow up issue to fix the issues here:

- Display issues on spacing and positioning
- Headings are not updated
- Multiple `{{ form | crispy }}` -- all other views only display a single form so the rendered in predictable
- Drop magic links for proper team invites

agjohnson (approved):

PR #350, opened 2024-05-16
allauth: Update templates

stsewd:
Matches the changes from https://github.com/readthedocs/readthedocs.org/pull/11327/


agjohnson:
What does this change look like in the UI? Does it change the display of anything?

readthedocsext/theme/templates/account/email_confirm.html
```diff
@@ -1,4 +1,4 @@
-{% extends "account/base.html" %}
+{% extends "account/base_entrance.html" %}
```
agjohnson, line 1: I'm not familiar with this template, but I assume this is probably a minor change. Assuming the UI hasn't changed at all, this seems safe.

stsewd: Yeah, all templates are now inheriting from this one, which in turn inherits from `allauth/layouts/base.html`

stsewd:
> What does this change look like in the UI? Does it change the display of anything?

All looks the same, there is only one bug that I'm trying to find the root cause

Screenshot 2024-05-20 at 22-36-35 Log in - Read the Docs

agjohnson (approved):

PR #351, opened 2024-05-17
Listing filter UX changes

### LINKED ISSUE #343: Replace search dropdown filter on project/version/etc listing views
Currently, there are a few mock filters that use API driven mechanism to search/etc. These are old implementations and have been updated with some modern work on newer Django filters.

These are the filters that I'm describing:

image

For example, on the organization listing view, the dropdown is a simple filter object and is displayed as normal from the filter menu:

image

These last two filters can be updated to match, which would also remove some of the bad UX around this filter UI.

Places this pattern was used:

- [x] Project listing view
- [x] Project version listing view
- [x] Build listing view

### LINKED ISSUE #344: Filters: show selected values
This is a change I started in 

- https://github.com/readthedocs/ext-theme/pull/250

The current filters show _that_ there is a select filter item, but not what the value is.

agjohnson:
This drops the custom API driven filters and relies on application
FilterSet instances instead, matching the rest of the filter fields.
Because of this, we are able to delete much of the JS and API logic from
the dashboard side.

- Fixes #343
- Fixes #344 
- Closes #250
- Requires https://github.com/readthedocs/readthedocs.org/pull/11330


readthedocsext/theme/templates/includes/filters/form.html
agjohnson, line 1: This diff will be easier to view without whitespace differences. The main changes here where adding a `.ui.sub.header` where there was none before, and adding some classes to improve the UI experience -- scrolling, in-menu search, etc.

humitos (approved):
I'm happy we are relying more on the backend for these things -- easier to maintain, for sure 😉

PR #355, opened 2024-05-22
Impersonate: add link to stop impersonating the user

humitos:
Closes https://github.com/readthedocs/readthedocs-corporate/issues/1789

### DIFF
--- a/readthedocsext/theme/templates/base.html
+++ b/readthedocsext/theme/templates/base.html
@@ -54,8 +54,8 @@
           {% include "includes/utils/messages.html" %}
 
           {% if user.is_impersonate %}
-            <div class="ui small basic fitted center aligned segment">
-              <i class="fad fa-warning violet icon"></i>
+            <div class="ui small center aligned warning message">
+              <i class="fad fa-warning icon"></i>
               Be careful!
               <strong>
                 You are impersonating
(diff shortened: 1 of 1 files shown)


readthedocsext/theme/templates/base.html
```diff
@@ -52,6 +52,19 @@
       <div class="ui very padded container">
         {% block notifications %}
           {% include "includes/utils/messages.html" %}
+
+          {% if user.is_impersonate %}
+            <div class="ui small basic fitted center aligned segment">
+              <i class="fad fa-warning violet icon"></i>
```
agjohnson, line 58: We should use more prominent styles here. This seems to be copied from a notification that intentionally doesn't have strong styling so it doesn't distract users.

In the case of the impersonate link, I like that the block on the dashboard now is prominent -- especially when I forget that I'm impersonating a user or I'm on the wrong tab.

```suggestion
            <div class="ui small center aligned warning message">
              <i class="fad fa-warning icon"></i>
```

image

readthedocsext/theme/templates/base.html
```diff
@@ -52,6 +52,19 @@
       <div class="ui very padded container">
         {% block notifications %}
           {% include "includes/utils/messages.html" %}
+
+          {% if user.is_impersonate %}
```
agjohnson, line 56: I'd also put this at the top of the notification block, before any other messages/notifications. This will make it harder to miss. I don't yet know how my changes for layout shift will affect the display of multiple notifications just yet.

humitos:
I addressed the suggestions already. Merging this so it goes out in today's deploy. I'm happy to make any missing change.

Issue #334, opened 2024-04-17
Build details page: add option to disable line wrapping


humitos:
I'd like to be able to click a checkbox to disable word wrap and show horizontal scroll bars to scroll right/left. There are scenarios where it's a lot easier to find differences between two/multiples lines if the text is aligned between lines.

![Screenshot_2024-04-17_19-31-35](https://github.com/readthedocs/ext-theme/assets/244656/51e9deed-9581-4f81-83cf-a4a36effb6e1)

PR #356, opened 2024-05-23
Fix mobile viewport size on build detail view

agjohnson:
The text menu was not stacking so the viewport width was matching the
menu width. This also removes some unnecessary padding on the wrapping
segment.

### DIFF
--- a/readthedocsext/theme/templates/builds/build_detail.html
+++ b/readthedocsext/theme/templates/builds/build_detail.html
@@ -28,7 +28,7 @@
 {% endblock %}
 
 {% block content %}
-  <div class="ui basic segment padded" data-bind="using: BuildDetailView({}, '{% url "build-detail" build.id %}', '{% url "projects-builds-notifications-list" build.project.slug build.id %}')">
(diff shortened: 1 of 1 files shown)


humitos (approved):
Looks good!

It seems it still breaks in some cases, tho.

Example:
Peek 2024-05-24 17-01

agjohnson:
There are separate, unrelated issues with wrapping that I'll be addressing with something like #334 and some other updates to that.

But it seems you also aren't testing with a proper mobile viewport too. Mobile viewports/DPR can be tested with Firefox responsive design mode. When using a viewport that respects `width=device-width` viewport size, I haven't noticed that overflow issue.

image

The command UI is much more broken on mobile -- command expansion is broken completely there.

humitos:
I'm using that same Firefox feature using the default device I think. I don't remember touching it 🤷‍♂️

Anyways, this PR seems okay to be merged. We can address the other issue in other PRs

agjohnson:
The default device is just a custom window size with DPR: 1. You can use one of the predefined devices to view something closer to what mobile users will see.

PR #362, opened 2024-06-05
Staff: show spam info

humitos:
Allow us to quickly know the spam score of the project and what are their matching rules.

Screenshot_2024-06-05_15-40-34

Screenshot_2024-06-05_15-47-16

### DIFF
--- a/readthedocsext/theme/templates/projects/partials/header.html
+++ b/readthedocsext/theme/templates/projects/partials/header.html
@@ -176,7 +176,11 @@ collapsed, to save visual space and can be expanded with the right label.
                   <div class="ui vertical list">
                     {% for rule in project.spam_rules.all %}
                       <div class="item">
-                        <a href="https://admin.ops.{{ PRODUCTION_DOMAIN }}:10000/admin/spamfighting/spamrule/{{ rule.pk }}/change/">
+                        <a href="https://admin.ops.{{ PRODUCTION_DOMAIN }}:10000/admin/spamfighting/spamrule/{{ rule.pk }}/change/"
+                           {% if rule.description %}
+                             data-content="{{ rule.description }}"
+                           {% endif %}
+                           target="_blank">
                           {{ rule.spam_rule_type|cut:"readthedocsext.spamfighting.rules." }}
                         </a>
                       </div>
--- a/readthedocsext/theme/templates/projects/partials/header.html
+++ b/readthedocsext/theme/templates/projects/partials/header.html
@@ -216,9 +216,6 @@ collapsed, to save visual space and can be expanded with the right label.
               {% for rule in project.spam_rules.all %}
                 <div class="item">
                   <a href="{{ ADMIN_URL }}/spamfighting/spamrule/{{ rule.pk }}/change/"
-                     {% if rule.description %}
-                       data-content="{{ rule.description }}"
-                     {% endif %}
                      target="_blank">
                     {{ rule.spam_rule_type|cut:"readthedocsext.spamfighting.rules." }}
                   </a>
--- a/readthedocsext/theme/templates/projects/partials/header.html
+++ b/readthedocsext/theme/templates/projects/partials/header.html
@@ -202,8 +202,8 @@ collapsed, to save visual space and can be expanded with the right label.
         and also know what rules are matching this project.
       {% endcomment %}
       <div class="content">
-        {% if not ALLOW_PRIVATE_REPOS %}
-          <h3>Spam</h3>
+        {% if not USE_ORGANIZATIONS %}
+          <h3>{% trans "Spam" %}</h3>
           <div class="ui sub header">
             {% trans "Spam score" %}
           </div>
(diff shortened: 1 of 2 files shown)


agjohnson:
This is a great addition, however I would add this into a modal instead. I prefer to always see what the user sees in our UI.

I would follow one of the other templates that invokes a modal and copy the pattern to a button here:
image

There is other information I'd love to have in that modal eventually, and this amount of information will overwhelm the header.

agjohnson (changes requested):
The changes look great and I really like this addition, but I would keep site admin content separate from user content, so that there is only a single UI experience. It should be easy to move this into a list in a model though.

humitos:
Cool, I followed your suggestion and added a modal for this. It gives a lot of room to add more useful debug information here in the future.

Screenshot_2024-06-06_18-28-20

readthedocsext/theme/templates/projects/partials/header.html
```diff
@@ -177,13 +177,61 @@
...
+            <div class="ui sub header">
+              {% trans "Spam matching rules" %}
+            </div>
+            <div class="ui vertical list">
+              {% for rule in project.spam_rules.all %}
+                <div class="item">
+                  <a href="https://admin.ops.{{ PRODUCTION_DOMAIN }}:10000/admin/spamfighting/spamrule/{{ rule.pk }}/change/"
+                     {% if rule.description %}
+                       data-content="{{ rule.description }}"
+                     {% endif %}
```
humitos, line 221: This stopped working once I moved inside a modal. Do you know why?

agjohnson: I'm guessing this is likely because `modal()` is called on the parent element. FUI has some limitations/quirks around floating elements inside floating elements as well, though my guess is JS.

There is also a JS free option for popups/tooltips described in the FUI popup module documentation, which I believe is just CSS. You can try that or using `data-bind="semanticui: { popup: {...}}"`. The implicit, global rules for popups might just not work in modals.

agjohnson:
That looks great! I really like the idea of having a debug menu for projects (and other objects too!)

readthedocsext/theme/templates/projects/partials/header.html
humitos: I'll quickly try that, but I'm not too worried if it doesn't work. We can investigate more on this later

readthedocsext/theme/templates/projects/partials/header.html
```diff
@@ -177,13 +177,61 @@
...
+      <div class="header">
+        {% trans "Debug information" %}
+      </div>
+      {% comment %}
+        Show spam information on community instance.
+        Allows us to quickly understand what's the spam score,
+        and also know what rules are matching this project.
+      {% endcomment %}
+      <div class="content">
+        {% if not ALLOW_PRIVATE_REPOS %}
```
agjohnson, line 205: I don't believe `ALLOW_PRIVATE_REPOS` is included in the context at all, and if it is, it is on a per-view basis.

I always use `USE_ORGANIZATIONS` for this reason. It is exported at a global level, in `readthedocs_context()`. We should really have a more specific global context variable here -- `IS_COMMUNITY`/etc -- but for now `USE_ORGANIZATIONS` is closer, global, and by the time organizations are used on community, we should hopefully have finally added a `IS_COMMUNITY` variable :laughing:

readthedocsext/theme/templates/projects/partials/header.html
```diff
@@ -177,13 +177,61 @@
...
+          <h3>Spam</h3>
```
agjohnson, line 206: Not super important, but for consistency with below:

```suggestion
          <h3>{% trans "Spam" %}</h3>
```

readthedocsext/theme/templates/projects/partials/header.html
```diff
@@ -177,13 +177,61 @@
...
+          </div>
+          <span>{{ project|get_spam_score }} </span>
+          {% if project.spam_rules.exists %}
+            <div class="ui sub header">
+              {% trans "Spam matching rules" %}
+            </div>
+            <div class="ui vertical list">
+              {% for rule in project.spam_rules.all %}
+                <div class="item">
+                  <a href="https://admin.ops.{{ PRODUCTION_DOMAIN }}:10000/admin/spamfighting/spamrule/{{ rule.pk }}/change/"
```
agjohnson, line 218: We should not be exposing this URL in our public repos. It should be a configuration variable.

humitos: We already have this. It's called `ADMIN_URL` 👍🏼

readthedocsext/theme/templates/projects/partials/header.html
humitos: I tried this but I wasn't able to make it work. We can come back later to this, it was just a "nice to have".

humitos:
This PR is ready to merge 👍🏼

agjohnson (approved):

PR #363, opened 2024-06-05
Projects: disable elipsis when user doesn't have permissions

humitos:
Show the elipsis as disabled when the user doesn't have permissions. Follows the same pattern with the documentation/book icon that is disabled when there is no successful build yet.

### Previously

Screenshot_2024-06-05_16-06-53

### New behavior

Screenshot_2024-06-05_16-17-29

### DIFF
--- a/readthedocsext/theme/templates/projects/partials/project_list.html
+++ b/readthedocsext/theme/templates/projects/partials/project_list.html
@@ -59,7 +59,12 @@
           </a>
         </div>
       </button>
+    {% else %}
+      <button class="ui disabled dropdown button">
+        <i class="fa-solid fa-ellipsis icon"></i>
+      </button>
     {% endif %}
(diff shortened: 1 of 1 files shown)


agjohnson (approved):
This is much better! It might be worth considering if we want anything else in this menu and maybe just updating the icon too. Right now the menu only has the configure button inside, so that could just be the top level icon too. If we want more in this dropdown, it's probably worth keeping as a generic menu though.

humitos:
Yeah, I'm sure we will find extra common actions over time and we will need to change this logic to keep making the button available but without the "admin" submenu. However, I don't have any good action to include there in mind yet.

agjohnson:
Same, I thought I might find more. We could reverse this and just make it the admin icon only, instead of ellipsis.

PR #364, opened 2024-06-05
Traffic: show 200 and 404 pages

### LINKED ISSUE #349: Traffic analytics listing view is not showing top pages
The traffic analytics page should show a list of top pages, but is instead showing the placeholder listing content instead. Most likely the queryset context variable name is incorrect or something similar. The page should show a listing instead of the placeholder segment.

humitos:
This PR makes the minimal changes to show the data when it's available. There may be some tweaks and styles we want to improve, but I focus on making the data to be shown first. We can come back later and improve it in a second iteration.

Screenshot 2024-06-05 at 18-40-37 Traffic Analytics - Read the Docs

Closes #349

### DIFF
--- a/readthedocsext/theme/templates/projects/partials/edit/traffic_list.html
+++ b/readthedocsext/theme/templates/projects/partials/edit/traffic_list.html
@@ -18,6 +18,8 @@
 {% block top_menu %}
   {% if not enabled %}
     {% include "organizations/includes/feature_disabled.html" with project=project %}
+  {% else %}
+    {{ block.super }}
   {% endif %}
 {% endblock top_menu %}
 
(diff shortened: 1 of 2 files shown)


agjohnson:
Yeah, to follow the patterns on all the other listing UIs, this should be a filter, not two separate tables. I'd like to return to this to fix this, as I don't want to break consistency in our UI/template code.

agjohnson (approved):
This all looks good, however I was hoping to avoid bringing this pattern into the new dashboard. I've opened an issue to fix this in a consistent way and we should follow up with that change before going much further on this view.

humitos:
> this should be a filter, not two separate tables

Sounds like a good idea to me 👍🏼

PR #365, opened 2024-06-05
Search Analytics: show results and queries

humitos:
Improve how we show the results the query returned and also how many times users have searched for that particular query)
Screenshot_2024-06-05_18-52-51

### DIFF
--- a/readthedocsext/theme/templates/projects/partials/edit/search_query_list.html
+++ b/readthedocsext/theme/templates/projects/partials/edit/search_query_list.html
@@ -54,7 +54,7 @@
 {% block list_item_right_menu %}
   <div class="ui right floated text menu">
     <div class="item">
-      {{ object.1 }}
+      {% blocktrans trimmed count num=object.1 %}{{ num }} query{% plural %}{{ num }} queries{% endblocktrans %}
     </div>
   </div>
 {% endblock list_item_right_menu %}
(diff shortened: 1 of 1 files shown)


readthedocsext/theme/templates/projects/partials/edit/search_query_list.html
```diff
@@ -54,17 +54,19 @@
 {% block list_item_right_menu %}
   <div class="ui right floated text menu">
     <div class="item">
-      {% blocktrans trimmed count num=object.1 %}
-        {{ num }} result
-      {% plural %}
-        {{ num }} results
-      {% endblocktrans %}
+      {{ object.1 }}
```
agjohnson, line 57: We should reuse the same translation logic but update the nouns if these are wrong. I wasn't sure what the numbers were in the UI without this.

That is `4` is unclear, but `4 queries` is really clear.

readthedocsext/theme/templates/projects/partials/edit/search_query_list.html
```diff
@@ -54,17 +54,19 @@
...
     </div>
   </div>
 {% endblock list_item_right_menu %}
 
 {% block list_item_header %}
   {{ object.0 }}
+  <div class="sub header">
+    <label>
```
agjohnson, line 65: Label is not needed here, it can be dropped.

agjohnson (approved):

PR #367, opened 2024-06-05
Fix distribution not including some templates

stsewd:
For some weird reason, graft isn't including some directories in the final distribution (organizations/settings for example). Using recursive-include fixes that problem,
not sure if this is a bug with setuptools or something else.


agjohnson:
I bet it's because those paths are symlinks, it probably drops those.

agjohnson (approved):

PR #368, opened 2024-06-06
Add project: show more `yaml` examples

### LINKED ISSUE #184: Add JS for swapping examples on project creation form

humitos:
I used tabs to show different examples of YAML files, as suggested in #184. It looks great for an initial interactive approach.

I like the tabs over the dropdown because it gives the user a clear idea there are more than Sphinx and MkDocs tools supported, which is a good marketing strategy, in my opinion.

It also adds an "Others" option that's more generic and shows how to use `build.commands` in a generic way: install dependencies, build, copy assets under `$READTHDOCS_OUTPUT` directory.

Peek 2024-06-06 13-33

* Closes #184
* Closes https://github.com/readthedocs/readthedocs.org/issues/10352

### DIFF
--- a/readthedocsext/theme/templates/projects/import_config.html
+++ b/readthedocsext/theme/templates/projects/import_config.html
@@ -1,9 +1,9 @@
 {% extends "projects/import_base.html" %}
 {% load i18n %}
 
-{% block project_add_content_subheader %}
+{% block project_add_subheader %}
   {% trans "Add a configuration file to your project" %}
-{% endblock project_add_content_subheader %}
+{% endblock project_add_subheader %}
 
 {% block project_add_content_classes %}ui fourteen wide tablet twelve wide computer column{% endblock %}
 
@@ -12,43 +12,33 @@
     {% blocktrans trimmed %}
       A <code>.readthedocs.yaml</code> configuration file is required at the root of your repository in order to build your documentation.
     {% endblocktrans %}
-
-    <a href="https://docs.readthedocs.io/page/config-file/index.html" target="_blank">
-      {% trans "Learn how to add a configuration file to your project." %}
-    </a>
   </div>
 
-  <div class="ui basic fitted right aligned segment">
-    <span>{% trans "Example configuration for:" %}</span>
-    <div class="ui inline dropdown" data-bind="semanticui: { dropdown: {}}">
-      <input type="hidden" name="tool" value="sphinx">
-      <span class="text">Sphinx</span>
-      <i class="dropdown icon"></i>
-      <div class="menu">
-        <div class="item" data-value="sphinx">Sphinx</div>
+  <p>
 
-        {% comment %}
-          Adding a second option here will require a bit of JS or FUI:
-          https://github.com/readthedocs/ext-theme/issues/184
-        {% endcomment %}
+    Here you have some simple examples for the most common documentation tools.
+    If you are using a different tool, you can read our documentation to
+    <a href="https://docs.readthedocs.io/page/config-file/index.html" target="_blank">learn how to write your own</a>.
+  </p>
 
-        {# The `actionable` class here prevents the select from selecting the text #}
-        <a class="actionable item" href="https://docs.readthedocs.io/page/config-file/index.html#getting-started-with-a-template" target="_blank">
-          {% trans "See more examples" %}
-        </a>
+  <div class="ui top attached tabular menu">
+    <div class="active item" data-tab="sphinx">Sphinx</div>
+    <div class="item" data-tab="mkdocs">MkDocs</div>
+    <div class="item" data-tab="pelican">Pelican</div>
+    <div class="item" data-tab="docusaurus">Docusaurus</div>
+    <div class="item" data-tab="jekyll">Jekyll</div>
+    <div class="item" data-tab="others">Others</div>
+  </div>
+  <div class="ui bottom attached active tab segment" data-tab="sphinx">
+    <div class="ui mini padded inverted scrolling segment">
+      <div class="ui top attached label">
+        <code>.readthedocs.yaml</code>
       </div>
-    </div>
-  </div>
-
-  <div class="ui mini padded inverted scrolling segment">
-    <div class="ui top attached label">
-      <code>.readthedocs.yaml</code>
-    </div>
-    <a class="ui top right attached icon label" href="#" data-clipboard-target="#project-create-sample-sphinx">
-      <i class="fas fa-copy icon"></i>
-    </a>
-    <code class="highlight">
-      <pre id="project-create-sample-sphinx">
+      <a class="ui top right attached icon label" href="#" data-clipboard-target="#project-create-sample-sphinx">
+        <i class="fas fa-copy icon"></i>
+      </a>
+      <code class="highlight">
+              <pre id="project-create-sample-sphinx">
 # Read the Docs configuration file
 # See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
 
@@ -59,7 +49,7 @@ version: 2
 build:
   os: ubuntu-22.04
   tools:
-    python: "3.11"
+    python: "3.12"
     # You can also specify other tool versions:
     # nodejs: "19"
     # rust: "1.64"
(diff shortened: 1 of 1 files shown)


readthedocsext/theme/templates/projects/import_config.html
```diff
@@ -12,43 +12,33 @@
...
-    <div class="ui top attached label">
-      <code>.readthedocs.yaml</code>
-    </div>
-    <a class="ui top right attached icon label" href="#" data-clipboard-target="#project-create-sample-sphinx">
-      <i class="fas fa-copy icon"></i>
-    </a>
-    <code class="highlight">
-      <pre id="project-create-sample-sphinx">
+    Here you have some simple examples for the most common documentation tools.
+    If you are using a different tool, you can read our documentation to
```
agjohnson, line 20: This UI is already obvious, this paragraph isn't adding much information. As a general rule we should avoid using prose to _explain our UI_ -- it takes up space and is distracting to users. Prose content can be helpful to describe edge cases.

For example, this accomplishes the same in two words:

image

I'd prefer retaining this and the dropdown, but we really only need add an `Example configuration` header back instead of this paragraph

readthedocsext/theme/templates/projects/import_config.html
```diff
@@ -12,43 +12,33 @@
...
+  </p>
+
+  <div class="ui top attached tabular menu">
+    <div class="active item" data-tab="sphinx">Sphinx</div>
+    <div class="item" data-tab="mkdocs">MkDocs</div>
+    <div class="item" data-tab="pelican">Pelican</div>
+    <div class="item" data-tab="docusaurus">Docusaurus</div>
+    <div class="item" data-tab="jekyll">Jekyll</div>
+    <div class="item" data-tab="others">Others</div>
+  </div>
```
agjohnson, line 31: Tabs work okay here, but I did make this a dropdown on purpose. Is there a reason we want tabs instead?

readthedocsext/theme/templates/projects/import_config.html
```diff
@@ -12,43 +12,33 @@
...
-      <i class="fas fa-copy icon"></i>
-    </a>
-    <code class="highlight">
-      <pre id="project-create-sample-sphinx">
+    Here you have some simple examples for the most common documentation tools.
+    If you are using a different tool, you can read our documentation to
+    <a href="https://docs.readthedocs.io/page/config-file/index.html" target="_blank">learn how to write your own</a>.
+  </p>
+
+  <div class="ui top attached tabular menu">
```
agjohnson, line 24: This needs to be stackable to work okay on narrow viewports, tablet, and mobile:

```suggestion
  <div class="ui top attached stackable tabular menu">
```

readthedocsext/theme/templates/projects/import_config.html
```diff
@@ -12,43 +12,33 @@
...
+    <div class="active item" data-tab="sphinx">Sphinx</div>
+    <div class="item" data-tab="mkdocs">MkDocs</div>
+    <div class="item" data-tab="pelican">Pelican</div>
+    <div class="item" data-tab="docusaurus">Docusaurus</div>
+    <div class="item" data-tab="jekyll">Jekyll</div>
+    <div class="item" data-tab="others">Others</div>
```
agjohnson, line 30: This needs localization:

```suggestion
    <div class="item" data-tab="others">{% trans "Others" %}</div>
```

readthedocsext/theme/templates/projects/import_config.html
```diff
@@ -12,43 +12,33 @@
...
+  <div class="ui bottom attached active tab segment" data-tab="sphinx">
+    <div class="ui mini padded inverted scrolling segment">
+      <div class="ui top attached label">
+        <code>.readthedocs.yaml</code>
+      </div>
+      <a class="ui top right attached icon label" href="#" data-clipboard-target="#project-create-sample-sphinx">
+        <i class="fas fa-copy icon"></i>
+      </a>
+      <code class="highlight">
+              <pre id="project-create-sample-sphinx">
```
agjohnson, line 41: This looks like DjHTML didn't run? The indentation is way off

agjohnson (commented):
This is looking good! I think I might want to retain the dropdown over tabs if possible, the length of tabs and count of tabs will cause this to wrap in awkward ways. The dropdown was used to avoid these two issues.

readthedocsext/theme/templates/projects/import_config.html
humitos: Yes. I explained this in the description. It gives a more broad view to the user about what they could use. It works as discover marketing, and it looks great to me.

readthedocsext/theme/templates/projects/import_config.html
humitos: It runs on pre-commit and and I always accept all the changes, so 🤷‍♂️

readthedocsext/theme/templates/projects/import_config.html
agjohnson: The tabs are likely to run into display issues very quick. Make sure you have tested this on multiple display widths. I prefer the dropdown over tabs for this very specific reason.

The dropdown can still activate tabbed segments, like this: 

https://github.com/readthedocs/readthedocs.org/issues/10352#issuecomment-2153444971

readthedocsext/theme/templates/projects/import_config.html
agjohnson: When this happens, it usually means it is confused about the indentation level because an error was made at some point.

readthedocsext/theme/templates/projects/import_config.html
```diff
@@ -12,43 +12,33 @@
...
-      <i class="fas fa-copy icon"></i>
-    </a>
-    <code class="highlight">
-      <pre id="project-create-sample-sphinx">
+    Here you have some simple examples for the most common documentation tools.
+    If you are using a different tool, you can read our documentation to
+    <a href="https://docs.readthedocs.io/page/config-file/index.html" target="_blank">learn how to write your own</a>.
+  </p>
+
+  <div class="ui top attached stackable tabular menu">
```
agjohnson, line 24: I was wrong here, we want `wrapping` here. It's a relatively new class, but instead of stacking all of the tabs vertically, it first tries to fit the tabs horizontally

image

This looks much better than vertically stacking all of the menu items.

image

readthedocsext/theme/templates/projects/import_config.html
humitos: OK. Let's remove the tabs then and only keep the dropdown. What's the right way to do that?

humitos: I was able to do it already 👍🏼

readthedocsext/theme/templates/projects/import_config.html
humitos: It seems it's a valid file for djhtml:

```
▶ djhtml readthedocsext/theme/templates/projects/import_config.html 
0 templates have been reindented.
1 template was already perfect!
▶ djhtml -t 2 readthedocsext/theme/templates/projects/import_config.html
0 templates have been reindented.
1 template was already perfect!
```

agjohnson: Yeah, it's not that it djhtml is erroring out, it is just confused and applying the wrong formatting. Two things can be the case I've found: either there is an HTML error like unclosed tags somewhere, or djhtml is just ignoring the line. I think it's likely the second.

These errors seem to be from editing, djhtml doesn't reindent comment blocks, and apparently `pre` blocks.

This is another case for `djlint`, which does what I would expect on this file. djhtml helps with formatting, but tends to be nondeterministic with it's rules. If you edit the indenting and rerun djhtml, it won't correct these lines.

agjohnson: I put this PR up as a test:

- https://github.com/readthedocs/ext-theme/pull/371

I suppose we could switch to djlint finally and use pre-commit to control which files are being reformatted/linted -- not all files at once, but file by file as we update them.

For this PR, and until we switch, be aware that editor spacing leaks through djhtml in a number of cases.

humitos: Fixed in [9bbca23](https://github.com/readthedocs/ext-theme/pull/368/commits/9bbca2338681edb0ba166f0f208a79a66361d7ea)

agjohnson (approved):
This looks great. I think we can tune these examples quite a bit once we have more examples, but they are good for now.

PR #369, opened 2024-06-06
Add project: highlight yaml examples

humitos:
I used [`highlight.js`](https://highlightjs.org/) to highlight the examples when importing a project. It requires importing a CSS file that I'm not sure where to put it yet.

https://cdnjs.cloudflare.com/ajax/libs/highlight.js/11.9.0/styles/dark.min.css

Besides, I need to know where to place the small chunk of JS code that I added to make this work.

Screenshot_2024-06-06_14-02-30

@agjohnson I would appreciate some directions here to clean up and finish this work.

Based on #368


humitos:
I moved the code where I think it belongs. The only missing piece here is understanding how to include my CSS stylesheet file. Other than that, it seems that everything works as expected 👍🏼

humitos:
Boom! I figured it out!

readthedocsext/theme/templates/projects/import_config.html
```diff
@@ -37,8 +39,8 @@
       <a class="ui top right attached icon label" href="#" data-clipboard-target="#project-create-sample-sphinx">
         <i class="fas fa-copy icon"></i>
       </a>
-      <code class="highlight">
-              <pre id="project-create-sample-sphinx">
+      <pre>
+          <code  id="project-create-sample-sphinx" class="language-yaml highlight">
```
humitos, line 43: `highlight.js` requires this to be `pre > code` and not the other way around 🤷🏼

agjohnson:
We are already using Pygments on the server side in other places. We should very likely continue with that pattern, I don't feel we should have two ways of highlighting.

You can see the integration HTTP exchange UI, templates, and views for examples of this. I'm guessing it will plug in easily here as well.

agjohnson:
Noted on the original issue too, the plan when this UI was first added was to come back and switch the examples with tabbed menus and use Pygments for highlighting:

- https://github.com/readthedocs/ext-theme/issues/184

agjohnson (changes requested):
Probably should have noted my feedback in review here. I think Pygments is the better option here, it doesn't need anything horribly new and will match where we've used Pygments in other UIs.

humitos:
OK. The original issue said that either worked: pygments or a front-end solution. I will come back to this once I have some extra time and check the pygments solution. However, I will de-prioritize this work for now.

agjohnson:
Adding a front end dependency works if pygments doesn't work for some reason, but it helps to keep patterns the same and dependencies minimal on the front end side if we can.

PR #371, opened 2024-06-10
Try djlint instead of djhtml

### LINKED ISSUE #329: Lint: try swapping to djlint
We are currently using djhtml here for html formatting. This package only does part of what djlint does, but was providing nicer html formatting when I first tried both. I did try djlint recently and the formatting was mostly okay. I think we could swap to djlint if we wanted to, as djlint does have some more checks on templates.

Some things to evaluate:

- [ ] Template indentation and formatting looks okay/better with djlint?
- [ ] djlint catches easy to miss things like multiline, single line comments (`{# #}`), maybe HTML errors?
- [ ] djlint catches/fixes other issues?

This will be a disruptive change, so would hopefully be an obvious choice after playing with djlint a bit more.

agjohnson:
Djhtml is inconsistent and doesn't provide any linting checks. Running djlint instead, and through pre-commit on a per-file basis, gives a fairly easy way to migrate to stronger formatting/linting on template code.

- Fixes #329

### DIFF
--- a/.djlintrc
+++ b/.djlintrc
@@ -1,4 +1,5 @@
 {
   "indent": "2",
-  "preserve_blank_lines": true
+  "preserve_blank_lines": true,
+  "ignore": "H006"
 }
--- a/.djlintrc
+++ b/.djlintrc
@@ -1,0 +1,4 @@
+{
+  "indent": "2",
+  "preserve_blank_lines": true
+}
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -26,12 +26,10 @@ repos:
       - id: destroyed-symlinks
       - id: end-of-file-fixer
         exclude: "^readthedocsext/theme/static/.*$"
-  - repo: https://github.com/rtts/djhtml
-    rev: "3.0.6"
+  - repo: https://github.com/djlint/djLint
+    rev: v1.34.1
     hooks:
-      - id: djhtml
-        entry: djhtml --tabwidth 2
-        files: '^readthedocsext/theme/templates/.*\.html'
+      - id: djlint-django
   - repo: local
     hooks:
       - id: check-build-assets
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -30,6 +30,7 @@ repos:
     rev: v1.34.1
     hooks:
       - id: djlint-django
+      - id: djlint-reformat-django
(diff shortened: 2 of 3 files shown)


agjohnson:
If we want to enable this, it should only run on future edits to templates. The reformatting is more consistent than djhtml, but the linting rules are quite helpful and enforce our development rules. However, this would involve fixing some unrelated template issues fairly frequently. Here's an example of fixes:

```
readthedocsext/theme/templates/includes/footer.html

T002 75:34 Double quotes should be used in tags. {% trans 'Version' %
H022 77:16 Use HTTPS for external links. <a href="http://docs
T002 86:34 Double quotes should be used in tags. {% trans 'Language'
D018 90:18 (Django) Internal links should use the {% url ... %} pattern. <form action="/i18n/
H006 144:16 Img tag should have height and width attributes. <img class="ui tiny
H006 153:16 Img tag should have height and width attributes. <img class="ui tiny
H006 162:16 Img tag should have height and width attributes. <img class="ui tiny
H006 171:16 Img tag should have height and width attributes. <img class="ui tiny
H006 180:16 Img tag should have height and width attributes. <img class="ui tiny
H006 189:16 Img tag should have height and width attributes. <img class="ui tiny
```

We could start with reformatting and introduce rules next, tune rules back, or just accept that the edits are all really minor and go 100% on djlint from the start.

ericholscher:
I'm 👍 on moving forward here. It would be really nice if there was a black-style auto-formatter, so we could format everything once, and then move forward with stricter rules. I really dislike hitting random linter issues on every PR.

agjohnson:
To clarify, djhtml is already doing autoformatting on templates, it just isn't as strict as djlint. djlint has more opinionated formatting and has the option of strict rule checks in addition, which is the stronger upgrade. The autoformatting will only change slightly -- indentation is more opinionated and attribute wrapping is more frequent.

I might be wrong about how this will work, but if we're okay fixing template issues a file at a time, I think the configuration here works. That is, when a template file is touched next, djlint will update the whole file. We will have to tune ignore rules as we go, which is fine, and perhaps just punt on some rules for now too, which I would say is also fine.

ericholscher:
Yea, I'm fine being pretty aggressive with ignore rules, and then removing them as a later step.

ericholscher (approved):
I'm fine to this try 👍

PR #333, opened 2024-04-17
Dashboard: promote features on sidebar

humitos:
Initial iteration to promote some features to our users.

Screenshot_2024-04-17_17-52-40

Closes #103
Requires https://github.com/readthedocs/readthedocs.org/pull/11287

Issue #103, opened 2022-05-12
Add initial dashboard page sidebar content


agjohnson:
I'd love to add a marketing/onboarding/education block in our UI, probably on the main dashboard page. I couple examples I like:

- Stripe has a small banner ad sized element that they use for new feature promotion. I think we could easily author something similar, but perhaps more in HTML. A background image draws in attention here, which I really like
- GitHub has this:

![image](https://user-images.githubusercontent.com/1140183/168120254-0d5fd97a-050f-4c4a-8e62-9c76e8c0c378.png)

It's blog post then changelog specifically. Not just blog posts. I'd probably lean towards specific user features instead of just changelog though. We don't summarize changelogs anyways.

agjohnson:
This is a good issue to jump into @humitos I'm throwing it your way.

I noted some options, and the commercial sidebar has some placeholder content for creating an organization (this is not a good fit here).

humitos:
With this code:

```html
        <div class="ui card">
          <div class="image">
            <img src="https://about.readthedocs.com/images/read-the-docs-loves-ruby.jpg">
          </div>
          <div class="content">
            <a class="header" target="_blank" href="https://about.readthedocs.com/blog/2024/03/read-the-docs-loves-ruby/">Read the Docs ❤️ Ruby</a>
            <div class="description">
              We are happy to announce support for Ruby is now available as an option as pre-defined building tools (<code>build.tools</code>).
              Ruby is now officially supported to build project documentation on Read the Docs 🎉
            </div>
          </div>
        </div>
```

I was able to generate the following:

![Screenshot_2024-04-15_12-00-57](https://github.com/readthedocs/ext-theme/assets/244656/e9a7076f-307c-4221-a1c7-21ac0d5a4470)


From here, I think we can create a small HTML template that renders data from the database:

* show only one at a time
* add a small X to close the card and save a cookie to not show it again to the same user
* expect `image` (optional), `link`, `header` and `description`
```python
class DashboardPromotion(TimeStampedModel):
  image = models.ImageField()
  link = models.URLField()
  header = models.CharFiedl()
  description = models.TextField()
  expires = models.DateField()   # when stop showing this promotion
```

Then we can simply manage these promotions from the Django Admin, without building a UI or similar.

@agjohnson what do you think about this approach?

agjohnson:
[...]

PR #377, opened 2024-06-11
Announcements: allow user to close feature announcements

humitos:
Add a small "X" next in the right top corner of the announcement to let the users to close the announcement. It uses a cookie to avoid showing it next time they load the page.

Peek 2024-06-11 17-08

Related
* #333
* #103


readthedocsext/theme/templates/projects/partials/dashboard/traffic-analytics.html
```diff
@@ -1,21 +1,24 @@
...
-    <div class="content">
-      <a class="header" target="_blank" href="{{ url }}">{% trans "Prioritize your most-read pages" %}</a>
-      <div class="description">
-        {% blocktrans %}
-          Use <strong>traffic analytics</strong> to know <em>which</em> documents your users are reading.
-          This allows you to understand how your documentation is being used and learn more about your audience.
-          Then you can prioritize your efforts on expanding and updating pages people are reading most.
-        {% endblocktrans %}
+  <div data-bind="using: new ProjectAnnouncementView('traffic-analytics')">
+    <div class="ui card" data-bind="visible: !closed()">
```
agjohnson, line 5: I think what you have here might be prone to flashing content to the user -- the content will start visible and hide once the JS loads (and technically  display again if the user hasn't dismissed?). I might be wrong here though, you can test by slowing down your network connection in Firefox.

There is a helper class that you can use for these sorts of cases. It just applies `display: none`, but avoids flashing the content to the user on load, before the JS hides the element. It uses some inverted logic of sort though, which would look like: remove the `hidden` class if the notification should be visible.

There are some examples of this in the templates, but I think it is something like this:

```html
    <div class="ui card ko hidden" data-bind="css: {hidden: closed}">
```

agjohnson (approved):
This looks great!

I am wondering now if we do more display time logic on the front end side. That is, with templates deciding what announcement to show, we can back down to the next notification. With front end logic in addition, templates will decide to show a notification but then it will just be hidden instead of backing down to another notification.

readthedocsext/theme/templates/projects/partials/dashboard/traffic-analytics.html
humitos: Yes. Actually, I have this exact issue and I commented on Slack about this. It works fine 👍🏼

humitos:
I will update this PR to add the same X button to all the other cards as well. I added only on "traffic analytics" just as a test to get some feedback.

> I am wondering now if we do more display time logic on the front end side. That is, with templates deciding what announcement to show, we can back down to the next notification. With front end logic in addition, templates will decide to show a notification but then it will just be hidden instead of backing down to another notification.

I'm not sure to understand this. Do you mean returning the HTML for all the notifications and decide which one to show on the front-end side? What's that logic? I'm happy to discuss this more and include it in another PR.

humitos:
I updated the PR to reflect this. Let me know if it's ready to merge 👍🏼

agjohnson:
> I will update this PR to add the same X button to all the other cards as well. I added only on "traffic analytics" just as a test to get some feedback.

:+1: Yeah figured this was just a start.

> I'm not sure to understand this. Do you mean returning the HTML for all the notifications and decide which one to show on the front-end side? What's that logic? I'm happy to discuss this more and include it in another PR.

I haven't thought of how to solve this. We could use front end logic for 100% of the announcement decision, or use a proper cookie and use the cookie in the backend/template when deciding which announcement to show.

The goal would be to ensure the user always has an announcement available to show. With mixed template and front end logic deciding if an announcement shows, users will end up pinned to an announcement they have hidden. These are onboarding steps, so this would effectively end the onboarding progression for the user.

humitos:
OK, I quickly opened https://github.com/readthedocs/ext-theme/issues/378 to discuss this more. I'm merging this chunk of work for now.

PR #384, opened 2024-06-13
Add proxito error templates

stsewd:
Header and footer have references to URLs from our app, so they are completely overridden. In .org we have all proxito templates in the /errors/404 directory, so I'm just creating a symlink from /errors/proxito/404.

image


readthedocsext/theme/templates/errors/proxito/404/base.html
agjohnson, line 1: I was aiming to use this directory structure as well :+1:

PR #388, opened 2024-06-18
Use custom 401 page for proxito

stsewd:
Similar to https://github.com/readthedocs/ext-theme/pull/384, this is just a placeholder for now.


readthedocsext/theme/templates/errors/401.html
agjohnson, line 1: If we can't yet move `401.html` to `errors/proxito/401.html` this path is okay.

agjohnson (approved):

PR #392, opened 2024-06-19
Fix get_account_username for SAML

stsewd:
This breaks our tests, since SAML's extract_common_fields is used in a different way than the other providers.


readthedocsext/theme/templates/socialaccount/login.html
```diff
@@ -4,7 +4,7 @@
 
 {% block head_title %}{% trans "Sign in" %}{% endblock %}
 
-{% block content_social %}
+{% block content %}
```
stsewd, line 7: This was wrong, the base template from django-allauth this uses names this block `content`.

readthedocsext/theme/templatetags/ext_theme_tags.py
```diff
@@ -108,6 +108,10 @@ def get_account_username(socialaccount):
     name of the human attached to said account.
     """
     provider = socialaccount.get_provider()
+    # SAML doesn't store the username in extra_data,
+    # and extract_common_fields doesn't expect a dictionary, but an object.
```
humitos, line 112: Is there any reason to not follow what the other providers do and have this standardized?

stsewd: Nope, that's how it works in allauth.

humitos (approved):

PR #394, opened 2024-06-20
Use sign up instead of sign in when signing up

stsewd:
Currently this reads "Sign in" when the user is signing up. I'm guessing this was a mistake.

The sign in page still says "Sign in with".

Screenshot 2024-06-20 at 11-45-34 Sign up - Read the Docs


stsewd:
We could also change the `Log in using:` text next to the tabs to `Sign up using:`

ericholscher (approved):

PR #396, opened 2024-06-21
Help topics on base dashboard

humitos:
This PR copies the same pattern we have in other places (eg. project's settings) using "Help topics". I copied it from there because it renders nicer than the one we had and also to keep consistency between them all.

### DIFF
--- a/readthedocsext/theme/templates/projects/base_dashboard.html
+++ b/readthedocsext/theme/templates/projects/base_dashboard.html
@@ -36,31 +36,26 @@
           {% include promotion %}
         {% endif %}
 
-        <div class="basic card">
-          <div class="content">
-            <a class="header" href="https://docs.readthedocs.io/en/stable/">
-              {% trans "Learn more" %}
-            </a>
-            <div class="meta">{% trans "Some resources from our documentation" %}</div>
-            <div class="description">
-              <div class="ui mini header">{% trans "Tutorials" %}</div>
-              <div class="ui bulleted list">
-                {% include "includes/elements/link.html" with url="https://docs.readthedocs.io/page/tutorial/index.html" text="Getting started" is_external=True class="item" %}
-                {% include "includes/elements/link.html" with url="https://docs.readthedocs.io/page/intro/getting-started-with-sphinx.html" text="Getting started with Sphinx" is_external=True class="item" %}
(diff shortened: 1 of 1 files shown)


readthedocsext/theme/templates/projects/base_dashboard.html
```diff
@@ -36,31 +36,26 @@
...
-              <div class="ui mini header">{% trans "Reference" %}</div>
-              <div class="ui bulleted list">
-                {% include "includes/elements/link.html" with url="https://docs.readthedocs.io/page/config-file/index.html" text="Configuration file reference" is_external=True class="item" %}
-                {% include "includes/elements/link.html" with url="https://docs.readthedocs.io/page/build-customization.html" text="Build process customization" is_external=True class="item" %}
-              </div>
-            </div>
+          <div class="ui mini header">{% trans "Reference" %}</div>
+          <div class="ui bulleted list">
+            {% include "includes/elements/link.html" with url="https://docs.readthedocs.io/page/config-file/index.html" text="Configuration file reference" is_external=True class="item" %}
+            {% include "includes/elements/link.html" with url="https://docs.readthedocs.io/page/build-customization.html" text="Build process customization" is_external=True class="item" %}
```
agjohnson, line 54: This looks great, just noting because I saw it somewhere else: I like this pattern of using a heading to describe the type of content we're linking to and then less bullet points. This helps with content wrapping on the links.

That is, originally we had added links prefixed with content type like:

- Tutorial: Getting started with Sphinx
- Tutorial: Getting started with RTD
- Blog post: Some heading here

But the prefix wraps everything awkward and we get the same from a more structured list like we have here:

- Tutorials:
  - Getting started with Sphinx
  - Getting started with RTD
- Blog posts:
  - Some page title
  - Some other page title

Nothing to change here, just noting the change in this PR is a good one to center on.

agjohnson (approved):

readthedocsext/theme/templates/projects/base_dashboard.html
humitos: Ah, yeah, I found the "Prefix: <title>" pattern in other places. I kept the same pattern there when editing links, but I'm 👍🏼 on using small headings instead of these prefixes 👍🏼

PR #398, opened 2024-06-21
Update links from the footer

humitos:


readthedocsext/theme/templates/includes/footer.html
```diff
@@ -59,6 +59,8 @@ <h4 class="ui inverted sub header">About us</h4>
               {# TODO move this out of the footer and drop this menu from this block/column  #}
               <h4 class="ui inverted sub header">Tools</h4>
               <a class="item" href="https://about.readthedocs.com/tools/sphinx/">Sphinx</a>
+              <a class="item" href="https://about.readthedocs.com/tools/mkdocs/">MkDocs</a>
+              <a class="item" href="https://about.readthedocs.com/tools/jupyter-book/">Jupyter Book</a>
```
ericholscher, line 63: I think now these are in the header, we can probably drop them down here?

ericholscher (approved):

readthedocsext/theme/templates/includes/footer.html
humitos: They are in the header on about.readthedocs.com, but this is the footer from the new application dashboard.

PR #403, opened 2024-06-29
Replace project avatars

### LINKED ISSUE #131: Replace Dicebear with static generated images

agjohnson:
- [x] Fix URL for web component styles and avatar image
- [x] Create a 100x100px image, mapped to our current color scheme colors on the dashboard. This creates a lot of nice variance in colors and matches the look and feel in addition.
- [x] Resize to something easy for bitwise hashing math, 128x128?
- [x] Find the right pattern for avoiding inline styles and background positioning
- [x] Fix positioning on the project header, it's out of place now

----

- Resolves #131


agjohnson:
Image

agjohnson:
I worked through patterns for attributes, using a static URL for the image, updated the underlying sprite, and fixed an issue on the project header.

With the new sprite, the dashboard looks like this now:

image

The image is a palette mapped image, using colors from the FUI base color scheme. If this looks too busy, I can try a palette mapped image using less of the colors.

agjohnson:
I like the look of the 8x8 pixel avatar, but would say the 4x4 looks more like it is a generated icon. The color limited color scheme works much nicer too, all the colors are prominent in the dashboard already.

I think my favorite is 4x4 pixels, limited FUI color palette:

image

humitos (commented):
Looks good to me. The only thing I would change is the background image used. I'd propose to generate one ourselves in some random way by taking a color palette from https://colorhunt.co/ that looks nice.

agjohnson:
I am already using the same color palette as FUI. What is the concern around generating an image vs processing a photo?

ericholscher (approved):
🚢

PR #404, opened 2024-07-02
Improve page load performance on listing views

### LINKED ISSUE #49: Replace usage of `is_admin:project` filter in object lists

### LINKED ISSUE #128: Page responses times

agjohnson:
- Requires https://github.com/readthedocs/readthedocs.org/pull/10978
- Fixes #128
- Fixes #49


agjohnson:
I'm calling this ready, but we're blocked here on https://github.com/readthedocs/readthedocs.org/pull/10978

agjohnson:
@humitos ping, this could use review still

humitos (approved):

PR #414, opened 2024-07-04
Build popups: simplification

humitos:
This commit simplifies the build chip a little:

- removes the `text` for the list item since it always repeat the name of the project and there is no need to show it so many times
- removes the name of the project from the popup title for the same reason
- adds the text "Build" to the title since it conveys a more clear meaning
- makes the build number clickable, so we can go to the build details page from the chip's title
- adds a link to the commit URL

|Old|New|
|--- | --- |
| Screenshot_2024-07-04_13-58-38 | Screenshot_2024-07-04_13-58-02 |
| Screenshot_2024-07-04_13-58-29 | Screenshot_2024-07-04_13-57-53 |

### DIFF
--- a/readthedocsext/theme/templates/includes/elements/chips/build.html
+++ b/readthedocsext/theme/templates/includes/elements/chips/build.html
@@ -55,7 +55,7 @@
   {% if build.commit %}
     <span>
       <i class="fad fa-code-commit icon"></i>
-      {{ build.commit|slice:"0:8" }}
+      <a href="{{ build.get_commit_url }}">{{ build.commit|slice:"0:8" }}</a>
     </span>
   {% endif %}
   <div class="right floated">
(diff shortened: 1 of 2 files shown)


readthedocsext/theme/templates/includes/elements/chips/build.html
```diff
@@ -19,7 +19,9 @@
     instance was deleted at some point. The version name is stored on the build
     in this scenario.
   {% endcomment %}
-  {% firstof text build.get_version_name %}
+  {% if not notext %}
+    {% firstof text build.get_version_name %}
+  {% endif %}
```
agjohnson, line 24: This does need some text, for a couple reasons:

- This is the main content of the `.ui.label` and it looks too padded/empty without this
- The other popup cards use all this same pattern - icon, main content, and detail content

image

While I agree repeating the project name isn't helpful, it is consistent. For comparison, CircleCI also is redundant when describing builds:

image

This states the project name as the word `Build` looks even more redundant and isn't useful information in addition:

image

So, I think project name is best, maybe `Build` is okay, but this text should be _something_.

We should revisit patterns here later. I have a pile of work on popup cards in the milestone https://github.com/readthedocs/ext-theme/milestone/6

PR #370, opened 2024-06-06
Project: use `zh-cn` if the language code is the unsupported `zh`

humitos:
Simple and small fix for this issue.
We should still decide what to do at the db level still.

Closes #358

PR #417, opened 2024-07-09
Project: use a custom template filter to avoid zh languages

humitos:
We keep returning 500 on projects with `zh`.
This another interation on this fix for now.

Sentry https://read-the-docs.sentry.io/issues/5066249292/
Related #370

### DIFF
--- a/readthedocsext/theme/templates/includes/elements/chips/project.html
+++ b/readthedocsext/theme/templates/includes/elements/chips/project.html
@@ -8,7 +8,9 @@
   '{% url "projects-detail" project.slug %}'
 {% endblock chip_view_params %}
 
-{% block chip_url %}{{ project.get_absolute_url }}{% endblock %}
+{% block chip_url %}
+  {{ project.get_absolute_url }}
+{% endblock chip_url %}
 
 {% block chip_classes %}
   basic image
@@ -16,7 +18,7 @@
 
 {% block chip_icon %}
   {% if project.remote_repository %}
-    <img src="{{ project.remote_repository.avatar_url }}" />
+    <img alt="Project avatar" src="{{ project.remote_repository.avatar_url }}" />
   {% else %}
     {% comment %}
       For now, this is using Dicebear, which has a few random generators for
@@ -24,7 +26,8 @@
       similar to generate files on disk and use an md5 hash or similar to link
       to the same image always.
     {% endcomment %}
-    <img src="https://api.dicebear.com/5.x/shapes/svg?size=64&backgroundColor=b8abd4&shape1Color=8c74c1&shape2Color=7854c5&shape3Color=f0f0ff&seed={{ project.pk }}" />
+    <img alt="Project avatar"
+         src="https://api.dicebear.com/5.x/shapes/svg?size=64&backgroundColor=b8abd4&shape1Color=8c74c1&shape2Color=7854c5&shape3Color=f0f0ff&seed={{ project.pk }}" />
   {% endif %}
 {% endblock chip_icon %}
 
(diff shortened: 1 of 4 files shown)


humitos:
I couldn't fix the linting regarding inline CSS styles because we need them.

readthedocsext/theme/templates/includes/elements/chips/project.html
```diff
@@ -2,28 +2,32 @@
...
+{% endblock chip_url %}
 
 {% block chip_classes %}
   basic image
 {% endblock chip_classes %}
 
 {% block chip_icon %}
   {% if project.remote_repository %}
-    <img src="{{ project.remote_repository.avatar_url }}" />
+    <img alt="Project avatar" src="{{ project.remote_repository.avatar_url }}" />
```
agjohnson, line 21: This needs translation and should mention the project name. I did a version of this in 

- #403 

Note, there are a few cases of this below that need the same fix.

readthedocsext/theme/templatetags/ext_theme_tags.py
```diff
@@ -124,3 +129,29 @@ def get_spam_score(project):
...
+# Simple solution to not supported "zh" language code.
+#
+# When `project.language="zh"` the Django `language_name_local` raises an exception and returns 500.
+# This simple solution checks for this unsupported language code and replace it by "zh-cn" (Simplified Chinese).
+# We need to decide what to do at the DB level still, but at least this approach solves the immediate issue.
+@register.filter
+def readthedocs_language_name(lang_code):
+    if lang_code == "zh":
+        return language_name("zh-cn")
+    return language_name(lang_code)
```
agjohnson, line 143: This is what I was thinking of doing originally. It seems like a fine fix for now :+1:

readthedocsext/theme/templates/includes/footer.html
```diff
@@ -136,15 +145,15 @@ <h4 class="ui inverted sub header">Tools</h4>
     {% block footer_sponsors %}
       {% if USE_PROMOS %}
 
-        <div class="ui inverted horizontal divider very padded">
-          Sponsored by
-        </div>
+        <div class="ui inverted horizontal divider very padded">Sponsored by</div>
```
agjohnson, line 148: Guess we need to tune djlint for these, I prefer the original on all of these cases where it's reduced down to a single line. Not a now thing, we should open an issue to track some of the djlint fixes perhaps though.

readthedocsext/theme/templates/projects/partials/header.html
```diff
@@ -119,7 +119,9 @@
               <div class="ui overlapping avatar images">
                 {% for user in project.users.all|slice:":8" %}
                   <a href="{% url "profiles_profile_detail" user.username %}">
-                    <img class="ui image" src="{% gravatar_url user.email 32 %}" />
+                    <img alt="User avatar"
```
agjohnson, line 122: Also needs translation and should mention the user.

readthedocsext/theme/templates/projects/partials/header.html
```diff
@@ -138,7 +140,7 @@
               {% if project.remote_repository %}
                 {% with repo=project.remote_repository %}
                   <a href="{{ repo.html_url }}" class="ui basic image nowrap label">
-                    <img src="{{ repo.avatar_url }}" />
+                    <img alt="Project avatar" src="{{ repo.avatar_url }}" />
```
agjohnson, line 143: Also needs translation and mention of project name

agjohnson (approved):
This looks great :+1:

I figured we might need to come back and do this change. It seems like the best approach until we have a database fix.

humitos:
We need a way to ignore these...

```
readthedocsext/theme/templates/projects/partials/header.html
───────────────────────────────────────────────────────────────────────────────
H021 47:10 Inline styles should be avoided. <div class="eight wi
H021 110:10 Inline styles should be avoided. <div class="right al

Linted 2 files, found 2 errors.
```

humitos:
I'll proceed and merge this since the linting issues are not related to this PR. We can fix them in another one.

agjohnson:
You can add a djlint rule. The documentation covers how to exclude rules globally and skip locally in templates.

PR #418, opened 2024-07-09
Avatar: proposal that doesn't require external image/js package

humitos:
Use an approach borrowed from GitHub: use a background color and remove some
pixels (fill them with white).

This does not require any external image or similar.

It took the color pallete from
https://colorhunt.co/palette/77e4c836c2ce478ccf4535c1 and it's pretty easy to
select a different one if we want to.


humitos:
Screenshot_2024-07-09_18-25-27

humitos:
Ideally, we should:

- have bigger white pixels
- don't collide these pixels

PR #424, opened 2024-07-17
Add notifications block to AllAuth templates

ericholscher:
The allauth templates are overwriting the base templates block:

* https://github.com/readthedocs/ext-theme/blob/main/readthedocsext/theme/templates/allauth/layouts/base.html#L3 is overriding
* https://github.com/readthedocs/ext-theme/blob/main/readthedocsext/theme/templates/base.html#L63

Fixes https://github.com/readthedocs/ext-theme/issues/425

### DIFF
--- a/readthedocsext/theme/templates/allauth/layouts/base.html
+++ b/readthedocsext/theme/templates/allauth/layouts/base.html
@@ -6,7 +6,7 @@
     {# This is overriding the normal base.html content-wrapper, #}
     {# so we need to include notifications here so they're displayed. #}
     {% block notifications %}
-      {% include "includes/utils/messages.html" %}
+      {{ block.super }}
     {% endblock notifications %}
 
     <div class="ui centered grid">
--- a/readthedocsext/theme/templates/allauth/layouts/base.html
+++ b/readthedocsext/theme/templates/allauth/layouts/base.html
@@ -3,6 +3,8 @@
 {% block content-wrapper %}
   <div class="ui very padded container">
 
+    {# This is overriding the normal base.html content-wrapper,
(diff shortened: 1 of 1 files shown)


readthedocsext/theme/templates/allauth/layouts/base.html
```diff
@@ -2,6 +2,13 @@
 
 {% block content-wrapper %}
   <div class="ui very padded container">
```
agjohnson, line 4: Not entirely sure, but the HTML structure is the same up until this block. We might be able to just move the override to `content` block instead. This requires removing a closing div below too

```suggestion
{% block content %}
```

agjohnson (approved):
What you have here definitely should work, though I suspect `content-wrapper` is maybe just the wrong element to override. If we can override `content` instead, that would allow the base notifications/messages to show still.

readthedocsext/theme/templates/allauth/layouts/base.html
ericholscher: @agjohnson This template is also defining the content inside the wrapper a few lines down, so I think we still need the wrapper here.

readthedocsext/theme/templates/allauth/layouts/base.html
```diff
@@ -2,6 +2,13 @@
...
+
+    {# This is overriding the normal base.html content-wrapper, #}
+    {# so we need to include notifications here so they're displayed. #}
+    {% block notifications %}
+      {% include "includes/utils/messages.html" %}
```
agjohnson, line 9: This at least avoids duplicating logic

```suggestion
      {{ block.super }}
```

PR #431, opened 2024-07-24
Integrations: use input elements for URL and secret

### LINKED ISSUE #294: Integrations: convert long data fields to input elements
On the integration detail view, we show a secret that can be quite long:

image

In other views, this would be in an input field with a copy button. This is a nicer look and the copy button is really helpful. The organization invite UI is one that comes to mind, but there are a few views that use this pattern

humitos:
Screenshot_2024-07-24_17-02-46

Closes #294

### DIFF
--- a/readthedocsext/theme/templates/projects/integration_webhook_detail.html
+++ b/readthedocsext/theme/templates/projects/integration_webhook_detail.html
@@ -54,11 +54,11 @@
             {% with PUBLIC_API_URL|add:webhook_url as integration_url %}
               <div class="ui fluid action input">
                 <input type="text" readonly value="{{ integration_url }}" />
-                <span class="ui compact icon basic button"
-                      data-clipboard-text="{{ integration_url }}"
-                      data-content="{% trans "Copied!" %}">
+                <button class="ui teal right icon button"
+                        data-clipboard-text="{{ integration_url }}"
+                        data-content="{% trans "Copied!" %}">
                   <i class="fa-duotone fa-copy icon"></i>
-                </span>
+                </button>
               </div>
             {% endwith %}
           </div>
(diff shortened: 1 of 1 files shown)


readthedocsext/theme/templates/projects/integration_webhook_detail.html
```diff
@@ -50,16 +48,18 @@
...
-              <span class="ui compact icon basic button" data-clipboard-text="{{ integration_url }}" data-content="{% trans "Copied!" %}">
-                <i class="fa-duotone fa-copy icon"></i>
-              </span>
+              <div class="ui fluid action input">
+                <input type="text" readonly value="{{ integration_url }}" />
+                <span class="ui compact icon basic button"
+                      data-clipboard-text="{{ integration_url }}"
+                      data-content="{% trans "Copied!" %}">
+                  <i class="fa-duotone fa-copy icon"></i>
+                </span>
```
humitos, line 61: I'm not sure about what's the correct patter for this. In the new SAML templates we are using a `button` element: https://github.com/readthedocs/ext-theme/blob/600e2fb9171a83002c3819bc6e59ce02839d8fd2/readthedocsext/theme/templates/organizations/settings/saml.html#L29-L32

and it renders different as well:

Screenshot_2024-07-24_17-06-55

humitos: I went ahead and used a `button` element here with `teal` color since I think it renders better. Let me know.

agjohnson: That UI uses `basic button`, the SAML view should as well:

https://github.com/readthedocs/ext-theme/blob/86e9e6f5db808abf2af45863dcf2f8dc08ceea12/readthedocsext/theme/templates/projects/integration_webhook_detail.html#L60-L62

Using a primary button color makes these buttons seem more significant than they are. When using a primary button, it should be the only action we expect the user to take on a form/listing/etc -- so we should not have multiple primary buttons.

agjohnson: I checked a few other places we use this pattern, and we don't have a strong pattern here. Teal is probably okay, but I prefer basic as multiple primary buttons is confusing UX.

agjohnson (approved):

readthedocsext/theme/templates/projects/integration_webhook_detail.html
agjohnson: I think the correct solution here is to use neither, I described this here: https://github.com/readthedocs/ext-theme/issues/433

humitos: Males sense 👍

humitos:
I'm merging this PR for now. I can probably take #433 in the next sprint, tho.

PR #434, opened 2024-07-25
Add project: change copy for "Finish" button

humitos:
Idea we talked in the Book Club.

I'm opening the PR here to continue improving this idea.

### DIFF
--- a/readthedocsext/theme/templates/projects/import_base.html
+++ b/readthedocsext/theme/templates/projects/import_base.html
@@ -87,7 +87,9 @@
                   {% if wizard.steps.next %}
                     <input class="ui button primary" type="submit" value="{% trans "Next" %}" />
                   {% else %}
-                    <input class="ui button primary" type="submit" value="{% trans "I've added the file already" %}" />
+                    <input class="ui button primary"
+                           type="submit"
+                           value="{% trans "This file exists" %}" />
                   {% endif %}
                 </div>
               {% endblock wizard_actions %}
--- a/readthedocsext/theme/templates/projects/import_base.html
+++ b/readthedocsext/theme/templates/projects/import_base.html
@@ -3,11 +3,15 @@
 {% load i18n %}
 {% load crispy_forms_tags %}
 
-{% block title %}{% trans "Add project" %}{% endblock %}
+{% block title %}
+  {% trans "Add project" %}
+{% endblock title %}
 
 {% block content %}
   {% block project_add_grid %}
-    <div class="ui very padded centered stackable grid" {% block project_add_data_bind %}{% endblock %}>
+    <div class="ui very padded centered stackable grid"
+         {% block project_add_data_bind %}
+         {% endblock project_add_data_bind %}>
 
       {% block project_add_header %}
         <div class="ui five wide computer four wide large screen computer only column">
@@ -62,7 +66,9 @@
 
         <div class="ui sixteen wide tablet eleven wide computer ten wide large screen column">
           {% block project_add_content_form %}
-            <form class="ui form" action="{% url "projects_import_manual" %}" method="post">
+            <form class="ui form"
+                  action="{% url "projects_import_manual" %}"
+                  method="post">
 
               {% block wizard_form %}
                 {% if wizard.form.forms %}
@@ -82,9 +88,9 @@
               {% block wizard_actions %}
                 <div>
                   {% if wizard.steps.prev %}
-                    <button class="ui button" name="wizard_goto_step" value="{{ wizard.steps.prev }}">
-                      {% trans "Previous" %}
-                    </button>
+                    <button class="ui button"
+                            name="wizard_goto_step"
+                            value="{{ wizard.steps.prev }}">{% trans "Previous" %}</button>
                   {% endif %}
 
                   {% if wizard.steps.next %}
@@ -101,4 +107,4 @@
       {% endblock project_add_content %}
     </div>
   {% endblock project_add_grid %}
-{% endblock %}
+{% endblock content %}
--- a/readthedocsext/theme/templates/projects/import_base.html
+++ b/readthedocsext/theme/templates/projects/import_base.html
@@ -50,6 +50,16 @@
       {% endblock project_add_sidebar %}
 
       {% block project_add_content %}
+
+        {% comment %}
+        Mention what are the steps the user needs to follow before going to the next step:
+        1. Copy the YAML from the example
+        2. Create a `.readthedocs.yaml` file in their project
+        3. Open a PR
+        4. Come back to this wizard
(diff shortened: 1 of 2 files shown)


humitos:
This is how it looks now.

Screenshot_2024-08-05_17-31-12

readthedocsext/theme/templates/projects/import_base.html
```diff
@@ -72,15 +79,15 @@ <h2 class="ui small header">{% trans "Help topics" %}</h2>
...
+                    <button class="ui button"
+                            name="wizard_goto_step"
+                            value="{{ wizard.steps.prev }}">{% trans "Previous" %}</button>
                   {% endif %}
 
                   {% if wizard.steps.next %}
                     <input class="ui button primary" type="submit" value="{% trans "Next" %}" />
                   {% else %}
-                    <input class="ui button primary" type="submit" value="{% trans "Finish" %}" />
+                    <input class="ui button primary" type="submit" value="{% trans "I've added the file already" %}" />
```
ericholscher, line 90:
```suggestion
                    <input class="ui button primary" type="submit" value="{% trans "I've setup my YAML file" %}" />a
```

readthedocsext/theme/templates/projects/import_config.html
```diff
@@ -5,12 +5,16 @@
...
+{% block project_add_content_classes %}
+  ui fourteen wide tablet twelve wide computer column
+{% endblock project_add_content_classes %}
 
 {% block project_add_content_form %}
   <div class="ui small info message">
     {% blocktrans trimmed %}
-      Save and commit the following example configuration to your repository root as <code>.readthedocs.yaml</code>.
+      Read the Docs <strong>requires a <code>.readthedocs.yaml</code> file at the root of the repository</strong> to build your project's documentation.
+      Open a pull request in your repository with this file using the following example for your documentation tool, commit, push and merge the changes.
```
ericholscher, line 16:
```suggestion
      You should commit this change to your repository.
      You can pick an example for your documentation tool as a starting point below.
```

readthedocsext/theme/templates/projects/import_config.html
```diff
@@ -5,12 +5,16 @@
...
+      Then, come back to this page and finish the process.
```
ericholscher, line 17:
```suggestion
      Then come back to this page and finish building your project.
```

ericholscher (approved):
Just a couple copy suggestions.

readthedocsext/theme/templates/projects/import_config.html
agjohnson: I'd second something like the first line in this suggestion. It's good that we explain the configuration file is required, but the original text actually feels clearer to me overall. More words means it's less likely _any of this_ will be scanned by the user.

Specifically, describing the pull request workflow feels a bit too thorough.

Generally though, we should avoid writing copy to explain UI. If the UI is good, the text is unnecessary. If a UI needs text to explain it, there is something in the UI that should be changed.

But this selection doesn't feel like a problem yet. We first need users to consistently use add the configuration file on project creation. I would make this selection mandatory next, as a form field and even blocking the form until the user selects an example or "I don't need help".

readthedocsext/theme/templates/projects/import_config.html
agjohnson, line 17: Summarizing everything above, I feel it works best clear and short. Mentioning the file is required is a great upgrade, as is setting the expectation we're waiting on the user for something.

```suggestion
      A <strong><code>.readthedocs.yaml</code> configuration file is required at the root of your repository</strong>.
      Save and commit this file before continuing.
```

readthedocsext/theme/templates/projects/import_base.html
humitos: Originally, I've used "I've already added the YAML file" -- but I think that "YAML file" may not be clear for everybody, so I changed for something more explicit like "configuration file", but then the button was too big, and finally use the generic "the file" version 😄 

> we might make a button that says "This file doesn't exist" that just pops up a modal with the same directions we just gave

I think I like this approach. In the first place I wanted to give users _very clear_ instructions numbered step by steps -- not just a paragraph, but I didn't want to change the whole layout for this. A modal would give us some extra space to list each of the steps required here and fit my original goal.

What do you think?

readthedocsext/theme/templates/projects/import_config.html
humitos: > Generally though, we should avoid writing copy to explain UI. If the UI is good, the text is unnecessary. If a UI needs text to explain it, there is something in the UI that should be changed.

I agree with this rule. However, I prefer to add some context in copy for those UIs that we are aware they lack them and are not easy to follow, instead of just leaving them as they are and continue confusing our users. To me, adding text is usually the first iteration of this process. The following one, is improving the UI with the collected feedback and new ideas.

> I would make this selection mandatory next, as a form field and even blocking the form until the user selects an example or "I don't need help".

So, should we revert the "I've added the YAML file" button to "Finish" and add a checkbox saying "I've added the configuration file" that's mandatory?

agjohnson: >  I prefer to add some context in copy for those UIs that we are aware they lack them and are not easy to follow, instead of just leaving them as they are and continue confusing our users

Yeah, and this is okay as a temporary, hacky solution if/when we hit it. But we do also tend to preemptively over-explain UIs in a number of views.

I don't think this UI has this issue though, we haven't seen many indicators users aren't finding these examples. It could be contributing, for sure, but we also know the main problem is more that we're not communicating expectations to users on project creation.

> So, should we revert the "I've added the YAML file" button to "Finish" and add a checkbox saying "I've added the configuration file" that's mandatory?

We could do this, and it is most explicit, but I think I actually prefer the softer approach you have here first.

readthedocsext/theme/templates/projects/import_base.html
agjohnson: Hah yeah, I figured you might have gone through the same process :upside_down_face: 

> In the first place I wanted to give users very clear instructions numbered step by steps -- not just a paragraph

I didn't mention it, but I did really like this idea. It does feel like maybe a lot of information to fit in the main view though

> What do you think?

I really like this idea! It gives us room for the content we want and it's a soft way of catching the user. Hopefully we're giving them the details they immediately need to continue, without sending them on a side quest to figure out what we're talking about.

Also, I see how this pattern might naturally evolve into performing a check for our config automatically -- where maybe the buttons become "Check my repository" and this same modal is shown if this check fails.

humitos:
OK. I've updated this PR with the feedback received. I think we can move forward with this for now and come later to add the modal with more step by step instructions.

ericholscher (approved):

PR #437, opened 2024-07-29
Social accounts: disable disconnect button if the account can't be disconnected

stsewd:
Closes https://github.com/readthedocs/readthedocs-corporate/issues/1019

Needs https://github.com/readthedocs/readthedocs.org/pull/11503

Screenshot

### DIFF
--- a/readthedocsext/theme/templates/socialaccount/partials/social_account_list.html
+++ b/readthedocsext/theme/templates/socialaccount/partials/social_account_list.html
@@ -19,9 +19,11 @@
 {% block create_button %}
   <div class="ui green button"
        data-bind="click: $root.show_modal('socialaccount-connections')">{% trans "Add new connection" %}</div>
-{% endblock %}
+{% endblock create_button %}
 
-{% block list_placeholder_icon_class %}fa-duotone fa-user-unlock{% endblock %}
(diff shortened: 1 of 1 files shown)


readthedocsext/theme/templates/socialaccount/partials/social_account_list.html
```diff
@@ -2,6 +2,7 @@
 
 {% load i18n %}
 {% load gravatar %}
+{% load can_be_disconnected from readthedocs.socialaccounts %}
```
stsewd, line 5: Namespaces are great!

agjohnson: I didn't know about this! I would like to use this as much as possible. Wildcard imports in templates are very confusing and I don't like struggling to guess where template tags/filters are coming from.

agjohnson (approved):
Looks great!

readthedocsext/theme/templates/socialaccount/partials/social_account_list.html
stsewd: Same, it's nice going back to the Django docs and find good things like this :D

PR #439, opened 2024-08-06
Standardize copy buttons

### LINKED ISSUE #433: Standardize clipboard button styles
Some views use a `.ui.basic.button` and some views use `.ui.teal.button` or `.ui.primary.button`. The issue with using a primary button color for secondary actions is that it dilutes other, actually important buttons like form save buttons. The UX around buttons should be that secondary actions are _not_ primary colored.

I think the correct pattern here is to use `.ui.button` instead for all of these inputs. This mimics listing views, where icon buttons are not `.ui.basic` or `.ui.primary.button`:

image

And would look like this:

image

humitos:
I split this PR in 2 commits to make it easier to review.

Closes #433

### DIFF
--- a/readthedocsext/theme/templates/organizations/settings/saml.html
+++ b/readthedocsext/theme/templates/organizations/settings/saml.html
@@ -27,7 +27,7 @@
 
         <div class="ui fluid action input">
           <input type="text" readonly id="rtd-saml-login-url" value="{{ rtd_saml_login_url }}" />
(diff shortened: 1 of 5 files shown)


agjohnson (approved):
:+1:

agjohnson:
Yet another lint check random failure. Is something happening on the CircleCI side maybe? We've had a lot of these lately.

humitos:
I've seen issues on CircleCI trying to install NodeJS a few times already. I'm not surprised. I think it's usually a network issue or similar. Most of the times, re running the job works.

PR #440, opened 2024-08-07
Use target="_blank" for external links

humitos:
This is based on a feedback we received from a user.
As a general rule, we should use `target=_blank` for all the links that go
_outside_ the dashboard.

**There is a lot of lint changes**. I split the work into different commits to make it easier to review.


agjohnson (approved):
Seems good, I just scanned this though

humitos:
There is some inline style lint issues still, but I'm going to merge this as-is for now because I don't know how to skip them. Also, I'm a little tired of fixing _such amount_ of lint issues in this PR. It touched a lot of files. We can continue fixing them in future iterations.

PR #441, opened 2024-08-07
Tune djlint configuration

agjohnson:
Use max_line_length=80, this only controls when attributes/blocks are
wrapped. This setting is why nicer block formatting is sometimes
condensed down to single lines.

The lint rule is for catching wildcard load imports.

### DIFF
--- a/.djlint_rules.yaml
+++ b/.djlint_rules.yaml
@@ -1,0 +1,6 @@
+- rule:
+    name: T901
+    message: Wildcard load, use load X from Y
+    flags: re.DOTALL
+    patterns:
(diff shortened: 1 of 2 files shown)


.djlint_rules.yaml
```diff
@@ -0,0 +1,6 @@
+- rule:
+    name: T901
+    message: Wildcard load, use load X from Y
+    flags: re.DOTALL
+    patterns:
+      - '{% load \w+(?! from \w+) %}'
```
stsewd, line 6: Nice!

stsewd (approved):

PR #442, opened 2024-08-08
Add support for proxito error pages

### LINKED ISSUE #304: Pages: add documentation/user error pages

### LINKED ISSUE #411: Error pages use inline JS
https://github.com/readthedocs/ext-theme/blob/846d19484eb2972f80365cb48d3ac53b935815b0/readthedocsext/theme/templates/errors/dashboard/base.html#L44-L46

https://github.com/readthedocs/ext-theme/blob/846d19484eb2972f80365cb48d3ac53b935815b0/readthedocsext/theme/templates/support/error.html#L22-L26

agjohnson:
These pages are minimal error pages compared to the dashboard pages, and don't
have a header, footer, our JS, or our icons.

Also tunes up dashboard error pages to match, right before the pages were
copied.

- Fixes #411
- Fixes #304
- Requires https://github.com/readthedocs/readthedocs.org/pull/11494
- Requires https://github.com/readthedocs/readthedocs-corporate/pull/1837
- Requires https://github.com/readthedocs/common/pull/220


agjohnson:
I finally had a chance today to get back to this work and wrap it up. I have tested this a good amount, including the underlying, required PRs. Here is where this all is right now.

The errors are centered in the window and are pretty minimal in comparison to what the error pages are now. The lack of major branding, header, and footer are all intentional.

The global search input was dropped for now too, as was most of the content from the contextualized 404 pages that seemed to target maintainers instead of project readers. The DNS 404 error still targets project maintainers however.

On desktop:

image

On mobile:

image

#### Errors
##### 400
image
##### 401
image
##### 403
image
##### 404 (unrecognized custom domain)
image
##### 404 (page not found)
image
##### 404 (translation not found)
image
##### 404 (version not found)
image
##### 404 (project not found)
image
##### 410
image
##### 418
image
##### 429
image
##### 500
image
##### 504
image

readthedocsext/theme/templates/errors/dashboard/base.html
```diff
@@ -16,14 +16,16 @@
...
+                 style="{% block error_icon_computer_style %}--fa-secondary-opacity: .5;
+                        {% endblock error_icon_computer_style %}"></i>
             {% endblock error_icon_computer %}
           </div>
           <div class="ui mobile only center aligned sixteen wide column">
             {% block error_icon_mobile %}
-              <i class="fad {% block error_icon_mobile_name %}{% endblock %} big black inverted shadowed circular icon"
-                 style="{% block error_icon_mobile_style %}--fa-secondary-opacity: .5;{% endblock %}"></i>
+              <i class="fad {% block error_icon_mobile_name %}{% endblock error_icon_mobile_name %} big black inverted shadowed circular icon"
+                 style="{% block error_icon_mobile_style %}--fa-secondary-opacity: .5;
```
humitos, line 27: Are we OK with inline CSS here?

readthedocsext/theme/templates/proxito/base.html
```diff
@@ -1,29 +1,50 @@
...
-        <div class="four wide computer five wide tablet eleven wide mobile left aligned column">
-          <div class="ui horizontally fitted basic segment">
-            <a href="https://{{ PRODUCTION_DOMAIN }}/"
-               aria-label="Read the Docs homepage">
-              <img class="ui image"
-                   src="{% static 'readthedocsext/theme/images/logo-wordmark-dark.svg' %}"
-                   width="220"
-                   alt="Read the Docs logo" />
-            </a>
+{% endblock header %}
```
humitos, line 16: It would be good to add a comment here explaining _why_ we are overriding these blocks to remove all the content from them. I see ourselves being confused about this in the future.

humitos (approved):
Looks great 💯

readthedocsext/theme/templates/errors/dashboard/base.html
agjohnson: We talked about these in the CSP PRs too I believe, for now they are fine.

Eventually, we could hardcode the FA specific rules as special CSS classes instead. This is the only place we're using inline styles at the moment, similar to the website. We permit these with htmlvalidate there:

https://github.com/readthedocs/website/blob/53ca2816a915c836c41404dbc0b276a9f4db813a/.htmlvalidate.json#L10

stsewd:
Goodbye maze :(

PR #448, opened 2024-08-13
Add project: show/hide `team` field based on selected organizatio

humitos:
Required by: https://github.com/readthedocs/readthedocs-corporate/pull/1851


src/js/project/create.js
```diff
@@ -180,3 +180,48 @@ export class ProjectCreateView extends ResponsiveView {
...
+    );
+    organizationSelector.addEventListener("change", function () {
+      const value = organizationSelector.value;
+      const teamSelector = document.querySelector(
+        `#div_id_basics-team__${value}`,
+      );
+
+      // TODO: we can't call an upper function using `this` from a lambda function :/
+      //
+      // this.removeAllTeamSelectors();
```
humitos, line 203: What's the best way to do this here?

agjohnson: This is what the anonymous function syntax, `() => {}`, solves. The old way of getting around this is explicit mapping of the scoped `this`, which is tedious and a hacky pattern:

```js
const this_old = this;
foo.addEventListener("change", function () {
  this_old.removeAllTeamSelectors()
});
```

`() => {}` doesn't rescope `this`.

src/js/project/create.js
```diff
@@ -180,3 +180,48 @@ export class ProjectCreateView extends ResponsiveView {
...
+    this.removeAllTeamSelectors();
+
+    const organizationSelector = document.querySelector(
+      "#id_basics-organization",
+    );
+    organizationSelector.addEventListener("change", function () {
+      const value = organizationSelector.value;
+      const teamSelector = document.querySelector(
+        `#div_id_basics-team__${value}`,
+      );
```
agjohnson, line 199: So, you are mixing two patterns in this class: Knockout view models and DOM based manipulation through native APIs.

When writing Knockout views, you should not replicate Knockout data binding functions with JS native APIs (`querySelector`, event listeners, and class manipulation). Knockout has data bindings for everything you're doing here and it's confusing to have conflicting or inconsistent patterns.

It's okay to use these patterns when writing a web component, as we are skipping all of Knockout.

PR #450, opened 2024-08-14
Add project: show/hide team field based on selected organization

humitos:
It allows users to search by team's name or organization's name.

Peek 2024-08-14 14-47

Required by: readthedocs/readthedocs-corporate#1851


readthedocsext/theme/templates/projects/import_basics.html
```diff
@@ -5,3 +5,21 @@
...
+{% block extra_head %}
+  {{ block.super }}
+
+  {% comment %}
+This is just a way to pass some data from the backend to the frontend.
+We currently don't have an API endpoint to list all the teams for all the organizations.
+This list is generated in the backend and dumped into the HTML,
+then the JavaScript code loads it and use it for SearchUI
+  {% endcomment %}
+  <script id="organizations-teams" type="application/json">{{ searchUI|escapejs }}</script>
```
agjohnson, line 22: This should be placed outside of `head` and should use the `jsonInit` binding, placed nested in an element initialized as a KO model view.

Search for `jsonInit` in other views for examples of how to do this.

`jsonInit` is also documented in the ext-theme docs: https://docs.ops.verbthenouns.com/projects/ext-theme/en/latest/api/js/plugins.html#application.plugins.jsonInit

src/js/project/create.js
```diff
@@ -180,3 +180,50 @@ export class ProjectCreateView extends ResponsiveView {
...
+        .text.replace(/\\u0027/g, '"'),
+    );
+
+    jquery(".ui.search").search({
+      type: "category",
+      minCharacters: 1,
+      source: content,
+      searchFields: ["title", "category"],
+      fullTextSearch: false,
+    });
```
agjohnson, line 226: This is a Knockout view class so should use native KO bindings instead of manual parsing and initializing of FUI elements. What you are doing with JSON parsing is already supported with the `jsonInit` and `semanticui` binding used in other views.

readthedocsext/theme/templates/projects/import_base.html
```diff
@@ -71,6 +71,21 @@ <h2 class="ui small header">{% trans "Help topics" %}</h2>
...
+                {% endcomment %}
+                <div class="ui horizontal category search">
+                  <div class="ui icon input">
+                    <input class="prompt"
+                           type="text"
+                           placeholder="Search organization and team...">
+                    <i class="search icon"></i>
+                  </div>
+                  <div class="results"></div>
+                </div>
```
agjohnson, line 87: This is close, but looks like it's missing some wrapping elements for form fields. This is the reason why there is no spacing between this field and the button below.

You should add the necessary field elements (`.ui.field` and `label`/etc) manually:

https://fomantic-ui.com/collections/form.html

readthedocsext/theme/templates/projects/import_base.html
```diff
@@ -71,6 +71,21 @@ <h2 class="ui small header">{% trans "Help topics" %}</h2>
...
+
+                {% comment %}
+                Temporal approach to render a new Search (category) field to select the Organization and Team
+                where the project is going to be added
+                {% endcomment %}
+                <div class="ui horizontal category search">
+                  <div class="ui icon input">
+                    <input class="prompt"
+                           type="text"
+                           placeholder="Search organization and team...">
```
agjohnson, line 83: Reminder to translate strings in templates:

```suggestion
                           placeholder="{% trans "Search organizations and teams..." %}">
```

agjohnson (commented):
The category search looks nicer :+1:

I noted already on the other PR or issue that this is maybe a good field to isolate, but I commented here specifically on how to maintain JS and template patterns used in the rest of the views. Be careful not to perform manual/native JS DOM manipulation, this is what KO provides us already and it's confusing to mix these patterns. Some of the standards and patterns are discussed more here:

https://docs.ops.verbthenouns.com/projects/ext-theme/en/latest/templates.html

This does not yet reflect replacing model views with web components as this still feels like a pattern we're testing. For now, we should still stick to using KO and keeping everything consistent.

readthedocsext/theme/templates/projects/import_base.html
```diff
@@ -71,6 +71,21 @@ <h2 class="ui small header">{% trans "Help topics" %}</h2>
                 {% else %}
                   {{ wizard.form|crispy }}
                 {% endif %}
+
+                {% comment %}
+                Temporal approach to render a new Search (category) field to select the Organization and Team
+                where the project is going to be added
+                {% endcomment %}
+                <div class="ui horizontal category search">
```
agjohnson, line 79: Also I think this should be `.ui.fluid.search` to match the other fields. `fluid` expands the field to match the width of the parent container.

PR #452, opened 2024-08-14
Link to old dashboard for SAML login

stsewd:
The new dashboard doesn't work for granting access to docs pages yet.


ericholscher (approved):
Would be good to do this for BitBucket as well if it isn't hard, but I imagine it's more tricky given that Allauth is generating them.

agjohnson (approved):

Issue #393, opened 2024-06-19
Missing template for `acl/shared_password_required.html`


stsewd:
This template is used by proxito when a documentation site is password protected.

ericholscher:
Is there something difficult about this, or should we just knock it out?

stsewd:
This depends on having a base template for proxito https://github.com/readthedocs/ext-theme/issues/304. We can also just do a very basic template to have as a placeholder.

PR #456, opened 2024-08-19
Proxito: add template for password protected docs

stsewd:
This is a simple placeholder until we have
the base templates for proxito ready.

We need this template, so tests from https://github.com/readthedocs/readthedocs-corporate/pull/1807 pass.

Ref https://github.com/readthedocs/ext-theme/issues/393

Screenshot 2024-08-19 at 16-49-03 Authentication required - Read the Docs

### DIFF
--- a/readthedocsext/theme/templates/acl/shared_password_required.html
+++ b/readthedocsext/theme/templates/acl/shared_password_required.html
@@ -23,6 +23,8 @@
   </form>
 
   <p>
-    {% blocktrans %}Already have an account? Then please <a href="//{{ PRODUCTION_DOMAIN }}/accounts/login/">log in</a>.{% endblocktrans %}
+    {% blocktrans trimmed with domain=PRODUCTION_DOMAIN %}
+      Already have an account? Then please <a href="https://{{ domain }}/accounts/login/">log in</a>.
+    {% endblocktrans %}
   </p>
 {% endblock content %}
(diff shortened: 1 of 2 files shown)


readthedocsext/theme/templates/acl/shared_password_required.html
```diff
@@ -0,0 +1,28 @@
...
+  </p>
+
+  <form class="ui form" method="post">
+    {% csrf_token %}
+    {{ form|crispy }}
+    <button class="ui primary button" type="submit">{% trans "Log in" %}</button>
+  </form>
+
+  <p>
+    {% blocktrans %}Already have an account? Then please <a href="//{{ PRODUCTION_DOMAIN }}/accounts/login/">log in</a>.{% endblocktrans %}
```
agjohnson, line 26: When using blocktrans, make sure to use a context variable instead of globals. It's a small change, but helps isolate translation changes from code changes:

```suggestion
    {% blocktrans with domain=PRODUCTION_DOMAIN %}Already have an account? Then please <a href="//{{ domain }}/accounts/login/">log in</a>.{% endblocktrans %}  
```

And maybe it's also worth consistently generating the path with `url`? djlint normally yells about this if the URL is relative

agjohnson (approved):
This looks good :+1: I'll come back for a second pass on this next, so I just left some general notes.

readthedocsext/theme/templates/acl/shared_password_required.html
stsewd: > And maybe it's also worth consistently generating the path with url?

We don't have access to the URLs from the main dashboard from proxito.

PR #457, opened 2024-08-19
Fix endblock

stsewd:
Found this one while running tests at https://github.com/readthedocs/readthedocs-corporate/pull/1807/ :D


agjohnson (approved):
Good catch :+1:

PR #465, opened 2024-08-27
djlint: update rules

humitos:
Make `custom_html` lint rule more generic.

### DIFF
--- a/.djlintrc
+++ b/.djlintrc
@@ -3,6 +3,6 @@
   "preserve_blank_lines": true,
   "ignore": "H006",
   "max_line_length": 80,
-  "custom_html": "readthedocs\\(-\\w+\\)+",
+  "custom_html": "readthedocs-\\w+-\\w+-\\w+,readthedocs-\\w+-\\w+,readthedocs-\\w+",
   "format_attribute_template_tags": false
 }
--- a/.djlintrc
+++ b/.djlintrc
@@ -3,6 +3,6 @@
   "preserve_blank_lines": true,
   "ignore": "H006",
   "max_line_length": 80,
-  "custom_html": "readthedocs-\\w+,readthedocs-\\w+-\\w+",
+  "custom_html": "readthedocs\\(-\\w+\\)+",
   "format_attribute_template_tags": false
 }
--- a/readthedocsext/theme/templates/builds/partials/build_list.html
+++ b/readthedocsext/theme/templates/builds/partials/build_list.html
@@ -41,8 +41,12 @@
         Rebuild version
       {% endblocktrans %}
 
-      <readthedocs-menu-build-rebuild class="ui button" url="{% url "projects-versions-builds-list" object.project.slug object.version.slug %}" csrf-token="{{ csrf_token }}" data-content="{{ text_retry }}" aria-label="{{ text_retry }}">
-      <i class="fas fa-arrows-rotate icon"></i>
+      <readthedocs-menu-build-rebuild class="ui button"
+                                      url="{% url "projects-versions-builds-list" object.project.slug object.version.slug %}"
+                                      csrf-token="{{ csrf_token }}"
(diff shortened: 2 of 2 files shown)


readthedocsext/theme/templates/builds/partials/build_list.html
```diff
@@ -41,16 +41,9 @@
...
 
-      {# djlint:off #}
-      <readthedocs-menu-build-rebuild
-                       class="ui button"
-                       url="{% url "projects-versions-builds-list" object.project.slug object.version.slug %}"
-                       csrf-token="{{ csrf_token }}"
-                       data-content="{{ text_retry }}"
-                       aria-label="{{ text_retry }}">
-        <i class="fas fa-arrows-rotate icon"></i>
+      <readthedocs-menu-build-rebuild class="ui button" url="{% url "projects-versions-builds-list" object.project.slug object.version.slug %}" csrf-token="{{ csrf_token }}" data-content="{{ text_retry }}" aria-label="{{ text_retry }}">
```
agjohnson, line 44: This still doesn't look right, I think the match is not right yet? This should reformat with a break after the first attribute. I think djlint is skipping the element.

.djlintrc
```diff
@@ -3,6 +3,6 @@
   "preserve_blank_lines": true,
   "ignore": "H006",
   "max_line_length": 80,
-  "custom_html": "readthedocs-\\w+,readthedocs-\\w+-\\w+",
+  "custom_html": "readthedocs\\(-\\w+\\)+",
```
agjohnson, line 6: This seems correct, I wonder why it doesn't look correct though.

I wonder if we're hitting a bug in djlint. These rules were working correctly just recently, and handled `readthedocs-foo-bar` just fine. I see they are not handling that either, so perhaps this isn't a matter of the extra hyphen.

agjohnson: Ah, looks familiar:

- https://github.com/djlint/djLint/issues/883

agjohnson: Hrm, maybe not. I think this is just the regex actually. Your regex isn't matching at all for some reason, and I found I had the order backwards apparently:

```diff
diff --git a/.djlintrc b/.djlintrc
index 2d9347e..e10b574 100644
--- a/.djlintrc
+++ b/.djlintrc
@@ -3,6 +3,6 @@
   "preserve_blank_lines": true,
   "ignore": "H006",
   "max_line_length": 80,
-  "custom_html": "readthedocs-\\w+,readthedocs-\\w+-\\w+",
+  "custom_html": "readthedocs-\\w+-\\w+-\\w+,readthedocs-\\w+-\\w+,readthedocs-\\w+",
   "format_attribute_template_tags": false
 }
```

This config produces what I'd expect:

```diff
readthedocsext/theme/templates/errors/proxito/4xx.html

@@ -27,7 +27,8 @@

 {% endblock error_content %}

 {% block error_content_help %}
-  <readthedocs-one-two-three foo="foo" bar="bar bar bar bar bar bar bar bar bar bar bar bar bar bar bar bar bar bar bar bar">
+  <readthedocs-one-two-three foo="foo"
+                             bar="bar bar bar bar bar bar bar bar bar bar bar bar bar bar bar bar bar bar bar bar">
     {# On purpose there is no link to our support form here #}
     {% blocktrans trimmed %}
       Check that your request is valid or try again later.

1 file was updated.
```

humitos: I used this rule in the end. It seems my trick with parenthesis doesn't work for some reason 🤷🏼

agjohnson: I think at very least the escaping is off, and "readthedocs(-\\w+)+" should match, but it still doesn't. Almost certainly this is some djlint internal stuff, it doesn't look like `custom_html` is a solid configuration element. :shrug:

I see others using the redundant version of this, so it's probably best.

agjohnson (approved):
That looks how I'd expect :+1:

Issue #474, opened 2024-09-06
Build: detail page updates


agjohnson:
- [x] Drop old output for build.output and no command list. Show something at least for now?
- [x] Replace old pattern of hardcoded `style="display: hidden"` for `ko hidden` class
- [x] There are a few related issues around linking, wrapping, etc as well

PR #475, opened 2024-09-07
Fix more issues with URL resolution and null version builds

### LINKED ISSUE #412: `NoReverseMatch` when hitting builds listing page
Go to https://app.readthedocs.org/projects/pycampes/builds/ and/or https://app.readthedocs.org/projects/pygsp2/builds/ and you will get a 500.

Sentry error: https://read-the-docs.sentry.io/issues/5564576585/

Front logo [Front conversations](https://app.frontapp.com/open/top_11t9zb)

### LINKED ISSUE #473: Build: Detail page NoReverseMatch
https://read-the-docs.sentry.io/issues/5690482287/

```
NoReverseMatch: Reverse for 'projects-versions-builds-list' with arguments '('[REDACTED]', '')' not found. 2 pattern(s) tried: ['api/v3/projects/(?P<parent_lookup_project__slug>[^/.]+)/versions/(?P<parent_lookup_version__slug>[^/]+)/builds\\.(?P<format>[a-z0-9]+)/?$', 'api/v3/projects/(?P<parent_lookup_project__slug>[^/.]+)/versions/(?P<parent_lookup_version__slug>[^/]+)/builds/$']
(30 additional frame(s) were not displayed)
...
  File "django/template/base.py", line 1005, in <listcomp>
    return SafeString("".join([node.render_annotated(context) for node in self]))
  File "django/template/base.py", line 966, in render_annotated
    return self.render(context)
  File "django/template/defaulttags.py", line 471, in render
    url = reverse(view_name, args=args, kwargs=kwargs, current_app=current_app)
  File "django/urls/base.py", line 88, in reverse
    return resolver._reverse_with_prefix(view, prefix, *args, **kwargs)
  File "django/urls/resolvers.py", line 828, in _reverse_with_prefix
    raise NoReverseMatch(msg)
```

agjohnson:
- Fixes #473
- Fixes #412

### DIFF
--- a/readthedocsext/theme/templates/builds/build_detail.html
+++ b/readthedocsext/theme/templates/builds/build_detail.html
@@ -145,11 +145,13 @@
             {% trans "Cancel" %}
           </a>
 
+          {% if object.project and object.version %}
           <readthedocs-menu-build-rebuild class="link item ko hidden" data-bind="css: { hidden: !can_retry() }" url="{% url "projects-versions-builds-list" object.project.slug object.version.slug %}" csrf-token="{{ csrf_token }}">
           <i class="fa-duotone fa-refresh icon"></i>
           {# Translators: this refers to rebuilding a version build #}
           {% trans "Rebuild" %}
           </readthedocs-menu-build-rebuild>
+          {% endif %}
         {% endif %}
 
         {% trans "View documentation for this build" as text_view_docs %}
--- a/readthedocsext/theme/templates/builds/build_detail.html
+++ b/readthedocsext/theme/templates/builds/build_detail.html
@@ -1,8 +1,7 @@
 {% extends "projects/base.html" %}
 
-{% load i18n %}
-{% load privacy_tags %}
-{% load static %}
+{% load trans from i18n %}
+{% load is_admin from privacy_tags %}
 
 {% block title %}
   {{ build.project.name }}
@@ -81,7 +80,8 @@
                 <div class="item">
                   <i class="grey fa-duotone fa-clock icon"></i>
                   <div class="content" data-bind="if: is_finished">
-                    <div data-bind="text: '{% trans "Build took %d" %}'.replace('%d', length_display())"></div>
+                    <div data-bind="text: '{% trans "Build took %d" %}'.replace('%d', length_display())">
+                    </div>
                   </div>
                 </div>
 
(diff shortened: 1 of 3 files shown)


readthedocsext/theme/templates/builds/build_detail.html
```diff
@@ -166,9 +171,12 @@
 
     {# TODO: remove this `build.output block completely #}
     {% if build.output %}
+      {# djlint: off H021 #}
```
agjohnson, line 174: Noted to come back to this and other lint skips in #474

readthedocsext/theme/templates/builds/build_detail.html
```diff
@@ -81,7 +80,8 @@
                 <div class="item">
                   <i class="grey fa-duotone fa-clock icon"></i>
                   <div class="content" data-bind="if: is_finished">
-                    <div data-bind="text: '{% trans "Build took %d" %}'.replace('%d', length_display())"></div>
+                    <div data-bind="text: '{% trans "Build took %d" %}'.replace('%d', length_display())">
```
humitos, line 83: Magic! 🪄 . This is the first time I see the usage of `text:` with `{% trans %}` and `.replace()` 🤯

readthedocsext/theme/templates/projects/partials/project_list.html
```diff
@@ -94,7 +94,10 @@
 {% block list_item_extra_items %}
   {% if object.has_good_build %}
     {% with build=object.get_latest_build %}
-      {% include "includes/elements/chips/build.html" with build=build %}
+      {# Safety check, there seems to be some case where `has_good_build` and `get_latest_build` still return a null build #}
```
humitos, line 97: After a quick search, I don't find how `has_good_build` could be returning `True` when it shouldn't:

https://github.com/readthedocs/readthedocs.org/blob/9ef753323822eafdd9c72cb57e6f2ef6b60b305e/readthedocs/projects/models.py#L922-L927

https://github.com/readthedocs/readthedocs.org/blob/9ef753323822eafdd9c72cb57e6f2ef6b60b305e/readthedocs/core/templatetags/privacy_tags.py#L32-L36

readthedocsext/theme/templates/projects/partials/project_list.html
```diff
@@ -94,7 +94,10 @@
 {% block list_item_extra_items %}
   {% if object.has_good_build %}
```
humitos, line 95: Here, I would add `get_latest_build` to the if condition probably, since we are depending on that object for rendering: https://github.com/readthedocs/ext-theme/issues/412#issuecomment-2337724200

humitos (commented):
I think this looks good, but I left a few comments that I think it answers your original question.

readthedocsext/theme/templates/builds/build_detail.html
agjohnson: Hah, this is probably a pattern to avoid using, it's mostly a hack :upside_down_face:

Because the JS isn't localized, this will lead to a application localized `Build took` and then in english `56 minutes`.

But until we are translating sources in JS (if we ever do) it's better than not translating though.

readthedocsext/theme/templates/projects/partials/project_list.html
agjohnson: Yeah same, I didn't see anything immediate that could cause this. Perhaps there is a timing bug here? I haven't noticed this in testing, it might be hard to replicate this one

readthedocsext/theme/templates/projects/partials/project_list.html
agjohnson: That's what the condition below for `build and build.project` accomplish, no? I am avoiding two calls to `get_latest_build` by testing the local `build`

humitos (approved):

Issue #462, opened 2024-08-23
Allauth: template regressions


agjohnson:
I just noticed

- #435 

I know this is a legitimately missing template, but I've noticed similar visual regressions in many of the allauth templates. Did the allauth upgrade move/break something?

This is the sort of thing I started noticing fairly recently:

![image](https://github.com/user-attachments/assets/e62ea794-1729-4d82-9544-2d7bf989f280)

I swear I went through many of these templates already though. I want to say these template layers are new and maybe interfering or were just renamed even.

![image](https://github.com/user-attachments/assets/5e4a3fa2-3608-482f-8973-3cbebf7e5327)

PR #479, opened 2024-09-09
Use proxito base templates for Allauth account/socialaccount templates

### LINKED ISSUE #427: Proxito: add account login templates
Similar to #304 we need to add minimal log in templates for proxito. These will probably share a similar base to proxito error templates, as we want minimal templates and the most potential for customization.

Without this, the old dashboard templates will be used and/or there will be errors trying to use these templates on our proxito instance.

- Requires application changes at https://github.com/readthedocs/readthedocs.org/issues/11483

### LINKED ISSUE #435: Email verifycation template is not styled
It seems we are using a default allauth template. This one in particular: https://github.com/pennersr/django-allauth/blob/301324e3ed2bf7c95987f24625ee8ebd7480ce85/allauth/templates/account/verification_sent.html

I think we want to override them to:

- Add an email icon (similar to our "Verify email" template shown when the user clicks the link from the email): https://github.com/readthedocs/ext-theme/blob/790e59de6c0e2a82a9df026a7553470c384bc49f/readthedocsext/theme/templates/account/email_confirm.html#L10-L13
- Don't use title case

Screenshot_2024-07-25_17-50-58

agjohnson:
This uses the entrance/manage base pattern from Allauth to provide two base templates:

 - Any template using base_entrance.html will use the Proxito style base template, which drops the header/footer from the application. CSS/JS is still used however
 - An application styled base_manage.html for pages that are more clearly account management templates

Some missing templates were not styled and were adapted from Allauth templates.

- Fixes #435
- Fixes #427
- Refs #462

### DIFF
--- a/readthedocsext/theme/templates/account/email_confirm.html
+++ b/readthedocsext/theme/templates/account/email_confirm.html
@@ -1,17 +1,17 @@
 {% extends "account/base_entrance.html" %}
 
-{% load i18n %}
-{% load account %}
+{% load blocktrans trans from i18n %}
+{% load user_display from account %}
 
-{% block head_title %}{% trans "Verify email" %}{% endblock %}
+{% block head_title %}
+  {% trans "Verify email" %}
+{% endblock head_title %}
+{% block content_title_text %}
+  <i class="fad fa-envelope icon"></i>
+  {% trans "Verify email" %}
+{% endblock content_title_text %}
 
-{% block content %}
-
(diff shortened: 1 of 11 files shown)


humitos (approved):
I think this looks good. This work with inherit templates is hard to review. In particular, when it inherit for multiple different places and there some external templates (allauth).

It would have been useful to have some screenshot of the pages to see the final rendered result as well.

PR #480, opened 2024-09-11
Add templates for Allauth MFA

### LINKED ISSUE #443: 2FA: implement templates
Allauth provides some templates for 2FA, but they don't look great with the new theme, and some are very simple.

Some examples:

Screenshot 2024-08-08 at 11-12-26 Two-Factor Authentication - Read the Docs
Screenshot 2024-08-08 at 11-11-51 Sign In - Read the Docs

The templates that we need to override are (the html files only):

- https://github.com/pennersr/django-allauth/tree/64.0.0/allauth/templates/mfa/recovery_codes
- https://github.com/pennersr/django-allauth/tree/64.0.0/allauth/templates/mfa/totp
- https://github.com/pennersr/django-allauth/tree/64.0.0/allauth/templates/mfa

The templates under webauthn we can ignore, since we aren't using that feature.

ref https://github.com/readthedocs/readthedocs.org/pull/11524

agjohnson:
I placed this as a subpage under the profile "Security" admin dashboard
for now, though this is still not linked anywhere.

The entrance pages use the minimal login base templates, while the
management pages use the user profile edit base as a base template, so
that all of the management functions appear under the user profile
admin.

- Fixes #443

[Screencast from 2024-09-11 14-07-31.webm](https://github.com/user-attachments/assets/efae06c4-c14f-448f-a496-7f39fbe370dd)

### DIFF
--- a/readthedocsext/theme/templates/profiles/base_edit.html
+++ b/readthedocsext/theme/templates/profiles/base_edit.html
@@ -1,13 +1,11 @@
 {% extends "base.html" %}
 
-{% load i18n %}
-{% load core_tags %}
+{% load trans from i18n %}
 
 {% block notifications %}
   {# User notifications only #}
-  <readthedocs-notification-list
-    url="{% url "users-notifications-list" request.user.username %}"
-    csrf-token="{{ csrf_token }}">
+  <readthedocs-notification-list url="{% url "users-notifications-list" request.user.username %}"
+                                 csrf-token="{{ csrf_token }}">
   </readthedocs-notification-list>
 
   {{ block.super }}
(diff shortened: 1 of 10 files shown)


readthedocsext/theme/templates/profiles/base_edit.html
agjohnson, line 1: I didn't end up changing this file, these are just lint fixes.

humitos (approved):
Looks good to me 💯 

The only thing that I noticed it's a little weird is the clickable sub-section header:

Screenshot_2024-09-12_13-58-23

I think we don't have that pattern anywhere else, but I think it's fine for now.

agjohnson:
We do use this pattern on other admin subpage views currently. I'd like to continue using this pattern more, it provides a bit more context to the admin navigation.

Issue #65, opened 2021-11-17
Add resync repository button in the version creation view template


agjohnson:
The resync repository button should be in the version list template so that it's obvious how to get newly added versions to show up. Right now, you have to know to trigger a build, and that is bad UX.

agjohnson:
@humitos you probably have the most context here, this is another relatively simple change, and some missing UX at the moment. I would add this on the version add view, not the listing view.

humitos:
Here, we would need:

1. an UI element that trigger a Celery task on the backend
2. the Celery task is this one https://github.com/readthedocs/readthedocs.org/blob/98ec4abf92d864b6e29cbb9cce9f036fc5f6c996/readthedocs/projects/tasks/builds.py#L244-L264
3. poll the "task's faked response" (as we do for re-sync remote repositories) and refresh the page with the new versions

I think this is pretty low priority tho, since the needing to re-sync the versions it's an edge case that only happens on not well-configured repositories.

agjohnson:
What number/percent of projects are not well-configured?

humitos:
> What number/percent of projects are not well-configured?

That's a hard question to answer without executing a big query in the database; but you can do it if you want from the replica or similar. This query would be:

- projects without _any integration_ configured
- projects that are not spam (this check is expensive since it's done 1 by 1)

Personally, I think this issue is not a priority at the moment, mainly because we don't have users complaining about this and there is a workaround (trigger a build for `latest`) in case they hit this issue. At this point, it's a "nice to have" in my opinion considering the amount of work required.

agjohnson:
Seems pretty easy to get, though I don't quite understand the state that triggers this UX either. To ballpark though:

```python
proj_misconfigured = Project.objects.filter(integrations__isnull=True, organizations__disabled=False).exclude(name="demo").count()
proj_total = Project.objects.filter(organizations__disabled=False).count()
```

```python
proj_misconfigured = Project.objects.filter(integrations__isnull=True, spam_rules__isnull=True).exclude(name="demo").count()
proj_total = Project.objects.filter(spam_rules__isnull=True).count()
```

Commercial: 3% of active organization projects are not set up correctly
Community: 0.4% are not set up correctly

Given I'm almost certainly not getting an accurate view here, I think it's probably worth _something_ here, but would also agree we don't need a special feature like a sync button.

[...]

PR #481, opened 2024-09-11
Add basic attempt at Resync verisons button

ericholscher:
This is a bit hacky but works for me locally.
Could use some UI/UX feedback to make it work better,
I mostly just copied what was being used in other places.

Fixes https://github.com/readthedocs/ext-theme/issues/65

Screenshot 2024-09-11 at 3 00 26 PM


src/js/modules/menus.js
```diff
@@ -107,3 +107,22 @@ customElements.define(
   "readthedocs-menu-build-rebuild",
   MenuBuildRebuildElement,
 );
+
+/** Menu item for resyncing all **/
+class MenuSyncAllVersionsElement extends APIEventWrapper {
+  // TODO translate this here or use a different pattern
+  static errorMessage = "There was an error syncing versions";
```
humitos, line 114: I always struggle with this, but I guess it's better to have final dots in sentences.

```suggestion
  static errorMessage = "There was an error syncing versions.";
```

readthedocsext/theme/templates/builds/partials/version_list.html
```diff
@@ -17,10 +17,26 @@
...
-    <a href="{% url "project_version_create" project.slug %}" class="ui primary button">
-      <i class="fa-solid fa-plus icon"></i>
-      <span class="text">{% trans "Add version" %}</span>
-    </a>
+    <div class="right aligned eight wide computer eight wide tablet sixteen wide left aligned mobile middle aligned column">
+      {% blocktrans trimmed asvar text_resync %}
+        Resync versions
+      {% endblocktrans %}
+      <readthedocs-menu-sync-all-versions
+                                 class="ui primary button"
```
humitos, line 25: I would make it more discrete here, since it's not the primary action of this page.

```suggestion
                                 class="ui button"
```

src/js/modules/menus.js
```diff
@@ -107,3 +107,22 @@ customElements.define(
...
+
+  getSuccessURL() {
+    // Refresh the existing page to hopefully show new versions in 3 seconds, so they have time to sync.
+    // This is a bit of a hack, but it's better than nothing.
```
humitos, line 118: 😄

ericholscher:
I've updated this to put it in the activate a new version page.

Screenshot 2024-09-18 at 2 31 07 PM

agjohnson:
I would reserve that placeholder space for onboarding content. All other views use this space to point the user to more information in our docs/etc, and the button feels quite prominent for what it's function is -- manually resyncing is great to have in our UI but it is still an outlier in UX.

I'd instead reuse the existing pattern from the repository listing I pointed out above

image

https://github.com/readthedocs/ext-theme/blob/8239442a87d657aa0ac82aff9de92a73e556c838/readthedocsext/theme/templates/projects/partials/project_create_automatic.html#L101-L112

src/js/modules/menus.js
```diff
@@ -107,3 +107,18 @@ customElements.define(
...
+  // TODO translate this here or use a different pattern
+  static errorMessage = "There was an error syncing versions.";
+
+  getSuccessURL() {
+    console.log("Successfully synced versions");
+  }
+}
+
+customElements.define(
+  "readthedocs-button-sync-all-versions",
```
agjohnson, line 122: You will need to update our djlint rules to support more hyphens:

https://github.com/readthedocs/ext-theme/blob/8239442a87d657aa0ac82aff9de92a73e556c838/.djlintrc#L6

Or shorten this to 3 hyphens. We tried a regex and djlint had trouble using this to match multiple levels of hyphenation.

ericholscher:
Something like this:

Screenshot 2024-09-24 at 2 19 18 PM

agjohnson:
That's pretty close, but looks like you added the element inside the `.ui.placeholder.segment` still. I'd check out that example from above, it shows the structure of the elements:

https://github.com/readthedocs/ext-theme/blob/8239442a87d657aa0ac82aff9de92a73e556c838/readthedocsext/theme/templates/projects/partials/project_create_automatic.html#L101-L112

In that, the `.ui.bottom.attached.message` is sibling to the outer `.ui.basic.segment > .ui.placeholder.segment`, but the `.ui.basic.segment` is only for hiding this placeholder block anyways.

ericholscher:
This class structure is super finicky... I don't feel like I'm really learning anything about why it should actually work this way, and it's just futzing with stuff without any real direction. Is there any context for this structure?

agjohnson:
The example I've linked to had the structure we want to replicate and copy. It uses two divs, summarized as:

```
<div class="ui top attached segment"></div>
<div class="ui bottom attached segment"></div>
```

This gives two _attached_ segments -- these have no spacing between them and they have a border separating them.

You can read about this here:

https://fomantic-ui.com/elements/segment.html#attached

Your last attempt was, summarized:

```html
<div class="ui top attached segment">
  <div class="ui placeholder segment">
    <div class="ui bottom attached segment">
    </div>
  </div>
</div>
```

This will render the bottom attached segment inside the placeholder segment, which has its own separate padding and border. The two attached segments need to be sibling elements instead.

agjohnson:
What was here was close, but needed [e9a17d8](https://github.com/readthedocs/ext-theme/pull/481/commits/e9a17d8742394ba7a421602bc70568b48c4c6e91)

With that, it renders correctly, and matches the repository listing UI:

image

ericholscher:
> The example I've linked to had the structure we want to replicate and copy. It uses two divs, summarized as:
> 
> ```
> <div class="ui top attached segment"></div>
> <div class="ui bottom attached segment"></div>
> ```
> 
> This gives two _attached_ segments -- these have no spacing between them and they have a border separating them.
> 
> You can read about this here:
> 
> https://fomantic-ui.com/elements/segment.html#attached
> 
> Your last attempt was, summarized:
> 
> ```
> <div class="ui top attached segment">
>   <div class="ui placeholder segment">
>     <div class="ui bottom attached segment">
>     </div>
>   </div>
> </div>
> ```
> 
> This will render the bottom attached segment inside the placeholder segment, which has its own separate padding and border. The two attached segments need to be sibling elements instead.
> 
> Placeholder segments are described here:
> 
> https://fomantic-ui.com/elements/segment.html#placeholder-segment

Nice -- we should drop this in the frontend docs somewhere, it's useful context 💯

humitos:
I think we should add some kind of feedback or message to the user once you press the button. It shows the spinner and it stops immediately. I'd say that we should probably put a time for 2 or 3 seconds, and add a message saying "Synced, your version should appear in the list now" or similar.

Peek 2024-10-09 16-23

PR #482, opened 2024-09-12
Use multiple tabs for addons configuration page

### LINKED ISSUE #293: Addons: split settings page per-addon
In the short term, as we keep adding configuration options for each addon, we will need to split the addons settings' page per-addon somehow. I'm creating this issue here to keep track of this and have this discussion together.

There are some different proposals here:

* Use the same settings page for all the addons and add a header for each of them (idea partially implemented in https://github.com/readthedocs/readthedocs.org/pull/11031)
* Split the settings page into different pages per-addon (**different** URL, forms, etc)
* Split the settings page into different **tags** per-addon (same URL, form, etc)

Reference: https://github.com/readthedocs/readthedocs.org/pull/11069#pullrequestreview-1846423142

agjohnson:
This is effectively the same as using Crispy Layout API, but allows for authoring more/richer HTML content in the tabs as we need. For example, we might want to include nicer placeholder segments or images in the tabs.

One improvement might be a way to define the fields for each tab in the Form instance, so that we don't have to hardcode field names in templates, but instead iterate over each field in `form.fieldsets.flyout`, or similar.

- Fixes #293

[Screencast from 2024-09-11 17-16-16.webm](https://github.com/user-attachments/assets/8998bd11-5930-4e67-adde-eecf6557025e)

### DIFF
--- a/readthedocsext/theme/templates/projects/addons_form.html
+++ b/readthedocsext/theme/templates/projects/addons_form.html
@@ -1,7 +1,7 @@
 {% extends "projects/project_edit_base.html" %}
 
-{% load i18n %}
-{% load crispy_forms_tags %}
+{% load blocktrans trans from i18n %}
+{% load as_crispy_field crispy from crispy_forms_tags %}
 
 {% block title %}
   {{ project.name }} - {% trans "Addons" %}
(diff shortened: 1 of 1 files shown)


readthedocsext/theme/templates/projects/addons_form.html
```diff
@@ -21,12 +21,88 @@
...
+          Addons provide features on your project's hosted documentation,
+          such as our flyout navigation menu, search, analytics, and more.
+        {% endblocktrans %}
+      </p>
+
+      {{ form.enabled|as_crispy_field }}
+      {{ form.project|as_crispy_field }}
+    </div>
+
+    {% if form.enabled.value == True %}
```
humitos, line 43: I would remove this condition from here. I should _always_ show this form so the user can see all the options and features that are available. However, if addons are disabled, I would make the whole form disabled instead without being able to modify it, similar to what we do in other places.

humitos:
> One improvement might be a way to define the fields for each tab in the Form instance, so that we don't have to hardcode field names in templates, but instead iterate over each field in `form.fieldsets.flyout`, or similar.

This would be pretty good. It will reduce the maintenance here a lot. Have you checked if this is possible already?

humitos (approved):
This looks great for the first iteration.

readthedocsext/theme/templates/projects/addons_form.html
agjohnson: This requires setting many elements as disabled -- the surrounding segments, the navigation bar, and individual fields. In the end, the effect was not much better either, I prefer the UX with the form hidden actually.

We can return to this later if we feel it's important.

agjohnson:
> This would be pretty good. It will reduce the maintenance here a lot. Have you checked if this is possible already?

I haven't, this was just an idea. I'd like to have a solution that allows directly referencing a group of fields by key/attribute, and would simply iterate over that to display the fields.

PR #484, opened 2024-09-12
Update password auth view

### LINKED ISSUE #393: Missing template for `acl/shared_password_required.html`

agjohnson:
Uses same base as Allauth views, which is a derivate of the Proxito base
view already. This fixes some copy and elements, and addresses the link
to log in with other methods at least temporarily.

- Fixes #393

image

### DIFF
--- a/readthedocsext/theme/templates/acl/shared_password_required.html
+++ b/readthedocsext/theme/templates/acl/shared_password_required.html
@@ -1,30 +1,39 @@
-{% extends "proxito/base.html" %}
+{# Follow same pattern as Allauth here #}
+{% extends "account/base_entrance.html" %}
+
 {% load trans blocktrans from i18n %}
 {% load crispy from crispy_forms_tags %}
 
 {% block title %}
   {% trans "Authentication required" %}
 {% endblock title %}
+{% block content_title_text %}
+  {% trans "Authentication required" %}
+{% endblock content_title_text %}
 
-{% block content %}
-  <h1>{% trans "Authentication required" %}</h1>
-
+{% block content_body %}
   <p>
-    {% blocktrans with project_name=project.name %}
-    You need special requirements to access "{{ project_name }}" project documentation.
-    Please, provide a valid password to see the docs.
+    {% blocktrans trimmed %}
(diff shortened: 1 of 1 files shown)


readthedocsext/theme/templates/acl/shared_password_required.html
```diff
@@ -1,30 +1,39 @@
...
-    {% blocktrans trimmed with domain=PRODUCTION_DOMAIN %}
-      Already have an account? Then please <a href="https://{{ domain }}/accounts/login/">log in</a>.
-    {% endblocktrans %}
-  </p>
-{% endblock content %}
+    <div class="ui stackable relaxed text menu">
+      {% comment %}
+        Using the `next` query param here is hopefully safe. It restarts the
+        login workflow, where just linking to the login form or production
+        domain both do not. Both of these link to a slightly different form.
```
humitos, line 30: I think this paragraph is somehow broken. I'm not able to understand it.

> where just linking to the login form or production domain both do not. Both of these link to a slightly different form.

That's the part that I don't follow. Is it correct? 😄

humitos (approved):

readthedocsext/theme/templates/acl/shared_password_required.html
agjohnson: Yeah, it's correct but I didn't put all the details there. The various ways here of linking to something like `//{{ PRODUCTION_DOMAIN }}/accounts/login` produce a login form that is different from the login form readers get when opening up the original URL.

The login form readers get when opening a private version has a different title and the password auth option:

image

If you just open the normal login page, it looks like this:

image

If there is a URL other than a user supplied URL to trigger the first login form, that could be used here too, but I'm not aware of any

stsewd: This is probably not safe... Do we have this in production already to test a couple of things?

stsewd: Looks like it isn't https://docs.ops.verbthenouns.com/_/projects/corporate-ops/sharing/password/, will test locally.

PR #487, opened 2024-09-17
Delete unused gold subscription template

### LINKED ISSUE #407: Gold view templates uses inline JS
We should probably just handle everything from Stripe instead of embedding stripe into our site.

https://github.com/readthedocs/ext-theme/blob/846d19484eb2972f80365cb48d3ac53b935815b0/readthedocsext/theme/templates/gold/subscription_form.html#L22-L35

I'll probably add a specific rule for this view to our CSP header.

agjohnson:
This was replaced by subscription_detail.html, where we work with the
subscription through Stripe Checkout instead.

- Fixes #407
- Refs https://github.com/readthedocs/readthedocs.org/pull/11605


agjohnson:
It's not clear if we actually hit an issue with this view in production or not:

https://github.com/readthedocs/readthedocs.org/pull/11448

I tested a full list of script-src minus "unsafe-inline" and was able to use this view. Before this change, the view was broken locally. We don't seem to actually use inline scripts in this view from what I can tell -- noted at #407

humitos (approved):

PR #490, opened 2024-09-19
Unify disabled feature upgrade patterns

### LINKED ISSUE #291: Add feature disabled template from organization security log view
There was a template included from the security log view in the legacy dashboard. I don't remember if this was ported. A couple notes here if not:

- This template should be reusable and not specific to the security log page.
- The partial that is included for the security log already has a `.ui.placeholder.segment` for when the list of security logs is empty. This feels like a natural place to have an icon and message "This feature is not available in your current plan.."
- Alternatively, wrapping the security log UI with a `.ui.disabled.segment` is an option, and then #281 could handle pointing the user to upgrading.

- #281

----

#### Options

image

image

agjohnson:
Use a callout and disable the page content if the feature is disabled. This
shows the full view but in a disabled state.

Changes are in 1314b79

- Fixes #291

### DIFF
--- a/readthedocsext/theme/templates/includes/crud/table_list.html
+++ b/readthedocsext/theme/templates/includes/crud/table_list.html
@@ -65,7 +65,7 @@
 {% endcomment %}
 
 {% load trans blocktrans from i18n %}
-{% load autopaginate from pagination_tags %}
+{% load autopaginate paginate from pagination_tags %}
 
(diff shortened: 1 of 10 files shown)


agjohnson:
image
image
image
image

humitos (approved):
Looks pretty good! 💯

PR #492, opened 2024-09-20
Build detail page fixes

### LINKED ISSUE #46: Make build detail page start time update as polling happens
Build detail start time text should update as polling happens. For example, started 5 min ago, started 6 min ago, etc. Currently, the time doesn't update immediately, though it does show once the build is done. It might show on reload while the build is active, which just means we need to update the KO observable with the time data on API response.

- [ ] Set the build start time via the polling API response while the build is still active.

### LINKED ISSUE #334: Build details page: add option to disable line wrapping

### LINKED ISSUE #357: Build: command display issues
#334 describes adding an option for toggling wrapping, but in addition to the UX here, there are several display issues with the build commands listing:

- [x] Most of the command UI does not work on mobile viewports. The expand buttons don't seem to collapse the elements and wrapping issues are more noticeable here with the narrow display
- [x] Command line and stdout wrapping on all viewport sizes can break on long lines.

### LINKED ISSUE #474: Build: detail page updates

agjohnson:
- Fixes #46, make date and length field update as the build progresses
- Add field labels to details
- Condense commit/type/PR fields
- Fixes #474
- Fixes #357 by making the output scrolling and unwrapped
- Fixes #334 by supporting optional wrapping

#### Changes

New elements in the detail view

image

Updating times and actually showing a duration, and wrapping

[Screencast from 2024-09-20 18-27-29.webm](https://github.com/user-attachments/assets/bf55d050-4f7e-45cf-8331-f736cd8f36c9)

And viewing on mobile actually works now, it's very broken in production

[Screencast from 2024-09-20 18-31-13.webm](https://github.com/user-attachments/assets/5ff2e906-83dd-4648-a12f-63663e7f692c)


readthedocsext/theme/templates/builds/build_detail.html
```diff
@@ -169,37 +177,23 @@
...
+      {% endcomment %}
+      <div class="ui inverted notification error message">
+        <div class="header">
+          <i class="fas fa-exclamation-circle icon"></i>
+          {% trans "Build output unavailable" %}
+        </div>
+        <p>
+          {% blocktrans %}
+            Due to the age of this build, the output of this build is no longer available.
+          {% endblocktrans %}
```
humitos, line 194: 💯 -- I'm happy we are not supporting this anymore 👍🏼

src/js/build/detail.js
```diff
@@ -414,62 +434,69 @@ export class BuildDetailView {
...
+        this.docs_url(data.docs_url);
+        this.commit_url(data.commit_url);
+        this.builder(data.builder);
+        this.config(data.config);
+        this.state(data.state);
+        this.state_display(data.state_display);
+
+        // Always update date and length, as these should update as the build progresses
+        this.date.valueHasMutated();
+        this.length.valueHasMutated();
```
humitos, line 454: Interesting. It's the first time I see `.valueHasMutated()`. I quickly tried to find the documentation for it but I wasn't able to 🤷🏼 

I understand that we tell Knockout here that the value that comes from the API is _automatically_ re-calculated internally in the object, right?

src/js/build/detail.js
```diff
@@ -414,62 +434,69 @@ export class BuildDetailView {
...
+        // This is a mock command used to preview the command output.
+        // TODO probably do this in the application instead
+        this.add_command({
+          id: 0,
+          command: "readthedocs-build --show-config",
+          output: JSON.stringify(data.config, null, "  "),
+          exit_code: 0,
+          run_time: 0,
+          is_debug: true,
+        });
```
humitos, line 465: We are now doing `cat readthedocs.yaml`, so we can probably remove this hack.

humitos (approved):
Looks great to me!

readthedocsext/theme/templates/builds/build_detail.html
agjohnson: Yeah, I always figured we just drop it, I doubt anyone is looking at 8 year old builds.

PR #503, opened 2024-10-02
Builds: follow the same icon/text pattern for "Wrap output"

humitos:
This PR follows the same pattern we have in the other UI elements instead of using a checkbox (which is the only one on that page).

Also, adds an icon to this element as the other elements in the same row.

Peek 2024-10-02 14-20


agjohnson:
I intentionally wanted this to be a checkbox for now, and was thinking of this UI specifically:

image

Same goes for the debug option eventually.

The problem here is that the active state on the tab doesn't quite communicate that there is an option available to enable/disable, especially because the other options are for navigation. It's also not clear if either option is selected or not selected by default. The intention was to eventually move stateful options out of this menu entire and keep them checkboxes. I'm aiming for UI like one of these:

image

humitos:
Closing, as we are going in another direction here.

PR #508, opened 2024-10-15
Add Visit docs button to header

ericholscher:
This gives a basic way to get to the docs like the old dashboard. I was doing a bunch of local dev on the new dashboard and kept getting annoyed at not being able to get to the docs in an obvious way. 

I think this is important and is a middle ground until we can work on a fancier verison.

Screenshot 2024-10-15 at 7 21 13 AM


agjohnson:
The reason I have been -1 on this approach is because it creates two view docs buttons on the build detail page. The top most button will go to the wrong docs too (latest) instead of the PR build/branch docs for the build.

I still think this should be a singular button

agjohnson:
This button will eventually move to the project wide header, here:

https://github.com/readthedocs/ext-theme/blob/82db2330003d46ca9e27ea4120b8947b68f82d11/readthedocsext/theme/templates/projects/partials/header.html#L188

A temporary solution until someone wants to make this into a dropdown would be to move the current view docs button in the build detail page up to a `{% block project_header_navigation_docs %}` and override this blck on the build detail page. This way at least a singe button doesn't make for bad UX on the build detail page and it points to a reasonable version on all of the project pages.

I still feel a dropdown is the best UX here and is my overall goal still.

ericholscher:
For now, I've just hidden the View docs button on build pages, which seems like a reasonable approach?

agjohnson:
We talked through this one in person, I mentioned:

- The top latest view docs button needs to react to the project not having a successful build
- The lower view docs should move up, so there is one button. However, this button state is driven by the surrounding KO view and build detail API response.
- The current, per-version view docs button also needs to react to build status, so can't be a simple link either

The answer here does need to be more complex. A web component fits here short and long term, I'm poking something more flexible now.

ericholscher:
I think this is still worth getting in, especially now that we are pointing folks to this. I think my current PR solves this problem in a simple way for now, and we can do the v2 later when we have more time to dive into it?

ericholscher:
Ping @agjohnson -- :+1: on shipping this simple version for now, and we can work to make it the better smart version in the future?

ericholscher:
I ran into this a bunch at the sprints, and think we really need to solve this. Merging it, and we can come up with a better solution later.

Issue #501, opened 2024-10-01
Update 2fa templates


humitos:
There was a new release of allauth that updates some (3) templates. We need to update ours to match those changes.

- Reference: https://github.com/readthedocs/readthedocs.org/pull/11625#discussion_r1780817896

stsewd:
We need to unpin allauth after this is done

agjohnson:
I believe I already implemented all of these, I didn't see this issue but think this covers this:

- #480 

The only thing I didn't implement yet is a link to the main view from the admin nav. Once we're tested this can be added.

stsewd:
This is related to a new change in allauth, mainly from this PR https://codeberg.org/allauth/django-allauth/pulls/4111/files

PR #511, opened 2024-10-28
allauth: logout when canceling 2fa

stsewd:
Closes https://github.com/readthedocs/ext-theme/issues/501


readthedocsext/theme/templates/mfa/authenticate.html
```diff
@@ -22,12 +22,22 @@
...
+      <button class="ui button" type="submit" form="logout-from-stage">
+        {% trans "Cancel" %}
+      </button>
       <div class="right menu">
         <button class="ui primary button" type="submit">{% trans "Verify" %}</button>
       </div>
     </div>
   </form>
+
+  <!-- We can't have nested forms, so we need to have a separate form for "Cancel" -->
```
agjohnson, line 35: Avoid using HTML comments unless we really need the output in the final HTML. Template comments should be used instead.

```suggestion
  {# We can't have nested forms, so we need to have a separate form for "Cancel" #}                                                                    
```

agjohnson (approved):
This looks good at a code level, but can you post a screen grab of the UI to verify?

stsewd:
### before

Screenshot from 2024-10-31 12-59-13

### after

Screenshot from 2024-10-31 12-58-53

The outline changed

readthedocsext/theme/templates/mfa/authenticate.html
```diff
@@ -22,12 +22,22 @@
     {% csrf_token %}
     {{ form|crispy }}
     {{ redirect_field }}
-
     <div class="ui stackable relaxed text menu">
-      <a class="item" href="{% url "account_login" %}">{% trans "Cancel" %}</a>
+      <button class="ui button" type="submit" form="logout-from-stage">
+        {% trans "Cancel" %}
+      </button>
```
agjohnson, line 28: For now, I'd like to retain that any time we use a left link like this, it _is not a button styling_. I might want to change this in the future, but for now, this has been the pattern on other UIs.

I'd move this button into the `.right.menu`. That would give a button cluster similar to a modal.

stsewd:
Screenshot 2024-11-04 at 14-16-32 Verify sign in - Read the Docs

readthedocsext/theme/templates/mfa/authenticate.html
```diff
@@ -22,7 +22,7 @@
     {% csrf_token %}
     {{ form|crispy }}
     {{ redirect_field }}
-    <div class="ui stackable relaxed text menu">
+    <div class="ui right floated text menu">
```
agjohnson, line 25: This is different than what I was describing, and probably has some negative effects.

You should move the button to the `div.right.menu` directly below the button. A right float menu probably displays poorly on mobile viewports.

stsewd:
Screenshot 2024-11-05 at 10-39-26 Verify sign in - Read the Docs


Screenshot 2024-11-05 at 10-39-16 Verify sign in - Read the Docs

agjohnson (approved):

PR #512, opened 2024-10-28
Account settings: link to 2FA page

stsewd:
Screenshot 2024-10-28 at 11-38-16 Change password - Read the Docs
Screenshot 2024-10-28 at 11-38-05 Two-factor authentication - Read the Docs

### DIFF
--- a/readthedocsext/theme/templates/mfa/base_manage.html
+++ b/readthedocsext/theme/templates/mfa/base_manage.html
@@ -13,16 +13,13 @@
   active
 {% endblock profile_admin_mfa %}
 {% block edit_content_header %}
-  {% trans "Security" %}
+  {% trans "Two-factor authentication" %}
 {% endblock edit_content_header %}
 {% block edit_content_subheader %}
   <div class="ui breadcrumb">
-    <a href="{% url "mfa_index" %}" class="section">
-      {% trans "Two-factor authentication" %}
-    </a>
+    <a href="{% url "mfa_index" %}" class="section">{% trans "Manage" %}</a>
     <span class="divider">/</span>
     {% block edit_content_subheader_section %}
-      <div class="active section">{% trans "Manage" %}</div>
     {% endblock edit_content_subheader_section %}
   </div>
 {% endblock edit_content_subheader %}
--- a/readthedocsext/theme/templates/mfa/base_manage.html
+++ b/readthedocsext/theme/templates/mfa/base_manage.html
@@ -9,6 +9,9 @@
 {% block profile_admin_security %}
   active
 {% endblock profile_admin_security %}
+{% block profile_admin_mfa %}
+  active
+{% endblock profile_admin_mfa %}
 {% block edit_content_header %}
   {% trans "Security" %}
 {% endblock edit_content_header %}
--- a/readthedocsext/theme/templates/profiles/base_edit.html
+++ b/readthedocsext/theme/templates/profiles/base_edit.html
@@ -21,8 +21,16 @@
              href="{% url 'profiles_profile_edit' %}">{% trans "Account" %}</a>
           <a class="item {% block profile_admin_change_email %}{% endblock %}"
              href="{% url 'account_email' %}">{% trans "Email addresses" %}</a>
-          <a class="item {% block profile_admin_security %}{% endblock %}"
-             href="{% url 'account_change_password' %}">{% trans "Security" %}</a>
+
+          <div class="item {% block profile_admin_security %}{% endblock %}">
+            <div>{% trans "Security" %}</div>
+            <div class="menu">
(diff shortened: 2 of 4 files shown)


readthedocsext/theme/templates/profiles/base_edit.html
```diff
@@ -21,8 +21,16 @@
...
-          <a class="item {% block profile_admin_security %}{% endblock %}"
-             href="{% url 'account_change_password' %}">{% trans "Security" %}</a>
+
+          <div class="item {% block profile_admin_security %}{% endblock %}">
+            <div>{% trans "Security" %}</div>
+            <div class="menu">
+              <a class="item {% block profile_admin_change_password %}{% endblock %}"
+                 href="{% url 'account_change_password' %}">{% trans "Change password" %}</a>
+              <a class="item {% block profile_admin_mfa %}{% endblock %}"
+                 href="{% url 'mfa_index' %}">{% trans "Two-Factor authentication" %}</a>
```
agjohnson, line 31:
```suggestion
                 href="{% url 'mfa_index' %}">{% trans "Two-factor authentication" %}</a>
```

readthedocsext/theme/templates/profiles/base_edit.html
```diff
@@ -21,8 +21,16 @@
              href="{% url 'profiles_profile_edit' %}">{% trans "Account" %}</a>
           <a class="item {% block profile_admin_change_email %}{% endblock %}"
              href="{% url 'account_email' %}">{% trans "Email addresses" %}</a>
-          <a class="item {% block profile_admin_security %}{% endblock %}"
-             href="{% url 'account_change_password' %}">{% trans "Security" %}</a>
+
+          <div class="item {% block profile_admin_security %}{% endblock %}">
+            <div>{% trans "Security" %}</div>
```
agjohnson, line 26: This should be a `header` element

```suggestion
            <div class="header">{% trans "Security" %}</div>
```

stsewd: With this class, the text is always in bold, even when no selected, are we okay with that?

Screenshot from 2024-10-28 14-24-19

agjohnson: Yes, this is how the project admin looks now too.

image

Eventually all options should be grouped like the project admin settings. If you'd like to match, you could throw all other options under a "General" header for now, or split them up more.

stsewd: What about

Screenshot 2024-10-28 at 16-15-34 Connected services - Read the Docs

I'd say that security log and api tokens should also be in the security section.

stsewd: Also, with this change, do we still want to have the category as main header on each option page? e.g (Settings\nTwo-factor authentication)

agjohnson: Yeah, I'd agree on API tokens and security log under `Security` header.

> Also, with this change, do we still want to have the category as main header on each option page?

For the 2fa page, this makes sense. I'd alter the main heading and sub header breadcrumbs a little bit in order to make this page more top level, and would aim for:

image

Where the first breadcrumb links back to the main page.

I'm aiming to eventually standardize the breadcrumb sub header on all pages.

stsewd:
This is how it looks like with the latest changes

Screenshot 2024-10-28 at 17-00-56 Connected services - Read the Docs

Screenshot 2024-10-28 at 17-01-07 Two-factor authentication - Read the Docs

Screenshot 2024-10-28 at 17-01-30 Recovery codes - Read the Docs

(codes are from my dev instance)

readthedocsext/theme/templates/mfa/base_manage.html
```diff
@@ -13,16 +13,13 @@
...
 {% endblock edit_content_header %}
 {% block edit_content_subheader %}
   <div class="ui breadcrumb">
-    <a href="{% url "mfa_index" %}" class="section">
-      {% trans "Two-factor authentication" %}
-    </a>
+    <a href="{% url "mfa_index" %}" class="section">{% trans "Manage" %}</a>
     <span class="divider">/</span>
     {% block edit_content_subheader_section %}
-      <div class="active section">{% trans "Manage" %}</div>
```
agjohnson, line 22: You'll still want something here as there is a divider element right above here, so on the main page, the breadcrumb is going to render with nothing after, as:

`Manage / `

The best thing to do is to override all of `edit_content_subheader` to be just the active section:

```
<div class="ui breadcrumb">
  <div class="active section">Manage</div>
</div>
```

This is what is done on other views, so that the `<a class="section">Manage</a>` seems to link to a page where `Manage` is now the active section.

agjohnson (approved):
Looks great! I am excited for this one, this was a big upgrade on the project admin page.

PR #516, opened 2024-10-31
Rename context_text to content_text

stsewd:
The remove_button partial expects a variable named `content_text`, but we were passing `context_text` instead. It was working before because all variables from the current template are passed to the partial.

### DIFF
--- a/readthedocsext/theme/templates/account/email.html
+++ b/readthedocsext/theme/templates/account/email.html
@@ -6,7 +6,9 @@
 {% block title %}
   {% trans "Email addresses" %}
 {% endblock title %}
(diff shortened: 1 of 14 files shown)


agjohnson (approved):
Good catch! Note that this introduced a lot of lint failures

stsewd:
> Note that this introduced a lot of lint failures

yeah, hopefully I didn't break anything while fixing them... or if I did, hopefully tests on .org/.com will catch them.

agjohnson:
Yeah, not sure what tests we have for loading views and templates currently. Anything that renders these templates should trigger exceptions though.

PR #519, opened 2024-11-04
Add close/cancel button to all modals

### LINKED ISSUE #510: Modals are missing a cancel/close button
When there is a destructive option, we are showing a modal, which is great. But we aren't showing a "cancel" or "close" button, so there is no much information about how to close the modal. Pressing escape or clicking outside the modal closes the modal, but that may not be obvious to everyone.

Screenshot 2024-10-28 at 10-50-58 test-builds-2 - Redirects - Read the Docs

stsewd:
The only modal that I didn't put a button on is

https://github.com/readthedocs/ext-theme/blob/f45f878b513633b1ee395c9e89d6e3bcf4d2884c/readthedocsext/theme/templates/organizations/partials/owners_list.html#L24-L36

since it has the form as content, doesn't have an "actions" div.

Closes #510


agjohnson (approved):
This must have been removed from my review list, I just came across it. Looks good! There are some conflicts here with some recent work I've done, I'll handle merging this all up.

PR #521, opened 2024-11-06
Addons: form for tooltips

humitos:
Screenshot_2024-11-06_10-52-12

Required by https://github.com/readthedocs/addons/pull/379
Required by https://github.com/readthedocs/readthedocs.org/pull/11746
Closes https://github.com/readthedocs/addons/issues/239


agjohnson (commented):
The technical changes here look fine, but I raised a point about the naming here. I would not call this element a "tooltip", this is a specific term for this element:

image

Noted in chat, but this is a clear review of the terms here:

https://medium.com/design-bootcamp/popups-dialogs-tooltips-and-popovers-ux-patterns-2-939da7a1ddcd

humitos:
I renamed "Tooltips" by "Link Previews"

humitos:
We decided to not merge this PR _yet_ so we can start testing it ourselves with our own projects first.

agjohnson (approved):

humitos:
This PR needs to be adapted now that we have `addons.options_root_selector`.

humitos:
I updated this PR to show the "CSS main content selector" as a global option.

With the name changed to "Link previews" a new row is generated for the tabs and it doesn't render correctly. What's the solution here?

Screenshot_2024-11-27_17-57-10

readthedocsext/theme/templates/projects/addons_form.html
```diff
@@ -37,6 +37,7 @@
       </p>
 
       {{ form.enabled|as_crispy_field }}
+      {{ form.options_root_selector|as_crispy_field }}
```
agjohnson, line 40: This could still come under an `Advanced` tab and then we don't need to have any top level configuration options.

humitos: Done

agjohnson (approved):

ericholscher:
Sounds like there's a `wrapped` class we need here

agjohnson:
There is a wrapping class that the menu can use:

https://fomantic-ui.com/collections/menu.html#wrapping

However, that also looks pretty similar to what we have. The centered wrapping looks a little better though:

https://fomantic-ui.com/collections/menu.html#centered-fluid-wrapping

humitos:
It still doesn't render _perfectly_ to me, but I'll move forward with this.

Screenshot_2024-12-05_11-59-13

humitos:
Good point. We don't have specific documentation for this yet. We only have the description of the backend API and a link to hoverxref in https://docs.readthedocs.io/en/stable/guides/embedding-content.html

I will try to quickly update that page to remove the recommendation for hoverxref and link to a new addons page for links preview with small content for now.

Issue #241, opened 2023-12-07
Redirects: show/use new fields


stsewd:
We have added new fields to redirects that may be shown in the UI, or used to implement some filtering, new fields are:

- Description
- HTTP status code
- Enabled
- Position

ref https://github.com/readthedocs/ext-theme/pull/237, https://github.com/readthedocs/readthedocs.org/pull/10881/

stsewd:
There was also a discussion around how we can show each type of redirect with the new syntax we have (`*`, `:splat`), and Clean/HTML redirects.

https://github.com/readthedocs/ext-theme/pull/237#issuecomment-1845688903

PR #522, opened 2024-11-07
Add redirect description to listing page

ericholscher:
Screenshot 2024-11-07 at 1 43 39 PM

Refs https://github.com/readthedocs/ext-theme/issues/241

### DIFF
--- a/readthedocsext/theme/templates/projects/partials/edit/redirect_list.html
+++ b/readthedocsext/theme/templates/projects/partials/edit/redirect_list.html
@@ -119,6 +119,12 @@
       {% endif %}
     {% endspaceless %}
   </div>
+  {% if object.description %}
+    <div class="sub header">
+      <label>{% trans "Description:" %}</label>
+      <span>{{ object.description }}</span>
+    </div>
+  {% endif %}
 {% endblock list_item_header %}
 
 {% block list_item_meta_column %}


readthedocsext/theme/templates/projects/partials/edit/redirect_list.html
```diff
@@ -119,6 +119,12 @@
       {% endif %}
     {% endspaceless %}
   </div>
+  {% if object.description %}
+    <div class="sub header">
+      <label>{% trans "Description:" %}</label>
```
ericholscher, line 124: I'd be OK just putting this in italics and not having `Description` as a UI element, just copying the existing structure.

agjohnson: I prefer to limit the amount of information we're putting into listing display, the display of these tables already gets blown out when there is a long URL. More metadata would cause line wrapping and cram the listing with data.

I wonder if it would be better for now to use the description if there is one (and I would also drop the label), and if not, describe the redirect in the listing using the existing "Redirect to: ...". It feels like we don't need both.

This is a place where I'd use a popupcard next, so on hover of each item there is more information on each line. This solves both problems without overloading the listing view.

ericholscher: Cool.. I was just trying to make a quick improvement that made something better, but if this is a larger change, I'll go ahead and close this until we have time to dive into it deeper.

agjohnson: I think what I'm describing is still a small change, though I haven't played around with this either. I think instead of _adding_ the description to the listing, we could move the condition for `{% if object.description %}` around so that if there is a description, the description is in the listing, and if not the original `Redirect to: ....` information is in the listing instead.

PR #530, opened 2024-11-21
favicon dev

ericholscher:
- Revert "Revert "Fix button action on invite user page""
- Add favicon-dev so you can tell dev apart

### DIFF
--- a/readthedocsext/theme/templates/base.html
+++ b/readthedocsext/theme/templates/base.html
@@ -17,7 +17,13 @@
     {% endblock head_meta %}
 
     {% block head_links %}
-      <link rel="icon" type="image/png" href="{% static 'images/favicon.png' %}" />
+      {% if "devthedocs" in PRODUCTION_DOMAIN %}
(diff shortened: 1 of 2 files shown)


humitos:
@agjohnson is this missing anything or we can merge it?

agjohnson:
I mostly think there should just be one icon. In general, it's better that dev closely matches production. So, I'm not a particular fan of using a special favicon just for dev for that. I even prefer a single icon rather than differing icons for .org and .com. Browsers have profiles/multi-containers if it's really a trouble telling the tabs apart in dev.

Ultimately this should use multiple sizes though. Browsers, search engines, and phones use all the various sizes and display gets odd on some of these without these definitions.

We should match what we already do this on our website:

https://github.com/readthedocs/website/blob/19813c2bd9e1ad3ded60e6af7ac4a1da90441e3b/readthedocs_theme/templates/base.html#L15-L19

That's easier with just a single icon style.

PR #531, opened 2024-11-22
Add 403 CSRF template

ericholscher:
This is the default from Django:
https://docs.djangoproject.com/en/5.1/ref/settings/#csrf-failure-view

I'm actually not sure where these error templates are configured to be used, so not 100% sure it will work.

However, we currently show the default 403 template in prod, which is bad:

Screenshot 2024-11-22 at 8 51 31 AM

### DIFF
--- a/readthedocsext/theme/templates/errors/dashboard/403_csrf.html
+++ b/readthedocsext/theme/templates/errors/dashboard/403_csrf.html
@@ -1,33 +1,6 @@
-{% extends "errors/dashboard/base.html" %}
+{% extends "errors/dashboard/403.html" %}
 
 {% load blocktrans trans from i18n %}
-
-{% block title %}
-  {% trans "Access restricted" %}
-{% endblock title %}
-
-{% block error_status %}
-  403
-{% endblock error_status %}
-{% block error_title %}
-  {{ block.super }}
-  {% trans "Access restricted" %}
-{% endblock error_title %}
-{% block error_subtitle_text %}
-{% endblock error_subtitle_text %}
-
-{% block error_icon_computer_name %}
-  fa-lock-keyhole
-{% endblock error_icon_computer_name %}
-{% block error_icon_mobile_name %}
-  fa-lock-keyhole
-{% endblock error_icon_mobile_name %}
-{% block error_icon_computer_style %}
-  --fa-primary-color: orange;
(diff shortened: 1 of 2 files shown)


readthedocsext/theme/templates/errors/dashboard/403_csrf.html
```diff
@@ -0,0 +1,48 @@
...
+{% endblock error_icon_computer_name %}
+{% block error_icon_mobile_name %}
+  fa-lock-keyhole
+{% endblock error_icon_mobile_name %}
+{% block error_icon_computer_style %}
+  --fa-primary-color: orange;
+{% endblock error_icon_computer_style %}
+{% block error_icon_mobile_style %}
+  --fa-primary-color: orange;
+{% endblock error_icon_mobile_style %}
```
agjohnson, line 30: Instead of duplicating all this you can extend 403.html so the pages are kept in sync

```suggestion
{% extends "errors/dashboard/403.html" %}

{% load blocktrans trans from i18n %}
```

agjohnson (approved):
The easiest way to use this template is to add a symlink from `templates/403_csrf.html` to this file.

The harder way is to override the view used for CSRF failure:
https://github.com/django/django/blob/857b1048d53ebf5fc5581c110e85c212b81ca83a/django/views/csrf.py#L22

ericholscher:
@agjohnson is there a place we're using that symlink pattern? I don't see any of those in ext_theme, so confused how they're getting used.

agjohnson:
We do use symlinks already in the error templates path, but the base templates are controlled with URL pattern `handler400` and friends:

https://github.com/readthedocs/readthedocs.org/blob/ebe3b2c7f0ca9f206b2fed0c8e53269f18e62774/readthedocs/urls.py#L16-L19

You don't have that option here though, from what I can tell. CSRF doesn't seem to follow that pattern.

ericholscher:
Cool, added it to the path. Not 100% sure the best way to test this locally tho.

agjohnson:
You have to run the environment with ``--no-debug`` to test any of these handlers. The debug handlers override these in debug mode.

ericholscher:
Looks great 💯 

Screenshot 2024-12-02 at 5 27 14 PM

PR #532, opened 2024-11-23
Replace project and version KO views

### LINKED ISSUE #444: Documentation link sometimes does nothing
I found myself clicking on the 📚 icon to access the documentation of that particular version/project but it just does nothing. Then, the second time I click on it, it works just fine.

I think the issue is that we are performing a request behind the scenes when the mouse hovers the icon and sometimes the response is not immediate [^1]. I thought a good idea here would be to perform the fetch when the user hovers any section of _the row_ itself, not only the books icon.

[^1]: Are we caching these requests? I feel they should be _immediate_ most of the times.

### LINKED ISSUE #497: User profile page exception clicking links to docs
TBD, but the error referenced a missing configInit, likely the configuration json wasn't passed in.

agjohnson:
This solves a few issues and explores a pattern that we need to figure out in
order to replace Knockout views with web components: authoring and connection
between nested web components. This is a low stakes place to try this.

The issues around the buttons were addressed through adding a click wrapper to
the link. If you accidentally trigger the link before the API response, the
click event is stored and re-emitted after the response finishes.

The pattern being explored here is using the web component context API. This is
implemented in Lit already. You can read more here:

- https://lit.dev/docs/data/context/
- https://github.com/webcomponents-cg/community-protocols/blob/main/proposals/context.md

So far, we have been trying out manually using events to pass around data at
a global level. The context API functions similar under the hood, but the API is
a much cleaner way to pass data around, and it functions only between parent and
child elements, not globally.

Where this differs from global events is that we can use this pattern between
parent and child elements, and we don't need to manually search the DOM for the
parent element:

```html
<readthedocs-api url="...">
  <div>
    <div>
      <div>
        <readthedocs-item-docs>
          This element will traverse up the tree and listen to only the first
          readthedocs-api provider
        </readthedocs-item-docs>
      </div>
    </div>
  </div>
</readthedocs-api url="...">
<readthedocs-api url="...">
  <readthedocs-item-docs>
    And this element will only listen for data provided by that second
    readthedocs-api provider
  </readthedocs-item-docs>
</readthedocs-api>
```

## Why?

We want to author as little as we need to in JS, and as much as we can in
templates. Using the context API allows us to author isolated web components
that share data _while still using Django templates for as much of our page
structure as we can_.

The alternative is authoring everything nested in `readthedocs-api` in a big
web component, which is not a good pattern for a Django backend.

The work here so far this is a very basic example and the provider/consumer
elements are close in proximity in the DOM. A more complex example would be a
table listing, where the table row is a data provider and elements nested
anywhere in that row element can share data.

We want to keep as much of that row in Django templates as possible, especially
if that row is 95% vanilla HTML elements and 5% dynamic web components.

----

Things I did not do here:

- Deprecate the older elements using the same base
- Remove old project/version KO views
- Split up code

This will come next.

----

- Fixes #444
- Fixes #497


agjohnson:
There is no significant visual change to the menus:

[Screencast From 2024-11-22 17-12-37.webm](https://github.com/user-attachments/assets/63f5089e-e869-4c24-8a26-a4380267a4ea)

There is a loading indicator at the moment, I asked for feedback but didn't get any so left it as is. It's not important to this work and the queued event hides this state well enough.

src/js/modules/menus.js
```diff
@@ -122,3 +149,264 @@ customElements.define(
...
+  }
+
+  /** @type {ContextProvider} Reactive data context for API response state **/
+  _providerState = new ContextProvider(this, {
+    context: contextState,
+  });
+
+  set state(value) {
+    this._providerState.setValue(value);
+  }
```
agjohnson, line 209: And the one thing I am leaving for now but would like to think about more is what the data structure looks like between elements. It seems like we don't need individual `state` provider and could do use an `Object` to pass around a complex data structure.

However we also can't change individual properties through `this._providerData.setValue()`. This means to set `state` we have to pass all properties through at the same time.

package.json
```diff
@@ -23,6 +23,8 @@
     "@babel/plugin-proposal-class-properties": "^7.16.7",
     "@babel/plugin-transform-modules-commonjs": "^7.23.0",
     "@babel/preset-env": "^7.22.20",
+    "@lit/context": "^1.1.3",
+    "@lit/localize": "^0.12.2",
```
humitos, line 27: Hrm, interesting. Aren't these installed automatically with `lit` already?

humitos (commented):
I didn't take a deep look at the code yet, but the description of the PR and the pattern followed seems good at first sight. I may need to read more about Lit context and follow the code more closely to be able to do a good review here.

package.json
agjohnson: They aren't, both of these are standalone packages. Installation directions are here:

https://lit.dev/docs/localization/overview/

src/js/modules/menus.js
```diff
@@ -1,17 +1,22 @@
...
- *  Handles error feedback via toast message and redirection on success.
+ * Generic button/link wrapper to POST to a URL on an event like "click".
+ * Handles error feedback via toast message and redirection on success.
  *
  * @param {string} csrfToken - CSRF token from Django, attribute ``csrf-token``
  * @param {string} url - API URL for request
  *
+ * TODO Deprecate this in favor of APIProviderElement
  **/
 export class APIEventWrapper extends LightDOMElement {
```
humitos, line 21: I'm confused with these changes. I understand we are not requiring `APIEventWrapper` anymore for these buttons/items now -- but we still want it for some other places where we are using it. Why do we need to modify its code then? 🤔

src/js/modules/menus.js
```diff
@@ -122,3 +149,264 @@ customElements.define(
...
+ * consumed by :js:class:`APIConsumerElement`. This allows a central element to
+ * execute an API request (table row, or a group of buttons) and for sub elements
+ * at any point in the nested DOM to use this response (a button nested in the
+ * table row, a submenu in the group of buttons).
+ **/
+const contextData = createContext(Symbol("data"));
+const contextState = createContext(Symbol("state"));
+
+/**
+ * API resopnse states enum
```
humitos, line 166:
```suggestion
 * API response states enum
```

src/js/modules/menus.js
```diff
@@ -122,3 +149,264 @@ customElements.define(
...
+const States = Object.freeze({
+  WAITING: Symbol("waiting"),
+  LOADING: Symbol("loading"),
+  DONE: Symbol("done"),
+});
+
+/**
+ * Context provider element for API response
+ *
+ * On an event (mouseover by default), this element starts an API request to the
```
humitos, line 179: Should we use `focusin` as well to support keyboard?

src/js/modules/menus.js
```diff
@@ -122,3 +149,264 @@ customElements.define(
...
+  }
+
+  /** @type {ContextConsumer} Reactive consumer for API request state context data **/
+  _consumerState = new ContextConsumer(this, {
+    context: contextState,
+    subscribe: true,
+  });
+
+  /** @type {State} Reactive context data for API request state **/
+  get state() {
```
humitos, line 271: Interesting... I wasn't aware of this `get/set state` syntax. 

* https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get
* https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/set

src/js/modules/menus.js
humitos: Yeah, there may be a reason why they are not using an `Object` in their examples. Maybe we can find other examples with more requirements where they use an `Object` for this to get some ideas, pros and cons.

src/js/modules/menus.js
```diff
@@ -122,3 +149,264 @@ customElements.define(
...
+export class ItemDocsElement extends APIConsumerElement {
+  render() {
+    let label = this.label || msg(`View documentation`);
+    return html`
+      <a
+        class="ui button ${classMap({
+          disabled: this.disabled,
+          loading: !this.disabled && this.state === States.LOADING,
+        })}"
+        href=${this.data?.urls?.documentation || `#`}
```
humitos, line 319: JavaScript question: there is something that I'm don't follow completely here.

What happens when we access `this.data` but the API request hasn't returned yet or hasn't even been triggered? I understand that `this.data` will automatically trigger the `context-request` event from the `ContextController`, right? 🤔 

Will this just be `null`/`undefined`? That's it? Or there is something else going on here that I'm missing?

readthedocsext/theme/templates/projects/partials/project_list.html
```diff
@@ -29,41 +29,20 @@
...
-  </div>
+  <readthedocs-api csrf-token="{{ csrf_token }}"
+                   url="{% url "projects-detail" object.slug %}?expand=permissions">
+    <div class="ui small icon buttons">
+      <readthedocs-item-docs {% if not object.has_good_build %}disabled{% endif %}
+                             label="{% blocktrans with name=object.name %}View documentation for project {{ name }}{% endblocktrans %}">
+      </readthedocs-item-docs>
+      <button class="ui dropdown button">
+        <i class="fa-solid fa-ellipsis icon"></i>
+        <readthedocs-menu-project-admin class="menu" settings="{% url "projects_edit" object.slug %}">
```
humitos, line 41:
```suggestion
        <readthedocs-menu-project-admin class="menu" settings-url="{% url "projects_edit" object.slug %}">
```

I think `settings-url` is easier to read and follow. I was confused about what `settings` was when reading the JavaScript code, and I had to come back here to understand its value.

src/js/modules/menus.js
```diff
@@ -122,3 +149,264 @@ customElements.define(
...
+    let _properties = APIConsumerElement.properties;
+    _properties["settings"] = { type: String };
+    return _properties;
+  }
+
+  render() {
+    const isAdmin = this.data?.permissions?.admin;
+    // This preemptively supports a settings URL, which can be passed in as an
+    // attribute in the meantime. The `settings` property can go away with this.
+    const urlSettings = this.data?.urls?.settings || this.settings;
```
humitos, line 396: Commenting also here since I found myself confused here.
```suggestion
    const urlSettings = this.data?.urls?.settings || this.settingsUrl;
```

humitos (approved):
It took me _a lot_ to review this PR because I didn't have the required Lit knowledge. I read the documentation you shared and I think this implementation is better than what we had.

I feel a lot more comfortable with Lit than with Knockout.

I left a few questions 😄

src/js/modules/menus.js
agjohnson: This is a reactive property, so it will have a varying state. Initially, `this.data` is going to be `null` and the `href=#`.

Eventually, `this.data` is updated by the context API and this will trigger a reactive property update and will re-render the element with the data and proper URL.

Just accessing `this.data` doesn't trigger any updates though, the value is just reflected from the parent element.

src/js/modules/menus.js
agjohnson: Yeah, I wanted to just show `click` for now, it should handle an array of events though.

src/js/modules/menus.js
agjohnson: The plan would be to use this new class instead. I only extended the parent class so I wasn't yet changing these elements too.

readthedocsext/theme/templates/builds/partials/version_list.html
```diff
@@ -34,83 +34,43 @@
...
+      {% if is_project_admin %}
+        <div class="ui dropdown button"
+             data-content="{{ label_admin }}"
+             aria-label="{{ label_admin }}">
+          <i class="fa-solid fa-ellipsis icon"></i>
+          <div class="menu">
+            <div class="header">{% trans "Admin" %}</div>
+            <readthedocs-menu-build-rebuild class="item"
+                                            url="{% url "projects-versions-builds-list" object.project.slug object.slug %}"
+                                            csrf-token="{{ csrf_token }}">
```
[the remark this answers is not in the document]
agjohnson: It should, I didn't want to refactor any other elements just yet though.

src/js/modules/menus.js
agjohnson: Added this to #535 for follow up soon

readthedocsext/theme/templates/builds/partials/version_list.html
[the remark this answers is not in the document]
agjohnson: Added to #535

src/js/modules/menus.js
agjohnson: I added this to #535 but it might not be a strong priority still.

agjohnson:
Great, and thanks @humitos for taking the time with the Lit API. I wanted to see how you felt about this vs KO especially, so it's good to know that it does feel like an improvement. I'd like to see how this pattern works out more before changing anything big, but it's a start.

Issue #71, opened 2021-11-18
Build: support terminal escape/control codes and color codes


agjohnson:
ext-theme is fancy and can handle ANSI codes in build command output if we would like. To support this, we need to TTY handling of build output and need to remove the NO_COLOR env variable.

- Previous attempt: https://github.com/readthedocs/readthedocs.org/issues/8733
- Also see: https://github.com/readthedocs/readthedocs.org/issues/6883

----

- [ ] Support tty output and control characters at the build level
- [ ] Restructure the build command output to first load ansiup and sanitize-html libraries. Currently, the build detail JS expects data rather immediately, so async import loading actually breaks a number of features -- linking to a command line, expanding a command line, etc.
- [ ] The `BuildCommand` handling of `output_lines` could be restructured a little bit to load the observables nicely
- [ ] The HTML template should use the `html` binding when sanitize-html is in place.

PR #537, opened 2024-12-05
Use dompurify to sanitize HTML on notifications

### LINKED ISSUE #322: Notifications: use sanitize-html for cleaning header/body content
This was a library that was brought in for displaying ANSI color codes in the build detail command output. It is currently unused, but was loaded async previously. It seems we could use this for notifications as well, making it a standard part of our bundled JS. Or, alternatively, load it async when there are notifications.

We only need to support some minor elements in notification body/header, mostly formatting and links.

- For background on this feature on the build detail page #71 

https://github.com/readthedocs/ext-theme/blob/bd12f154ae54d3d0c92bc37d4415a661b75552ef/src/js/build/detail.js#L153-L179

agjohnson:
This replaces sanitize-html as this package was a commonjs package and
it created some errors in tests.

- Fixes #322


src/js/build/detail.js
```diff
@@ -150,31 +150,16 @@ class BuildCommand {
...
       // Build output lines
       let ansi_up = new AnsiUp();
       ansi_up.use_classes = true;
       output = ansi_up.ansi_to_html(output);
-      output = sanitize_html(output, {
-        allowedTags: ["span"],
-        allowedAttributes: { span: ["class"] },
-      });
+      // TODO use dompurify here
+      //output = DOMPurify.sanitize_html(output);
```
humitos, line 162: Is there anything blocking us to use DOMPurify here or you missed un-commenting these lines after testing out something else? 😄

humitos: If something is blocking us, it would be good to update the comment with the reasons.

humitos (approved):

src/js/build/detail.js
agjohnson: This function is not used because we are not coloring output. I have some issues open for this.

- https://github.com/readthedocs/ext-theme/issues/71
- https://github.com/readthedocs/readthedocs.org/issues/8733

PR #540, opened 2024-12-10
Remove weird spacing from titles

ericholscher:
This shows up in slack oddly

Screenshot 2024-12-10 at 8 38 24 AM

### DIFF
--- a/readthedocsext/theme/templates/base.html
+++ b/readthedocsext/theme/templates/base.html
@@ -4,12 +4,12 @@
 <!DOCTYPE html>
 <html lang="en">
   <head>
-    <title>
-      {% block title %}
-      {% endblock title %}
-      {% block head_title %}
-      {% endblock head_title %}
-    - Read the Docs</title>
+
+    {# Hide weird spaces in title #}
+    {# djlint:off #}
+    <title>{% block title %}{% endblock title %}{% block head_title %}{% endblock head_title %} - Read the Docs</title>
+    {# djlint:on #}
+
     {% block head_meta %}
       <meta name="description"
             content="Read the Docs is a documentation publishing and hosting platform for technical documentation" />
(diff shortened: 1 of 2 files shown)


readthedocsext/theme/templates/base.html
```diff
@@ -4,12 +4,12 @@
...
-      {% block title %}
-      {% endblock title %}
-      {% block head_title %}
-      {% endblock head_title %}
-    - Read the Docs</title>
+
+    {# Hide weird spaces in title #}
+    {# djlint:off #}
+    <title>{% block title %}{% endblock title %}{% block head_title %}{% endblock head_title %} - Read the Docs</title>
+    {# djlint:on #}
```
stsewd, line 11: Maybe this template tag can help https://docs.djangoproject.com/en/4.2/ref/templates/builtins/#spaceless

stsewd: hmm, maybe not

> Only space between tags is removed – not space between tags and tex

agjohnson: Spaceless is handy for these scenarios, but yeah doesn't work without more elements. Usually `spaceless` still requires some extra `span` elements around the text, which if `span` is legal inside `title`, would be something like:

```
{% spacesless %}
  <title>
    <span>{% block title %}{% endblock title %}</span>
    <span>{% block head_title %}{% endblock head_title %}</span>
  </title>
{% endspaceless %}
```

This doesn't solve inner whitespace, but should collapse otherwise.

The hack in this PR is okay if `span` or other elements aren't valid in `title`.

agjohnson (approved):
I noted how to normally use spaceless filter, but this is okay if that won't work. This is another area I just wish we just switched to Jinja instead.

readthedocsext/theme/templates/base.html
ericholscher: > Titles may contain character entities (for accented characters, special characters, etc.), but may not contain other markup (including comments).


https://stackoverflow.com/questions/12168806/is-it-allowed-to-use-other-tags-inside-title

PR #541, opened 2024-12-10
Extend CSRF 403 from Proxito

ericholscher:
We do trigger CSRF errors from Proxito, so use the more minimal error page.

Fixes https://read-the-docs.sentry.io/issues/6108273096/?project=148442&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&sort=date&statsPeriod=14d&stream_index=7

### DIFF
--- a/readthedocsext/theme/templates/403_csrf.html
+++ b/readthedocsext/theme/templates/403_csrf.html
@@ -1,1 +1,1 @@
-errors/dashboard/403_csrf.html
+errors/proxito/403_csrf.html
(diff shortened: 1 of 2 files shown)


agjohnson:
That would also match all of the other patterns for error views too. The CSRF view in Django is a one-off contraption.

readthedocsext/theme/templates/errors/dashboard/403_csrf.html
```diff
@@ -1,4 +1,4 @@
-{% extends "errors/dashboard/403.html" %}
+{% extends "errors/proxito/403.html" %}
```
agjohnson, line 1: You also should move the file so we're not mixing templates. Proxito error pages are consolidated in `/errors/proxito/`

agjohnson (approved):

PR #543, opened 2024-12-12
Fixes for popupcards and project avatar in popupcard

### LINKED ISSUE #389: Build popup should be smaller and have more padding when loading
This is a pretty minor issue.

I noticed the popup shown when hovering on a build shows a spinner icon that I think it would look better by reducing the width of it and increasing the padding for the icon so it doesn't touch the borders.

Peek 2024-06-19 13-51

agjohnson:
- Fixes #389
- Fixes issue with project popupcard, where we were still using Dicebear

----

- Reorder data binding in popupcards to use loading state on correct element
- Update project popupcard to use readthedocs-avatar instead of dicebear
- Alter CSS rules to support readthedocs-avatar in FUI better

[Screencast From 2024-12-11 20-10-34.webm](https://github.com/user-attachments/assets/e7311057-a6f4-4d13-b92b-f561e18bf266)

image


humitos:
Taking a look at the video, it seems the books icon disappear once the popup finishes loading for some reason. Then, they re-appear and disappear again 🤷🏼

agjohnson:
This is likely an artifact of using network throttling to slow down the API request and UI updates, it should not be related to this change.

PR #548, opened 2024-12-16
Support past tense in invitations

### LINKED ISSUE #242: Invites: support past tense on invite lists
I noticed this on the organization team member invite list, but it probably is a problem everywhere we list invites. Invites that have expired are shown as "Expires X days ago", which should at least be "Expired X day ago". The listing should also show some disabled state on the invite to make it more clear. The single letter change is easy to miss.

image

agjohnson:
- Fixes #242 

image


ericholscher:
Another vote for linting the whole set of templates in 1 commit.

ericholscher (approved):

PR #561, opened 2025-01-28
Warn about changing a version slug

stsewd:
Ref https://github.com/readthedocs/readthedocs.org/pull/11930

### DIFF
--- a/readthedocsext/theme/templates/projects/version_form.html
+++ b/readthedocsext/theme/templates/projects/version_form.html
@@ -53,7 +53,7 @@
           <div class="content">
             {% if version.active %}
               <div class="ui warning message">
-                {% trans "Deactivating a version will result in the deletion of its documentation." %}
+                {% trans "Deactivating a version or changing its slug will result in the deletion of its documentation." %}
               </div>
             {% endif %}
 


readthedocsext/theme/templates/projects/version_form.html
```diff
@@ -53,7 +53,7 @@ <h2 class="ui medium dividing header">
           <div class="content">
             {% if version.active %}
               <div class="ui warning message">
-                {% trans "Deactivating a version will result in the deletion of its documentation." %}
+                {% trans "Deactivating a version or changing its slug will result in the deletion of its documentation." %}
```
ericholscher, line 56: We should make this more visible and mention it on the slug field itself probably. This is a good start, but I think we want to be really explicit here.

humitos: I would keep it here since we already have this warning in place. However, I think we should also update the `help_text` of the slug field so it's shown on the field itself as well.

stsewd: I think we are better by just showing a pop-up in js when the slug is changed, asking for confirmation (maybe a task for someone with more frontend skills). Another option is maybe have a separate form to change the slug.

ericholscher (approved):
I think I'm OK with this change to start, if it unblocked the change for this deploy and we work to improve the warning.

PR #566, opened 2025-02-25
Add initial Settings Overview page

ericholscher:
This gives users a bit more context about their project,
and is a starting point to introduce more information.

This is a first draft, but curious to get some feedback on what should be included.
I'm a little worried about the performance in production, so will put this behind a staff flag for now.

Screenshot 2025-02-24 at 4 20 55 PM

### DIFF
--- a/readthedocsext/theme/templates/projects/edit_base.html
+++ b/readthedocsext/theme/templates/projects/edit_base.html
@@ -11,6 +11,12 @@
     <div class="five wide computer five wide tablet sixteen wide mobile column">
 
       <div class="ui vertical pointing fluid menu">
+        {# Overview link standalone before sections #}
+        <a class="item {% block project_overview_active %}{% endblock project_overview_active %}"
+           href="{% url "projects_overview" project.slug %}">
+          <i class="dashboard icon"></i>
+          {% trans "Project Overview" %}
+        </a>
 
         <div class="item">
           <div class="header">{% trans "Setup" %}</div>
--- a/readthedocsext/theme/templates/projects/project_overview.html
+++ b/readthedocsext/theme/templates/projects/project_overview.html
@@ -1,0 +1,179 @@
+{% extends "projects/project_edit_base.html" %}
+
+{% load trans blocktrans from i18n %}
+
+{% block title %}
+  {% trans "Project Overview" %}
+{% endblock title %}
+
+{% block project_edit_content_header %}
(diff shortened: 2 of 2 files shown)


agjohnson:
This has a lot of focus on stats and analytics -- this could be another page we eventually need. But, I thought we instead wanted this page to provide users with more context on where settings were and how to change them.

In particular, just showing a count of items seems not super useful. If we did want to enumerate the number of each relationship, we can include this in the existing side menu without a separate page though:

image

I'm not necessarily advocating for including a count at all, I'm not certain this is useful for every item in that menu, but it's something we could consider.

The build count and build total time should be the most prominent, those are very helpful figures that we want to bill on that don't exist anywhere right now. Traffic and search analytics are secondary at very least. I'd push these off to a separate view entirely eventually, users shouldn't need to be admins to view these.

I'd skip all of the buttons, those are redundant to just linking using the counters -- however I'd still prefer to probably use descriptive settings instead of just counters for all of the relationships.

The things we wanted out of an overview page were to display build time and build count, link the user to settings that are commonly used or hard to find, and I think it's a good use of UI to surface derivative information and settings. Some things we've wanted here:

- Derivative information
  - Number of builds in the last month
  - Build time in the last month
  - Allowed concurrency for this project
  - Time since last versions synced or last webhook push
  - Version privacy level overview
- Read only information
  - Project slug
  - Main project URL
    - Point users towards custom domains and upgrading for full control of URL
- If pull request builds are disabled, describe pull request builds and link to the UI to enable
- If any project versions are private, describe sharing methods and link to UI
- If addons are disabled, describe addons features and point to the UI, or show an overview of addons configuration otherwise
- Move delete project to the overview page as it doesn't make sense under "Settings"

The CircleCI settings do a fair job at some of this. Stripe is too much of a damn mess to make any sense of their settings, but they also at least try to explain where to change important things. I think those are both okay pages to model our off of.

humitos:
> The build count and build total time should be the most prominent, those are very helpful figures that we want to bill on that don't exist anywhere right now.

Number of custom domain is also important, since they are related to the price plan.

ericholscher:
Yea, this was just a first attempt. I can do a revamp that shows the items you were speaking about a bit more visibly, I just wanted to get _something_ started to iterate on.

ericholscher:
Second draft:

Screenshot 2025-02-25 at 3 56 19 PM

PR #569, opened 2025-03-14
Toggle to view 2FA recovery codes

davidfischer:
This adds a toggle to view the codes. Feel free to reject this if you don't think it's necessary.


agjohnson:
Yeah, this is good. FUI has some polish to the native `details` element though. I was going to suggest some of these changes but ended up with a commit trying to describe the direction to head.

The display issues were minor, mostly spacing. It looks like this now:


https://github.com/user-attachments/assets/fc84fb5f-3062-4ad3-b4e0-9f2aefe09f64

PR #570, opened 2025-03-18
Add migration page for GitHub App

stsewd:
While this still requires https://github.com/readthedocs/readthedocs.org/pull/11942/, I think it should be safe to review, as the UI shouldn't change much. I still need to finishe a couple of things: make all strings translable, and check for users with multiple GH accounts connected.

https://github.com/user-attachments/assets/a836dfaf-4f30-41ca-a84b-8a669c838aef

### DIFF
--- a/readthedocsext/theme/templates/profiles/private/migrate_to_gh_app.html
+++ b/readthedocsext/theme/templates/profiles/private/migrate_to_gh_app.html
@@ -1,0 +1,366 @@
+{% extends "base.html" %}
+
+{% load provider_login_url from socialaccount %}
+{% load trans blocktrans from i18n %}
+
+{% block title %}
+  {% trans "Migrate account to GitHub App" %}
+{% endblock title %}
+
+{% block content %}
+
+  <div class="eleven wide computer eleven wide tablet sixteen wide mobile column">
+    <h2 class="ui medium dividing header">
+      {% trans "Migrate account to GitHub App" %}
+    </h2>
+
+    <div id="edit-content">
+      <div class="ui two column grid">
+        <div class="five wide computer five wide table sixteen wide mobile column">
+          <div class="ui vertical steps">
+            <a class="step {% if step == "overview" %}active{% endif %}"
+               href="{% url 'migrate_to_github_app' %}?step=overview">
+              <div class="content">
+                <div class="title">Overview</div>
+                <div class="description">Read before starting the migration process</div>
+              </div>
+            </a>
+            <a class="step {% if step == "connect" %}active{% endif %} {% if step_connect_completed %}completed{% endif %}"
+               href="{% url 'migrate_to_github_app' %}?step=connect">
+              <div class="content">
+                <div class="title">Connect account</div>
+                <div class="description">Connect your account to our new GitHub App</div>
+              </div>
+            </a>
+            <a class="step {% if step == "install" %}active{% endif %} {% if not step_connect_completed %}disabled{% endif %}"
+               href="{% url 'migrate_to_github_app' %}?step=install">
+              <div class="content">
+                <div class="title">Install GitHub App</div>
+                <div class="description">
+                  Install the GitHub App in all your repositories connected to a project
+                </div>
+              </div>
+            </a>
+            <a class="step {% if step == "migrate" %}active{% endif %} {% if not step_connect_completed %}disabled{% endif %}"
+               href="{% url 'migrate_to_github_app' %}?step=migrate">
+              <div class="content">
+                <div class="title">Migrate projects</div>
+                <div class="description">Migrate your projects to the GitHub App</div>
+              </div>
+            </a>
+            <a class="step {% if step == "revoke" %}active{% endif %} {% if not step_connect_completed %}disabled{% endif %}"
+               href="{% url 'migrate_to_github_app' %}?step=revoke">
+              <div class="content">
+                <div class="title">Revoke access to old integration</div>
+                <div class="description">
+                  Revoke access to the old GitHub OAuth app from your GitHub account
+                </div>
+              </div>
+            </a>
+            <a class="step {% if step == "disconnect" %}active{% endif %} {% if step_disconnect_completed %}completed{% endif %} {% if not step_connect_completed or not step_revoke_completed %}disabled{% endif %}"
+               href="{% url 'migrate_to_github_app' %}?step=disconnect">
+              <div class="content">
+                <div class="title">Disconnect old integration</div>
+                <div class="description">
+                  Disconnect the old GitHub OAuth app from your Read the Docs account
+                </div>
+              </div>
+            </a>
+          </div>
+        </div>
+        <div class="column">
+          <div class="ui basic segment">
+            {% if step == "overview" %}
+              <p>
+                We’re introducing a new GitHub App to replace the old GitHub OAuth app.
+                This new app offers more granular permissions and better GitHub integration.
+                Learn more in our <a href="#">blog post</a>.
+              </p>
+
+              <p>
+                This guide will help you migrate your account to the new GitHub App.
+                Some things you should know:
+              </p>
+
+              <p>
+                <ul>
+                  <li>
+                    Migrating all projects is optional,
+                    you can skip the &quot;Install GitHub App&quot; and &quot;Migrate Projects&quot; steps.
+                  </li>
+                  <li>
+                    Connecting your account to the new GitHub App and revoking the old GitHub OAuth app is required,
+                    as the old app is deprecated and will be removed.
+                  </li>
+                  <li>
+                    Some steps require you to do changes on GitHub,
+                    refresh the page after making the changes to see updates.
+                  </li>
+                  <li>
+                    New projects imported after connecting your account to the new GitHub App don't need migration,
+                    as they will be automatically connected to the new GitHub App.
+                  </li>
+                  <li>
+                    After revoking the old GitHub OAuth app, you won't be able to auto-migrate projects.
+                    <a href="#">Manual migration</a> is always an option.
+                  </li>
+                  <li>
+                    Projects not linked to a GitHub repository can be linked after connecting your account. These don't need migration.
+                  </li>
+                </ul>
+              </p>
+              <div class="ui divider"></div>
+              <a href="?step=connect" class="ui button">Start</a>
+            {% elif step == "connect" %}
+              <p>
+                Connect your account to the new GitHub app by clicking the button below.
+                As this is a new app, you will need to grant access to your account again.
+              </p>
+              <form class="ui form"
+                    method="post"
+                    action="{% provider_login_url "githubapp" process="connect" next=request.get_full_path %}">
+                {% csrf_token %}
+                <button class="ui button"
+                        type="submit"
+                        {% if step_connect_completed %}disabled{% endif %}>
+                  <i class="fa-brands fa-github icon"></i>
+                  {% if step_connect_completed %}
+                    {% trans "Connected" %}
+                  {% else %}
+                    {% trans "Connect" %}
+                  {% endif %}
+                </button>
+              </form>
+
+              <div class="ui divider"></div>
+              <a class="ui button {% if not step_connect_completed %}disabled{% endif %}"
+                 href="?step=install">Next</a>
+            {% elif step == "install" %}
+              <p>
+                Click on &quot;Install&quot; to install the GitHub App on each account,
+                all the repositories connected to a project will be pre-selected.
+              </p>
+
+              <div class="ui middle aligned relaxed divided list">
+                {% for installation_target_group in installation_target_groups %}
+                  <div class="item">
+                    <div class="right floated content">
+                      <a href="{{ installation_target_group.link }}"
+                         target="_blank"
+                         class="ui small button {% if installation_target_group.installed %}disabled{% endif %}">
+                        {% if installation_target_group.installed %}
+                          Installed
+                        {% else %}
+                          Install
+                        {% endif %}
+                      </a>
+                    </div>
+                    <div class="content">
+                      <i class="fa-brands fa-github icon"></i>
+                      <a href="https://github.com/{{ installation_target_group.target_name }}"
+                         target="_blank">
+                        {{ installation_target_group.target_name }}
+                      </a>
+                    </div>
+                  </div>
+                {% empty %}
+                  <div class="item">
+                    <div class="content">
+                      <div class="ui message info">
+                        You don't have projects to migrate, you can <a href="?step=revoke">skip this step</a>.
+                      </div>
+                    </div>
+                  </div>
+                {% endfor %}
+              </div>
+
+              <p>
+                Alternatively, you can <a href="https://github.com/apps/{{ gh_app_name }}/installations/new/"
+    target="_blank">manually install the GitHub App</a>
+                on each repository you want to migrate, or do it one by one on the next step.
+              </p>
+
+              <div class="ui message info">
+                <i class="fad fa-info-circle icon"></i>
+                If you see this step as incomplete after installing the GitHub App,
+                you may need to <a href="{{ request.get_full_path }}">refresh this page</a>, or if you don't want to install the GitHub App on all selected repositories,
+                you can <a href="?step=migrate">skip this step</a>.
+              </div>
+
+              <div class="ui divider"></div>
+
+              <a class="ui button" href="?step=migrate">Next</a>
+            {% elif step == "migrate" %}
+              <p>
+                In order to migrate a project to the new GitHub App,
+                the GitHub App needs to be installed on the repository connected to the project,
+                and you need to have admin access to the repository.
+              </p>
+
+              <div class="ui medium header">Projects to migrate</div>
+
+              <div class="ui middle aligned relaxed divided list">
+                {% for migration_target in migration_targets %}
+                  <div class="item">
+                    <div class="right floated content">
+                      {% if not migration_target.has_installation %}
+                        <a class="ui button small"
+                           href="{{ migration_target.installation_link }}"
+                           target="_blank">Install</a>
+                      {% else %}
+                        <form method="post" action="." class="ui form">
+                          {% csrf_token %}
+                          <input type="hidden"
+                                 name="project"
+                                 value="{{ migration_target.project.slug }}">
+                          <button class="ui button small"
+                                  type="submit"
+                                  {% if not migration_target.is_admin %}disabled{% endif %}>
+                            Migrate
+                          </button>
+                        </form>
+                      {% endif %}
+                    </div>
+                    <div class="right floated content">
+                      {% if not migration_target.has_installation or not migration_target.is_admin %}
+                        <a class="ui orange label"
+                           data-content="{% if not migration_target.has_installation %}Install the GitHub App in the repository by clicking on the &quot;Install&quot; button{% elif not migration_target.is_admin %}You need to have admin access to the repository in order to migrate the project{% endif %}">
+                          <i class="fad fa-info-circle icon"></i>
+                          Project can't be migrated
+                        </a>
+                      {% endif %}
+                    </div>
+                    <div class="content">
+                      <a class="header"
+                         href="{% url 'projects_detail' migration_target.project.slug %}">
+                        {{ migration_target.project.name }}
+                      </a>
+                      <div class="description">
+                        <i class="fa-brands fa-github icon"></i>
+                        <a href="{{ migration_target.project.remote_repository.html_url }}"
+                           target="_blank">
+                          {{ migration_target.project.remote_repository.full_name }}
+                        </a>
+                      </div>
+                    </div>
+                  </div>
+                {% empty %}
+                  <div class="item">
+                    <div class="content">
+                      <div class="ui message info">
+                        You don't have projects to migrate, you can <a href="?step=revoke">skip this step</a>.
+                      </div>
+                    </div>
+                  </div>
+                {% endfor %}
+              </div>
+              {% if migration_targets %}
+                <p>
+                  If you want to migrate all projects at once, click on the button below.
+                  Note that projects with warnings will not be migrated.
+                </p>
+
+                <form class="ui form" method="post" action=".">
+                  {% csrf_token %}
+                  <button class="ui button" type="submit">Migrate all</button>
+                </form>
+
+                <div class="ui message info">
+                  <i class="fad fa-info-circle icon"></i>
+                  If you don't see the &quot;Migrate&quot; button and you already installed the GitHub App on the repository,
+                  and have admin access to it,
+                  you may need to <a href="{{ request.get_full_path }}">refresh this page</a>.
+                </div>
+
+              {% endif %}
+
+              <div class="ui medium header">Projects connected to the GitHub App</div>
+
+              <div class="ui middle aligned relaxed divided list">
+                {% for project in migrated_projects %}
+                  <div class="item">
+                    <div class="content">
+                      <a class="header" href="{% url 'projects_detail' project.slug %}">
+                        {{ project.name }}
+                      </a>
+                      <div class="description">
+                        <i class="fa-brands fa-github icon"></i>
+                        <a href="{{ project.remote_repository.html_url }}" target="_blank">
+                          {{ project.remote_repository.full_name }}
+                        </a>
+                      </div>
+                    </div>
+                  </div>
+                {% empty %}
+                  <div class="item">
+                    <div class="content">
+                      <div class="ui message info">
+                        You don't have projects connected to the GitHub App yet.
+                      </div>
+                    </div>
+                  </div>
+                {% endfor %}
+              </div>
+
+              <div class="ui divider"></div>
+
+              <a class="ui button" href="?step=revoke">Next</a>
+            {% elif step == "revoke" %}
+              <p>
+                Click on the button below to revoke access.
+                You will be redirected to GitHub, where you need to click on "Revoke access".
+              </p>
+
+              {% if migration_targets %}
+                <div class="ui small warning message">
+                  <i class="fad fa-warning icon"></i>
+                  You have projects that need to be <a href="?step=migrate">migrated</a>.
+                  If you revoke access now, you'll need to <a href="#">manually migrate</a> them.
+                </div>
+              {% endif %}
+
+              <a href="{{ old_application_link }}"
+                 target="_blank"
+                 class="ui button {% if step_revoke_completed %}disabled{% endif %}">
+                <i class="fa-brands fa-github icon"></i>
+                {% if step_revoke_completed %}
+                  Revoked
+                {% else %}
+                  Revoke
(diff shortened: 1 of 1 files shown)


readthedocsext/theme/templates/profiles/private/migrate_to_gh_app.html
```diff
@@ -0,0 +1,470 @@
...
+          </div>
+        </div>
+      </div>
+    </div>
+  </div>
+{% endblock content %}
+
+{% block body_scripts %}
+  {{ block.super }}
+  <script type="text/javascript">
```
stsewd, line 438: Not sure how to add a new JS file just for this page, this should be in another file, otherwise it will be blocked by our CSP rules.

stsewd:
We need to upgrade Fomantic to the latest version in order to test the circled layout, but wasn't able to find formantic in the package.json file....

stsewd:
Thinking of manually implementing the checkmark (not sure how), as the builtin doesn't play nice when the description is longer than a few words

Screenshot 2025-03-19 at 14-03-17 Migrate account to GitHub App - Read the Docs

Screenshot 2025-03-19 at 14-05-02 Migrate account to GitHub App - Read the Docs

Or maybe just don't use icos at all...

agjohnson:
This looks great, awesome job!

I just had some minor notes on the visual example so far, I'll take a look at the code next.

----

This listing is going to be really cramped with the warning and button. It will easily wrap (in unexpected ways) if the repo or project name is longer

image

I think we should either show just the warning or just the "Install" button. What you are trying to communicate is that the user skipped a step and didn't install the GHA, but "Project can't be migrated" doesn't explain this either.

Having an "Install" button to effectively go back a step is okay too -- it does seem a little confusing why there are two steps to install our application though. With the steps split up by this, replacing the warning and the "Install" button with a warning "GitHub App not installed". The user would understand _why_ they can't migrate the project with that warning, and they should go back to install the application.

----

Also, we have an element for the repo tag style already:

image

This should be a reusable template:

image

----

Instead of using disabled buttons to indicate successful states, I would use a label with a green check:

image

Or, just "Installed" is fine too. The disabled buttons aren't meant to be readable and they do not pass accessibility contrast standards.

----

Lastly, toast notifications communicating errors within views are too easy to miss. We aren't exposing errors via toast notifications anywhere right now and I think that is best. There are better options to push the user away from entering the wizard twice, like showing an error state on the button that the user would click to enter this wizard.

However, is there a reason we don't want the user able to come back to this UI? I'm sure there will be legitimate use cases where the user can't complete this wizard for all repos -- for example a user can't reach the organization admins in time. I think the best UX for this would be to show the first steps as completed if the user comes back to this UI.

readthedocsext/theme/templates/profiles/private/migrate_to_gh_app.html
agjohnson: This is covered here in the docs:

https://docs.ops.verbthenouns.com/projects/ext-theme/en/latest/javascript.html#javascript

In short:

- Never use scripts within template code. Write JS in JS source files.
- Add this code as a Knockout view or, if you are feeling adventurous, a web component.

agjohnson:
> We need to upgrade Fomantic to the latest version in order to test the circled layout, but wasn't able to find formantic in the package.json file....

This is because we share some styles and code between our website and application, we have a common repo for this:

https://github.com/readthedocs/ext-theme/blob/4251fe1465d79c501b67701c4de0a0b24fcbbe31/package.json#L29

stsewd:
> I think we should either show just the warning or just the "Install" button. What you are trying to communicate is that the user skipped a step and didn't install the GHA, but "Project can't be migrated" doesn't explain this either.

There are two possible reasons a project can't be migrated, one is the app not being installed into the app, and the other one is the user not having admin access to the repo. So we need a way to communicate both errors.

> Having an "Install" button to effectively go back a step is okay too -- it does seem a little confusing why there are two steps to install our application though. With the steps split up by this, replacing the warning and the "Install" button with a warning "GitHub App not installed". The user would understand why they can't migrate the project with that warning, and they should go back to install the application.

The previous install step allows you to install the app in all the repos at once, the install button on the migrate page allows you to install one repo at a time. I thought about other options, but wasn't convinced they were good options:

- In addition to the install into account/org buttons, have a list of repos, so the user can install the projects from there, but that feels like duplicating the page from the migration step.
- Merge both steps, but that requires having the install into account/orgs buttons somewhere, and will probably crowd that page.

> There are better options to push the user away from entering the wizard twice, like showing an error state on the button that the user would click to enter this wizard.

The toast notification is meant for users that go directly to that URL.

> However, is there a reason we don't want the user able to come back to this UI? I'm sure there will be legitimate use cases where the user can't complete this wizard for all repos -- for example a user can't reach the organization admins in time. I think the best UX for this would be to show the first steps as completed if the user comes back to this UI.

Users don't need to migrate all repos at once, they can go back to the wizard any time. The wizard should be accessible until the user completes the last two steps (revoke and disconnect). Once the user has revoked/disconnected access to the old GH OAuth app, there is nothing left to do from the wizard, as we can't migrate the projects anymore.

readthedocsext/theme/templates/profiles/private/migrate_to_gh_app.html
stsewd: This JS is more of a snippet that acts on the whole page, it isn't specific to an element, so not sure if it makes sense as a view or webcomponent.

stsewd:
Something like this? Not sure the way the repo is displayed integrates well with the listing

Screenshot 2025-03-20 at 14-23-47 Migrate account to GitHub App - Read the Docs
Screenshot 2025-03-20 at 14-23-40 Migrate account to GitHub App - Read the Docs
Screenshot 2025-03-20 at 14-23-33 Migrate account to GitHub App - Read the Docs
Screenshot 2025-03-20 at 14-23-27 Migrate account to GitHub App - Read the Docs

readthedocsext/theme/templates/profiles/private/migrate_to_gh_app.html
agjohnson: It is best to not write JS that has wide side effects like that. You should attach the click event handling to specific elements, not intercept every click event on the page -- this will almost certainly have unwanted effects.

agjohnson:
> Something like this?

I would use `<span class="ui green text"><i class="fas fa-circle-check icon"></i>...</span>`, otherwise this looks like a green button.

> There are two possible reasons a project can't be migrated, one is the app not being installed into the app, and the other one is the user not having admin access to the repo.

This is another error state that should be communicated clearer. Do we even want an install button if there is a problem with permissions?

Removing the button and listing the specific reasons why the project can't be migrated would be most useful here. We should avoid cramming multiple elements like this into the listing view as _it will break in a number of different cases_, especially with the side bar menu.

> The toast notification is meant for users that go directly to that URL.

Yeah I picked up on that, but we still should not use toast notifications for this. Users should get a proper view UI where we explain the error state. The users shouldn't be dropped on the dashboard page, as this is unexpected and confusing.

> Once the user has revoked/disconnected access to the old GH OAuth app, there is nothing left to do from the wizard

What I meant was I would expect that if I went through the wizard I'd still at least be able to see which projects weren't migrated yet.

----

I tried to test this locally but it has required a lot of set up so far and I don't quite have it configured -- I think the last piece I need is a development GHA.

stsewd:
> I would use <span class="ui green text"><i class="fas fa-circle-check icon"></i>...</span>, otherwise this looks like a green button.

Is that in all places? like even in listings?

> Do we even want an install button if there is a problem with permissions?

The repo must be installed first in order for us to know if the user has admin permissions on it.

> What I meant was I would expect that if I went through the wizard I'd still at least be able to see which projects weren't migrated yet.

Not sure where would that list fit in the steps, like the steps are designed to follow each one in order, after the old social account has been revoked, it doesn't make sense displaying all the steps, and the migrate step where the repos missing migration are listed wouldn't make sense either as the user can't use the migrate option anymore.

stsewd:
I was able to decouple the old GitHub account from the migration process. So users can still user the migration page after disconnecting the account, or rather, we can still use that same code, since I'm still not sure how to display the steps after the old account has been disconnected. Maybe don't display the revoke/disconnect steps, and change the wording to mention that manual steps may be required for the user to remove the webhook/ssh key from the repo.

stsewd:
Also, what about using an icon to indicate that a project can't be migrated? If the user hovers over that icon we show what's required

Screenshot 2025-03-27 at 14-52-22 Migrate account to GitHub App - Read the Docs

Another idea would be to have four listings:

- Projects ready for migration
- Projects missing installation
- Projects without admin access
- Projects connected to the GitHub App

agjohnson:
I pulled these PRs down to look at them locally and I have a lot of small fixes and UI issues to resolve. I'm finding them hard to summarize in review feedback though, so I'll try opening up some PRs off this branch to help. Some of these PRs will just be guiding points, you'll probably have more work with them still.

agjohnson:
Something happened the first time I tried to connect my account, it gave a 500. I tried revoking the GHA and reconnecting and haven't had it since. I recall you mentioned this was an issue at some point, have you noticed the problem since?

readthedocsext/theme/templates/profiles/private/migrate_to_gh_app.html
```diff
@@ -0,0 +1,442 @@
...
+                  <i class="fa-brands fa-github icon"></i>
+                    {% trans "Connect" %}
+                </button>
+              </form>
+              {% endif %}
+
+              <div class="ui divider"></div>
+              <a class="ui button {% if not step_connect_completed %}disabled{% endif %}"
+                 href="?step=install">{% trans "Next" %}</a>
+            {% elif step == "install" %}
```
agjohnson, line 176: This step seems like it will be confusing as a user. It mentions `repositories will be pre-selected` but doesn't list repositories, it lists GH organizations (and accounts presumably?). I feel like the UI should reference this instead.

stsewd:
> Something happened the first time I tried to connect my account, it gave a 500. I tried revoking the GHA and reconnecting and haven't had it since. I recall you mentioned this was an issue at some point, have you noticed the problem since?

We have experienced that when there are cookies with large values, but most of that problem was from us duplicating the session cookie. I haven't experienced this with the new integration.

readthedocsext/theme/templates/profiles/private/migrate_to_gh_app.html
agjohnson: Sure, but our UI is still unclear by referencing repositories while not showing repositories. We should describe our UI better.

agjohnson:
> We have experienced that when there are cookies with large values, but most of that problem was from us duplicating the session cookie. I haven't experienced this with the new integration.

I'll see if I can replicate this, it seems like it was an intermittent issue but I didn't catch what happened in the logs.

agjohnson:
I put up a number of PRs to address some of the questions above and to use some more standard patterns that were established. I'll have more feedback on copy but there's a lot moving around. Be careful with my PRs, there is a lot of illustrative code in there that you'll need to fill in.

stsewd:
This is how it looks like with the latest changes

https://github.com/user-attachments/assets/26478361-720f-4285-aff2-f2710979054f

readthedocsext/theme/templates/profiles/partials/github_app_target_list.html
```diff
@@ -0,0 +1,82 @@
...
+{% endblock top_right_menu_items %}
+
+{% block list_item_meta_items %}
+{% endblock list_item_meta_items %}
+
+{% block list_item_image %}
+  {# XXXX use user/org avatar here #}
+  {% comment %}
+  <img class="ui rounded image" src="{{ object.target_avatar_url }}" height="28" alt="{% trans "Account avatar" %}" width="28" />
+  {% endcomment %}
```
stsewd, line 59: We don't have access to the account avatar here, that's why I used a github icon, we could maybe use an icon to differentiate between personal accounts and organizations

readthedocsext/theme/templates/profiles/partials/github_app_target_list.html
```diff
@@ -0,0 +1,82 @@
...
+{% endblock list_item_image %}
+
+{% block list_item_header %}
+  {{ object.target_name }}
+  <div class="sub header">
+    {# XXX URL? Something else that is helpful? This is just a placeholder, don't actually do this #}
+    https://github.com/{{ object.target_name }}
+  </div>
```
stsewd, line 67: So, do we want to create a link or not?

humitos:
> This is how it looks like with the latest changes

It's looking great! 💯

stsewd:
Note that this should be tested with https://github.com/readthedocs/readthedocs.org/pull/12112 instead of the old big PR, as I closed that one.

stsewd:
There is the question about how/when to link to this page. I'm thinking we link to this page when the user logs in. I added this

Screenshot 2025-04-16 at 11-46-13 Migrate account to GitHub App - Read the Docs

Maybe also create a notification so users that are already logged in can see (and we can dismiss this notification for users that log in to avoid showing this twice if we want, or keep it as a reminder)

readthedocsext/theme/templates/profiles/partials/github_app_target_list.html
agjohnson: No, I'd definitely use the user or organization avatar from GitHub. If we don't get this URL from the GHA responses it's in the user/org API responses.

readthedocsext/theme/templates/profiles/partials/github_app_target_list.html
agjohnson: I think the repo URL makes sense here. I did not do this because it requires adding a data class field including with the URL from whatever API response is populating the data class. It wasn't immediately clear so I left this as a placeholder that you can fill out.

agjohnson:
> There is the question about how/when to link to this page. I'm thinking we link to this page when the user logs in. I added this

I dislike altering the user's normal workflow with unexpected UX. If the user wants the dashboard, they should always get the dashboard. When I go to GitHub and it interrupts what I was doing to get me to update security information, I'm forced to skip that entire UI so I'm not forcefully sidetracked.

I think we definitely need a notification to get users to the connected services list, but we can add that next iteration.

I pointed out using a GitHub user listing to communicate migration status. I think this is important for the user to understand if there are more steps required still or not. We already have the a list of accounts, the connected services page. That page can have more UI informing the user that changes are required for each account:

image

This gets around the awkward UX of multiple GitHub accounts more easily.

readthedocsext/theme/templates/profiles/partials/github_app_target_list.html
```diff
@@ -0,0 +1,82 @@
...
+{% block list_placeholder_icon_class %}
+  fa-duotone fa-code-commit
+{% endblock list_placeholder_icon_class %}
+{% block list_placeholder_header %}
+  {% blocktrans trimmed %}
+    No installation targets found
+  {% endblocktrans %}
+  <div class="sub header">
+    {% blocktrans trimmed %}
+      You don't have any project connected to a GitHub repository.
```
agjohnson, line 22:
```suggestion
      You don't have any projects connected to GitHub repositories.
```

readthedocsext/theme/templates/profiles/private/migrate_to_gh_app.html
```diff
@@ -0,0 +1,303 @@
...
+                  {% trans "Migrate your projects to the GitHub App" %}
+                </div>
+              </div>
+            </a>
+            <a class="step {% if step == "revoke" %}active{% endif %} {% if not step_connect_completed %}disabled{% endif %}"
+               href="{% url 'migrate_to_github_app' %}?step=revoke">
+              <div class="content">
+                <div class="title">{% trans "Revoke access to old integration" %}</div>
+                <div class="description">
+                  {% trans "Revoke access to the old GitHub OAuth app from your GitHub account" %}
```
agjohnson, line 61:
```suggestion
                  {% trans "Revoke access to our old GitHub OAuth app from your GitHub account" %}
```

readthedocsext/theme/templates/profiles/private/migrate_to_gh_app.html
```diff
@@ -0,0 +1,303 @@
...
+                </div>
+              </div>
+            </a>
+            <a class="step {% if step == "disconnect" %}active{% endif %} {% if step_disconnect_completed %}completed{% endif %} {% if not step_connect_completed or not step_revoke_completed %}disabled{% endif %}"
+               href="{% url 'migrate_to_github_app' %}?step=disconnect">
+              <div class="content">
+                <div class="title">{% trans "Disconnect old integration" %}</div>
+                <div class="description">
+                  {% trans "Disconnect the old GitHub OAuth app from your Read the Docs account" %}
```
agjohnson, line 70:
```suggestion
                  {% trans "Disconnect our old GitHub OAuth app from your Read the Docs account" %}
```

readthedocsext/theme/templates/profiles/private/migrate_to_gh_app.html
```diff
@@ -0,0 +1,303 @@
...
+              </div>
+            </a>
+          </div>
+        </div>
+        <div class="eleven wide computer eleven wide tablet sixteen wide mobile column">
+          <div class="ui basic segment">
+            {% if step == "overview" %}
+              <p>
+                {% blocktrans trimmed with blog_post="#" %}
+                  We’re introducing a new GitHub App to replace the old GitHub OAuth app.
```
agjohnson, line 81:
```suggestion
                  We’re introducing a new GitHub App to replace our old GitHub OAuth app.
```

readthedocsext/theme/templates/profiles/private/migrate_to_gh_app.html
```diff
@@ -0,0 +1,303 @@
...
+                  {% url 'socialaccount_connections' as socialaccount_connections %}
+                  {% blocktrans trimmed with socialaccount_connections=socialaccount_connections github_account=old_github_account.extra_data.login %}
+                    You have <a href="{{ socialaccount_connections }}">multiple GitHub accounts</a> linked,
+                    <strong>you'll need to repeat the migration for each account</strong>.
+
+                    The current account to migrate is <a href="https://github.com/{{ github_account }}" target="_blank">{{ github_account }}</a>;
+                    you can skip repositories this account doesn't have access to.
+                  {% endblocktrans %}
+                </div>
+              {% endif %}
```
agjohnson, line 99: I'd still ditch this, it's a lot of text and this isn't where we should try to get the user to switch accounts to migrate.

The list of accounts to migrate can be the connected services page. This UI also has a better display of which account is being migrated currently on the GHA connection tab in the view.

```suggestion
```

readthedocsext/theme/templates/profiles/private/migrate_to_gh_app.html
```diff
@@ -0,0 +1,303 @@
...
+
+              <p>
+                {% blocktrans %}
```
agjohnson, line 102: You need trimmed on all these blocktrans:

```suggestion
                {% blocktrans trimmed %}
```

readthedocsext/theme/templates/profiles/private/migrate_to_gh_app.html
```diff
@@ -0,0 +1,303 @@
...
+                  <li>
+                    {% blocktrans %}
+                    Connecting your account to the new GitHub App and revoking the old GitHub OAuth app is required,
+                    as the old app is deprecated and will be removed.
+                    {% endblocktrans %}
+                  </li>
+                  <li>
+                    {% blocktrans %}
+                    Some steps require you to do changes on GitHub,
+                    refresh the page after making the changes to see updates.
```
agjohnson, line 125:
```suggestion
                    Some steps require you to make changes on GitHub.
                    Refresh the page after making the changes to see updates.
```

readthedocsext/theme/templates/profiles/private/migrate_to_gh_app.html
```diff
@@ -0,0 +1,303 @@
...
+                    {% endblocktrans %}
+                  </li>
+                  <li>
+                    {% blocktrans %}
+                    New projects imported after connecting your account to the new GitHub App don't need migration,
+                    as they will be automatically connected to the new GitHub App.
```
agjohnson, line 131:
```suggestion
                    New projects created after connecting your account to the new GitHub App don't need migration,
                    as they will be automatically connected to the new GitHub App.
```

readthedocsext/theme/templates/profiles/private/migrate_to_gh_app.html
```diff
@@ -0,0 +1,303 @@
...
+                  <li>
+                    {% blocktrans %}
+                    Migrating all projects is optional,
+                    you can skip the &quot;Install GitHub App&quot; and &quot;Migrate Projects&quot; steps.
+                    {% endblocktrans %}
+                  </li>
+                  <li>
+                    {% blocktrans %}
+                    Connecting your account to the new GitHub App and revoking the old GitHub OAuth app is required,
+                    as the old app is deprecated and will be removed.
```
agjohnson, line 119:
```suggestion
                    Connecting your account to our new GitHub App and revoking access to our old GitHub OAuth app is required,
                    as the old app is deprecated and will be removed.
```

readthedocsext/theme/templates/profiles/private/migrate_to_gh_app.html
```diff
@@ -0,0 +1,303 @@
...
+                  <li>
+                    {% blocktrans %}
+                    New projects imported after connecting your account to the new GitHub App don't need migration,
+                    as they will be automatically connected to the new GitHub App.
+                    {% endblocktrans %}
+                  </li>
+                  <li>
+                    {% blocktrans trimmed with manual_migration_docs="#" %}
+                      After revoking the old GitHub OAuth app, you won't be able to auto-migrate projects.
+                      <a href="{{ manual_migration_docs }}">Manual migration</a> is always an option.
```
agjohnson, line 137: This isn't a great place to push the user towards documentation as we don't want to distract the user from this UI. This note is adding to information overload on this page, and the view is already very text dense -- and the more text the less likely all of it will be read.

This seems more like a note for edge cases. If this is not necessary for initial migration, I think we can cut this and make this proper UI in the project admin.

```suggestion
                      After revoking our old GitHub OAuth app, you won't be able to automatically migrate projects.
                      <a href="{{ manual_migration_docs }}">Manual migration</a> is always an option.
```

readthedocsext/theme/templates/profiles/private/migrate_to_gh_app.html
```diff
@@ -0,0 +1,303 @@
...
+                  </li>
+                  <li>
+                    {% blocktrans %}
+                    Projects not linked to a GitHub repository can be linked after connecting your account. These don't need migration.
+                    {% endblocktrans %}
+                  </li>
+                  <li>
+                    {% blocktrans %}
+                    You can go back to this page page at any time to see the status of your migration,
+                    or to complete the migration later.
```
agjohnson, line 148: This is another block that could be removed. Users should see the status on the connected services page, we don't need to call that out here.

readthedocsext/theme/templates/profiles/private/migrate_to_gh_app.html
```diff
@@ -0,0 +1,303 @@
...
+
+              {% if not step_revoke_completed %}
+              <div class="ui message info">
+                <i class="fad fa-info-circle icon"></i>
+                {% blocktrans trimmed with current_page=request.get_full_path %}
+                  If you see this step as incomplete after revoking access,
+                  you may need to <a href="{{ current_page }}">refresh this page</a>.
+                {% endblocktrans %}
+              </div>
+              {% endif %}
```
agjohnson, line 270: I'd still reuse the same UI from the connection step and show this as a list of accounts with the button to the right -- and reuse that partial or extend from it. This would make the UI and UX match between the steps for connection and revoke and again makes it very explicit what we're about to do for the user.

readthedocsext/theme/templates/profiles/private/migrate_to_gh_app.html
```diff
@@ -0,0 +1,303 @@
...
+                {% endblocktrans %}
+              </div>
+
+              <form class="ui form"
+                    method="post"
+                    action="{% url 'socialaccount_connections' %}">
+                {% csrf_token %}
+                <input type="hidden" name="account" value="{{ old_github_account.id }}" />
+                <button class="ui button" type="submit">{% trans "Finish migration" %}</button>
+              </form>
```
agjohnson, line 296: And same here too, this can use the same UI/UX and reuse/extend the account listing partial to keep all these steps explicit and unified

agjohnson (approved):
I'd say this is good to go :+1:

On the connected services listing, let's follow up with that work. We don't need it until we are looking to get more users on the GHA but we should do more testing ourselves first.

- Opened https://github.com/readthedocs/ext-theme/issues/583

stsewd:
Here we go again...

https://github.com/user-attachments/assets/3110cb90-e81e-4af0-a2d4-e4ecbc8bb982


> I dislike altering the user's normal workflow with unexpected UX. If the user wants the dashboard, they should always get the dashboard. When I go to GitHub and it interrupts what I was doing to get me to update security information, I'm forced to skip that entire UI so I'm not forcefully sidetracked.

We can start with just a notification, but eventually we should be more aggressive asking users to migrate, we don't want to have to maintain the old app forever.

I took some time to test support with multiple accounts... it took some time, but it didn't end being that complicated. The migration steps feel a little weird, since the user needs to switch between accounts while doing the migration... maybe the requirements can be relaxed, but feels like we should still communicate to the user that they need to switch accounts in order to complete some steps. GH also doesn't do a great job when you need to install the app into an account you don't have access, but I guess that's to prevent enumeration, another step that could be misleading is when revoking access, if the user has already revoked access to the old app, GH redirects the user to the list of all OAuth apps where there is also a button called "revoke", but it's "revoke all"...

https://github.com/user-attachments/assets/27b71700-74ae-4ded-85c6-51a95a08267e

stsewd:
didn't see your review, I was playing with multiple accounts :D

agjohnson:
That UX seems great! It looked like you needed to reload the page a few times? We can probably address that, users won't know to refresh or go back a step. But like you said there aren't a lot of users that have multiple accounts, we can iterate on fixes there if it works otherwise. 

The only step that I think we'll lose users is that last step that kicks out to the social account list screen. If the user has another account we should figure out a different option that doesn't navigate the user away from that page after clicking disconnect. Also a next iteration thing if we want.

> We can start with just a notification, but eventually we should be more aggressive asking users to migrate, we don't want to have to maintain the old app forever.

We'll likely need to get aggressive as we go yeah. But I'd rely on notifications and email still, we can get more aggressive with both as we go.

stsewd:
> It looked like you needed to reload the page a few times? 

What I have is two tabs in two containers, in one container I'm signed in with my main GH account, and in the other container I'm signed in with my secondary account. To do some steps, you need to do them from the container where you are signed in with the correct account, otherwise you won't be able to do so, and GH will give you a 404 or redirect you to a listing page

PR #573, opened 2025-03-25
Add popup to support button

stsewd:
The icon is great, but it doesn't tell the user what it does.


agjohnson (commented):
Change looks fine, but this file hasn't been linted yet so is failing. The fixes are minor, I'd ignore on H021 though.

agjohnson (approved):

PR #576, opened 2025-04-10
List multiple GitHub accounts instead of hinting at using the UI repeatedly

agjohnson:
- Use an existing partial so this doesn't recreate UI that already exists
- Don't mention multiple accounts up front
- Instead just show the list of accounts

This needs:

- Per social account check that the GitHub user has a GHA connected account
- Handling at the connect button of individual social accounts

image

This is more clear that this connection is to a GitHub account. It wasn't as clear as just a button "Connect".

### DIFF
--- a/readthedocsext/theme/templates/profiles/partials/github_oauth_list.html
+++ b/readthedocsext/theme/templates/profiles/partials/github_oauth_list.html
@@ -1,0 +1,40 @@
+{% extends "socialaccount/partials/social_account_list.html" %}
+
+{% comment %}
+  This template is used by the GHA migration template to list _just_ GitHub
+  OAuth accounts that need connection to our new GHA.
+{% endcomment %}
+
+{% load i18n %}
+{% load provider_login_url from socialaccount %}
+
+{% block top_menu %}
+{% endblock top_menu %}
+
+{% block list_placeholder %}
+{% endblock list_placeholder %}
+
+{% block list_item_meta_items %}
+{% endblock list_item_meta_items %}
+
+{% block list_item_right_menu %}
+  {# TODO This logic needs to work per-socialaccount #}
+  {# if object.is_connected_with_github_app #}
+  {% if step_connect_completed %}
+    <span class="ui green text">
+      <i class="fas fa-circle-check icon"></i>
+      {% trans "Connected" %}
+    </span>
+  {% else %}
+    {# TODO and this needs a way to specify the social account? #}
+    <form class="ui form"
+          method="post"
+          action="{% provider_login_url "githubapp" process="connect" next=request.get_full_path %}">
+      {% csrf_token %}
+      <button class="ui button" type="submit">
+        <i class="fa-brands fa-github icon"></i>
+          {% trans "Connect" %}
+      </button>
+    </form>
+  {% endif %}
+{% endblock list_item_right_menu %}
--- a/readthedocsext/theme/templates/profiles/private/migrate_to_gh_app.html
+++ b/readthedocsext/theme/templates/profiles/private/migrate_to_gh_app.html
@@ -84,20 +84,6 @@
                 {% endblocktrans %}
               </p>
 
-              {% if has_multiple_github_accounts %}
-                <div class="ui message warning">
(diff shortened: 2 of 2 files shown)


readthedocsext/theme/templates/profiles/partials/github_oauth_list.html
```diff
@@ -0,0 +1,40 @@
...
+{% endblock top_menu %}
+
+{% block list_placeholder %}
+{% endblock list_placeholder %}
+
+{% block list_item_meta_items %}
+{% endblock list_item_meta_items %}
+
+{% block list_item_right_menu %}
+  {# TODO This logic needs to work per-socialaccount #}
```
agjohnson, line 21: You'll need to do something with this and the logic below it.

readthedocsext/theme/templates/profiles/private/migrate_to_gh_app.html
```diff
@@ -149,26 +135,12 @@ <h2 class="ui medium dividing header">
...
-                    action="{% provider_login_url "githubapp" process="connect" next=request.get_full_path %}">
-                {% csrf_token %}
-                <button class="ui button" type="submit">
-                  <i class="fa-brands fa-github icon"></i>
-                    {% trans "Connect" %}
-                </button>
-              </form>
-              {% endif %}
+              
+              {% include "profiles/partials/github_oauth_list.html" with objects=github_oauth_accounts %}
```
agjohnson, line 143: I replaced `has_multiple_github_accounts` in the view context data with:

```python
        context["github_oauth_accounts"] = (
            user.socialaccount_set.filter(provider=GitHubProvider.id)
        )
```

stsewd (commented):
So, all other steps rely on having just one account to migrate at a time, this is since for repos that aren't part of an organization, we don't know the ID of the owner, so we assume the owner is the account being migrated. And I wouldn't bother with supporting users with multiple accounts, as the number of these users is pretty low, like 160 on .org.

agjohnson:
Even if only supporting a single account with this view, I'd still list the account visually like the screenshot above. This makes it clear what is being connected. The UI felt vague about what is being connected together using just a button "Connect" and the only other place this is mentioned is in a big block of text on the overview page -- this will be missed by users.

Listing GitHub oauth social accounts could also give the user the ability to switch between migrating GitHub accounts if we want, we don't need the rest of the UI to support multiple GitHub accounts at once. We'd show multiple GitHub OAuth accounts and make it clear _which ones have switched_ and _which account we're currently migrating_.

This reuses a UI partial we already authored instead of having one-off UI, even if we just pass it a 1 count filtered queryset of social accounts to show the account we're migrating.

> So, all other steps rely on having just one account to migrate at a time, this is since for repos that aren't part of an organization, we don't know the ID of the owner, so we assume the owner is the account being migrated

However, I'm not sure I understand this need. Don't we know the GitHub account that we're listing the repositories from? It seems like we'd be able to use that data here and avoid needlessly pushing the user through this view multiple times.

stsewd:
> Don't we know the GitHub account that we're listing the repositories from? It seems like we'd be able to use that data here and avoid needlessly pushing the user through this view multiple times.

We know if the user has access to a repo, but we don't know the ID of the owner of that repo. For example, user/repo, we don't know if the current user is the owner of that repo, and we don't save the ID of the owner of the repo (we only do this if the repo is owned by an org). And GH requires using the ID, not the login to link users to the installation page... A hack could be to use the login to retrieve the ID by using the SocialAccount model...

stsewd:
But again, not sure if we want to invest too much into supporting multiple accounts jut for a couple of users.

stsewd:
I ported the using the listing for a single account into https://github.com/readthedocs/ext-theme/pull/570/

agjohnson:
After we start linking to this UI from the connected services page, I'll want to see what the UX feels like. Listing multiple GitHub accounts on this view might not be necessary, but it could be an easy way to allow the user to switch between accounts.

If the connected service listing has all of the same information just a single account in the listing with the connect button is probably fine too though.

PR #578, opened 2025-04-11
Improve GHA installation tab

agjohnson:
- Use less blocks of text and more UI elements
- Use common listing partial for list of installation targets instead of one-off listing structure. This matches the styling we have on all other list of objects as well.
- Use placeholder content area in listing instead of extra blocks of text

Note: I couldn't quite test the install button as it gives me a 404.

### With targets

image

### Without targets

image

### DIFF
--- a/readthedocsext/theme/templates/profiles/partials/github_app_target_list.html
+++ b/readthedocsext/theme/templates/profiles/partials/github_app_target_list.html
@@ -1,0 +1,83 @@
+{% extends "includes/crud/table_list.html" %}
+
+{% load trans blocktrans from i18n %}
+
+{% comment rst %}
+  This template is used by the GHA migration template to list GHA
+  installation targets like organization and user accounts.
+
+  :param objects: List of installation target organizations and user accounts
+  :type objects: InstallationTargetGroup
+{% endcomment %}
+
+{% block list_placeholder_icon_class %}
+  fa-duotone fa-code-commit
+{% endblock list_placeholder_icon_class %}
+{% block list_placeholder_header %}
+  {# XXX Why wouldn't the user have organizations or user accounts to install the GHA on here? #}
+  {% blocktrans trimmed %}
+    No installation targets found
+  {% endblocktrans %}
+  <div class="sub header">
+    {% blocktrans trimmed %}
+      You don't have permissions to install our GitHub App on any organizations or user accounts.
+    {% endblocktrans %}
+  </div>
+{% endblock list_placeholder_header %}
+{% block list_placeholder_text %}
+  <a class="ui button"
+     href="?step=revoke"
+     aria-label="{% trans "Skip this step step" %}"
+     target="_blank">{% trans "Skip this step" %}</a>
+{% endblock list_placeholder_text %}
+
+{% block top_right_menu_items %}
+  {% if objects %}
+    <div class="item">
+      <em>
+        {% blocktrans trimmed %}
+          You may need to refresh after installation to continue
+        {% endblocktrans %}
+      </em>
+    </div>
+  {% endif %}
+  <div class="item">
+    <a class="ui small compact primary {% if not objects %}disabled{% endif %} button"
+       href="{{ current_page }}">
+      <i class="fas fa-refresh icon"></i>
+      {% trans "Refresh" %}
+    </a>
+  </div>
+{% endblock top_right_menu_items %}
+
+{% block list_item_meta_items %}
+{% endblock list_item_meta_items %}
+
+{% block list_item_image %}
+  {# XXXX use user/org avatar here #}
+  {% comment %}
+  <img class="ui rounded image" src="{{ object.target_avatar_url }}" height="28" alt="{% trans "Account avatar" %}" width="28" />
+  {% endcomment %}
+{% endblock list_item_image %}
+
+{% block list_item_header %}
+  {{ object.target_name }}
+  <div class="sub header">
+    {# XXX URL? Something else that is helpful? This is just a placeholder, don't actually do this #}
+    https://github.com/{{ object.target_name }}
+  </div>
+{% endblock list_item_header %}
+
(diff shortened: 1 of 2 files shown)


readthedocsext/theme/templates/profiles/partials/github_app_target_list.html
```diff
@@ -0,0 +1,83 @@
...
+
+  :param objects: List of installation target organizations and user accounts
+  :type objects: InstallationTargetGroup
+{% endcomment %}
+
+{% block list_placeholder_icon_class %}
+  fa-duotone fa-code-commit
+{% endblock list_placeholder_icon_class %}
+{% block list_placeholder_header %}
+  {# XXX Why wouldn't the user have organizations or user accounts to install the GHA on here? #}
```
agjohnson, line 17: I wasn't really sure what was most accurate here, but the copy we were using didn't seem clear. What is the scenario where users would hit this UI and have no organizations or user accounts to connect to?

readthedocsext/theme/templates/profiles/partials/github_app_target_list.html
```diff
@@ -0,0 +1,83 @@
...
+  {% endcomment %}
+{% endblock list_item_image %}
+
+{% block list_item_header %}
+  {{ object.target_name }}
+  <div class="sub header">
+    {# XXX URL? Something else that is helpful? This is just a placeholder, don't actually do this #}
+    https://github.com/{{ object.target_name }}
+  </div>
+{% endblock list_item_header %}
```
agjohnson, line 69: These two need some extra fields on the data class but add a lot to the display

readthedocsext/theme/templates/profiles/private/migrate_to_gh_app.html
```diff
@@ -175,69 +175,27 @@ <h2 class="ui medium dividing header">
...
-
-              <p>
-                {% blocktrans trimmed with gh_app_name=gh_app_name %}
-                  Alternatively, you can
-                  <a href="https://github.com/apps/{{ gh_app_name }}/installations/new/" data-bind="click: trackLinkClick"
-                     target="_blank">manually install the GitHub App</a>
-                  on each repository you want to migrate, or do it one by one on the next step.
-                {% endblocktrans %}
-              </p>
+              {% include "profiles/partials/github_app_target_list.html" with objects=installation_target_groups %}
```
agjohnson, line 184: In general, the partial listing templates were meant to solve the problem where we reinvent how to display a list of objects in each UI. Instead, always create a partial or extend from an existing partial. This avoids making incompatible styles and repeating mistakes in display that we already might have dealt with.

readthedocsext/theme/templates/profiles/private/migrate_to_gh_app.html
```diff
@@ -175,69 +175,27 @@ <h2 class="ui medium dividing header">
...
+                  {% blocktrans trimmed %}
+                    If you skip installation during migration, you will need to manually migrate your project to use an account connected through our GitHub App later.
+                  {% endblocktrans %}
+                </div>
+              {% endif %}
 
               <div class="ui divider"></div>
 
               <a class="ui button" href="?step=migrate">{% trans "Next" %}</a>
+              <a class="ui basic button" href="?step=migrate">{% trans "Skip installation and continue" %}</a>
```
agjohnson, line 198: I'd suggest using UI elements over blocks of text as much as possible. Text is disregarded by users and most users only read a fraction of what is on the page. This was moved from a block of text and a text link and instead grouped with the other button that proceeds the steps, providing context this also proceeds the steps.

stsewd:
> Note: I couldn't quite test the install button as it gives me a 404.

What's the URL? You are probably missing defining the `RTD_GITHUB_APP_NAME` env var to the name of your app.

readthedocsext/theme/templates/profiles/partials/github_app_target_list.html
stsewd: This happens when the user doesn't have repositories linked to any project, this isn't related to permissions.

readthedocsext/theme/templates/profiles/partials/github_app_target_list.html
```diff
@@ -0,0 +1,83 @@
...
+  <div class="sub header">
+    {% blocktrans trimmed %}
+      You don't have permissions to install our GitHub App on any organizations or user accounts.
+    {% endblocktrans %}
+  </div>
+{% endblock list_placeholder_header %}
+{% block list_placeholder_text %}
+  <a class="ui button"
+     href="?step=revoke"
+     aria-label="{% trans "Skip this step step" %}"
```
stsewd, line 30:
```suggestion
     aria-label="{% trans "Skip this step" %}"
```

readthedocsext/theme/templates/profiles/private/migrate_to_gh_app.html
```diff
@@ -175,69 +175,27 @@ <h2 class="ui medium dividing header">
...
-                  If you see this step as incomplete after installing the GitHub App,
-                  you may need to <a href="{{ current_page }}">refresh this page</a>, or if you don't want to install the GitHub App on all selected repositories,
-                  you can <a href="{{ migrate_step }}">skip this step</a>.
-                {% endblocktrans %}
-              </div>
+              {% if installation_target_groups %}
+                <div class="ui message info">
+                  <i class="fad fa-info-circle icon"></i>
+                  {% blocktrans trimmed %}
+                    If you skip installation during migration, you will need to manually migrate your project to use an account connected through our GitHub App later.
```
stsewd, line 190: > ...to use an account connected through our GitHub App later.
Comment

Not sure about this part, maybe just omit it? Like, we don't really use accounts anymore to interact with the repo. Or maybe something like "...to keep the integration working".

stsewd:
Gonna merge all these and apply some fixes in the main PR.

PR #579, opened 2025-04-11
Clean up migration project listing

agjohnson:
- Use partials instead of one off listings
- Reduce text and move to UI elements
- Compact information in table listing and move warnings about installation/migration issues to the button. Show button as disabled when next steps are not possible.
- Use a partial for the second list at the bottom

Relevant commit is be49f17

image

This addresses issues with the UI being quite cramped and elements overlapping/wrapping and the UI generally having a bit too much text.

I did not test if pagination would work, I assume not. I don't love having two lists on this UI, it would be better as a singular list instead of two. This would require reworking the logic used, namely the use of dataclasses for each list of objects.

### DIFF
--- a/readthedocsext/theme/templates/profiles/partials/github_app_project_list.html
+++ b/readthedocsext/theme/templates/profiles/partials/github_app_project_list.html
@@ -1,0 +1,102 @@
+{% extends "includes/crud/table_list.html" %}
+
+{% load trans blocktrans from i18n %}
+
+{% comment rst %}
+  This template is used by the GHA migration template to list projects to migrate
+{% endcomment %}
+
+{% block list_placeholder_icon_class %}
+  fa-duotone fa-octagon-xmark
+{% endblock list_placeholder_icon_class %}
+{% block list_placeholder_header %}
+  {% blocktrans trimmed %}
+    No projects can be migrated
+  {% endblocktrans %}
+  <div class="sub header">
+    {% blocktrans trimmed %}
+      You don't have any projects to migrate, or you don't have the permissions necessary to migrate your projects to use our GitHub App.
+    {% endblocktrans %}
+  </div>
+{% endblock list_placeholder_header %}
+{% block list_placeholder_text %}
+  <a class="ui button"
+     href="?step=revoke"
+     aria-label="{% trans "Skip this step step" %}"
+     target="_blank">{% trans "Skip this step" %}</a>
+{% endblock list_placeholder_text %}
+
+{% block top_right_menu_items %}
+  {% if objects %}
+    <div class="item">
+      <em>
+        {% blocktrans trimmed %}
+          You may need to manually refresh after installation or migration
+        {% endblocktrans %}
+      </em>
+    </div>
+  {% endif %}
+  <div class="item">
+    <a class="ui small compact primary {% if not objects %}disabled{% endif %} button"
+       href="{{ current_page }}">
+      <i class="fas fa-refresh icon"></i>
+      {% trans "Refresh" %}
+    </a>
+  </div>
+{% endblock top_right_menu_items %}
+
+{% block list_item_image %}
+{% endblock list_item_image %}
+
+{% block list_item_header %}
+  <a href="{% url 'projects_detail' object.project.slug %}"
+     target="_blank">{{ object.project.name }}</a>
+{% endblock list_item_header %}
+
+{% block list_item_right_menu %}
+  {# XXX is this the correct state check required here? #}
+  {% if not object.has_installation and not object.is_admin %}
+    <div data-content="{% trans "You don't have the permissions to install our GitHub App" %}">
+      <button class="ui small disabled negative button">
+        <i class="fas fa-triangle-exclamation icon"></i> {% trans "Install" %}
+      </button>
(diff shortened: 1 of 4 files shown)


readthedocsext/theme/templates/profiles/partials/github_app_project_list.html
```diff
@@ -0,0 +1,102 @@
...
+{% block list_placeholder_icon_class %}
+  fa-duotone fa-octagon-xmark
+{% endblock list_placeholder_icon_class %}
+{% block list_placeholder_header %}
+  {% blocktrans trimmed %}
+    No projects can be migrated
+  {% endblocktrans %}
+  <div class="sub header">
+    {% blocktrans trimmed %}
+      You don't have any projects to migrate, or you don't have the permissions necessary to migrate your projects to use our GitHub App.
```
agjohnson, line 18: Not really sure if this is correct or not.

readthedocsext/theme/templates/profiles/partials/github_app_project_list.html
```diff
@@ -0,0 +1,102 @@
...
+{% block list_item_image %}
+{% endblock list_item_image %}
+
+{% block list_item_header %}
+  <a href="{% url 'projects_detail' object.project.slug %}"
+     target="_blank">{{ object.project.name }}</a>
+{% endblock list_item_header %}
+
+{% block list_item_right_menu %}
+  {# XXX is this the correct state check required here? #}
```
agjohnson, line 57: I think you're describing blocking the user here on both _inability to install the app_ and _inability to migrate to the app_. These can be individual warnings on these buttons.

readthedocsext/theme/templates/profiles/partials/github_app_project_list.html
```diff
@@ -0,0 +1,102 @@
...
+       target="_blank">{% trans "Install" %}</a>
+  {% elif not object.can_be_migrated %}
+    {# djlint: off H025 #}
+    {% if not object.is_admin %}
+ 
+      <div data-content="{% trans "Your GitHub user must have admin access to the repository in order to migrate this project" %}">
+      {% else %}
+        {# XXX Why? #} <div data-content="{% trans "This project can't be migrated." %}">
+      {% endif %}
+      {# djlint: on #}
```
agjohnson, line 77: This is due to the multiple opened elements. This should be fixed before proceeding, I'll leave this to you though. There are examples elsewhere where we've allowed this pattern for Djlint.

readthedocsext/theme/templates/profiles/partials/github_app_project_migrated_list.html
```diff
@@ -0,0 +1,35 @@
...
+
+{% block list_item_image %}
+{% endblock list_item_image %}
+
+{% block list_item_right_menu %}
+  <div class="ui green text">
+    <i class="fas fa-circle-check icon"></i>
+    {% trans "Migrated" %}
+  </div>
+{% endblock list_item_right_menu %}
```
agjohnson, line 35: My local dev is not working enough to test installation or migration, so this is untested.

readthedocsext/theme/templates/profiles/private/migrate_to_gh_app.html
```diff
@@ -175,194 +175,60 @@ <h2 class="ui medium dividing header">
...
-                and you need to have admin access to the repository.
+                {% blocktrans trimmed %}
+                  After installing our GitHub App on each of your repositories,
+                  you will need to migrate your Read the Docs projects to use
+                  the new connection.
                 {% endblocktrans %}
               </p>
 
-              <div class="ui medium header">{% trans "Projects to migrate" %}</div>
+              {% include "profiles/partials/github_app_project_list.html" with objects=migration_targets %}
```
agjohnson, line 208: Same note here on using partials instead of trying to manually create listings each time we need them. This will likely trigger pagination and might need the `use_pagination=False` attribute or whatever it is. This is not a great UX though, paginating would be best but likely requires models and can't be used on data classes unfortunately.

readthedocsext/theme/templates/profiles/private/migrate_to_gh_app.html
```diff
@@ -175,194 +175,60 @@ <h2 class="ui medium dividing header">
...
-                  <div class="item">
-                    <div class="content">
-                      <div class="ui message info">
-                        {% trans "You don't have projects connected to the GitHub App yet." %}
-                      </div>
-                    </div>
-                  </div>
-                {% endfor %}
-              </div>
+              {% include "profiles/partials/github_app_project_migrated_list.html" with objects=migrated_projects %}
```
agjohnson, line 231: I'm not sure in love with having multiple lists. It's awkward UX when projects from the top list shift to the bottom list -- like in the old version active/inactive UI. This feels like it would be best as a single table but the use of dataclasses likely is going to prevent this I'm guessing.

readthedocsext/theme/templates/profiles/partials/github_app_project_list.html
stsewd: Projects that can't be migrated for any problem would still be listed.

PR #581, opened 2025-04-15
Rename admin dashboard "Git integrations"

agjohnson:
We renamed this to "Git integrations" for some reason, but:

- This feels misleading because these are all webhooks, there is no integration with `git` itself.
- The subpages still all have a heading `Integrations`


stsewd (approved):

humitos:
> We renamed this to "Git integrations" for some reason, but:

I think we renamed it when we deprecated other VCS. I think it was VCS Integrations and we changes to Git Integrations when we deprecated the other ones; but 🤷🏼

PR #584, opened 2025-04-22
Import page: don't allow import of private projects if not allowed

stsewd:


agjohnson:
I'm not sure of the context here, can you provide a description for this change?

agjohnson (approved):
I can't verify what this is supposed to be protecting, but technically the change seems fine minus the lint failure.

stsewd:
> I'm not sure of the context here, can you provide a description for this change?

With the new GH app, all repositories the user grants access to are shown even if they are private, which we shouldn't allow to import (they won't work anyway since we don't generate a clone token for those).

agjohnson:
The lint fixes for inline styles should not us inline styles and should instead have used this pattern for initial hiding of content:

https://github.com/readthedocs/ext-theme/blob/4ad0608f3e281df685caac4664d19e857fd44664/readthedocsext/theme/templates/builds/build_detail.html#L56-L57

PR #587, opened 2025-04-30
Link to docs when project is private

stsewd:
The other page where we talk about private repos is https://docs.readthedocs.com/platform/stable/guides/importing-private-repositories.html, but feels like the commercial page is better suited here.


agjohnson (approved):
Looks good, it's better than what is there. We could probably use a page or a modal explaining _exactly_ what is happening at some point though.

PR #595, opened 2025-05-19
Block integrations page when using GitHub App

stsewd:
Needs https://github.com/readthedocs/readthedocs.org/pull/12193

This only blocks the add integration option, so users can still delete existing integrations manually.

Screenshot 2025-05-19 at 15-53-34 test-builds-stsewd - Integrations - Read the Docs

ref https://github.com/readthedocs/readthedocs.org/issues/12130

### DIFF
--- a/readthedocsext/theme/templates/projects/integration_list.html
+++ b/readthedocsext/theme/templates/projects/integration_list.html
@@ -1,15 +1,35 @@
 {% extends "projects/project_edit_base.html" %}
 
-{% load i18n %}
+{% load trans blocktrans i18n %}
 
-{% block title %}{{ project.name }} - {% trans "Integrations" %}{% endblock %}
+{% block title %}
+  {{ project.name }} - {% trans "Integrations" %}
+{% endblock title %}
 
-{% block project_integrations_active %}active{% endblock %}
-{% block project_edit_content_header %}{% trans "Integrations" %}{% endblock %}
+{% block project_integrations_active %}
+  active
+{% endblock project_integrations_active %}
+{% block project_edit_content_header %}
+  {% trans "Integrations" %}
+{% endblock project_edit_content_header %}
 
 {% block project_edit_content %}
+  {% if project.is_github_app_project %}
+    <div class="ui message icon">
+      <i class="fa-duotone fa-info-circle icon"></i>
+      <div class="content">
+        <div class="header">{% trans "No further configuration needed" %}</div>
+        <p>
+          {% blocktrans trimmed %}
+            This project is already configured to use the <a href="https://readthedocs-landing--12114.org.readthedocs.build/platform/12114/reference/git-integration.html#github-app">GitHub App integration</a>.
+            You don't need to configure any other integrations.
+          {% endblocktrans %}
+        </p>
+      </div>
+    </div>
+  {% endif %}
   {% include "projects/partials/edit/integration_list.html" with objects=object_list %}
-{% endblock %}
+{% endblock project_edit_content %}
 
 {% block project_edit_sidebar_help_topics %}
   {% include "includes/elements/link.html" with url="https://docs.readthedocs.io/page/guides/connecting-git-account.html" text="How to connect your Read the Docs account to your Git provider" is_external=True class="item" %}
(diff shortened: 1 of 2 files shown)


agjohnson:
Why block all integrations from being created. I get GitHub is in conflict, and Bitbucket/GitLab aren't useful for GHA projects, but generic API integration would still be a valid integration.

Feels like what we want here is instead:

- The integration form should reject GitHub integrations when the project is using GHA with a helpful error.
- Optional, our listing should emphasize previously connected GitHub integrations in an error state.

agjohnson:
Also, if there is a problem with the GHA integration where are we going to point users towards to debug? For example right now we can point them at this integration UI and tell them to look at the webhook request data for the integration. What will we do for GHA projects?

stsewd:
> generic API integration would still be a valid integration.

What would be the use case for that? Triggering builds or changing the default branch should be managed by the GH app, we don't want two services doing the same/competing with each other. If users want to trigger builds automatically, they are better off using the API instead.

> if there is a problem with the GHA integration where are we going to point users towards to debug?

In theory, users will have a problem only if their repo isn't connected to a remote repository (bc the user manually unlinked the project, or revoked access to that repo for the GH app).

agjohnson:
> What would be the use case for that?

I don't think we need a use case to leave this UI as is. This is more about making our UX conditional, we should avoid that unless there is a strong reason to block integrations from creation.

> Why block all integrations from being created?

This is my main question. I was more looking for a use case for _why we need to disable integrations for GHA projects_. Is allowing creation causing any problems for projects? It doesn't sound like we have a blocking reason to make this UI conditional. I'm looking to avoid this as much as possible, especially if there is something more helpful to show the user.

> In theory, users will have a problem only if their repo isn't connected to a remote repository (bc the user manually unlinked the project, or revoked access to that repo for the GH app).

This is the sort of information we could be showing the user for a GHA integration in the listing. We don't need to have an edit view for GHA integration, but it feels more confusing to remove this concept for GHA integrations.

stsewd:
> This is more about making our UX conditional, we should avoid that unless there is a strong reason to block integrations from creation.

We don't want users creating incoming webhooks when there is no need for it, and also avoid having two competing webhooks, which could lead to duplicate builds.

> This side effect will make project with Generic API webhooks to be a in a weird state after they migrate to GHA since they won't be able to edit/modify/delete/create these integrations anymore.

No, my suggestion is avoiding the creation of new integrations. Editing and removing existing integrations is still allowed.

agjohnson:
> We don't want users creating incoming webhooks when there is no need for it, and also avoid having two competing webhooks, which could lead to duplicate builds.

But we already allow projects to create multiple integrations. I'm not seeing a true disadvantage to special casing GHA projects here.

I'd prefer this UI/UX is consistent across providers instead. If that means projects can create redundant integrations, it feels fine to leave that up to maintainers. I don't think this is a strong enough problem to warrant making this UI complex and only for GHA projects.

We can avoid redundant integrations by:

- Showing the GHA integration in the listing so it's apparent that _something_ is connecting the project to the repository.
- Disallow GitHub integrations from being created if the project is a GHA project?

I still don't know about the UX around what we are doing for projects when the link between a GHA repository and our projects breaks. Users familiar with integrations might want to recreate something, and might even try recreating the GitHub integration. We'll have an opportunity to give the user a helpful message. Best case we can detect and show the state of the GHA "integration" in our listing though.

> In general, I'm 👍🏼 on deprecating Generic API webhooks. These project should use the APIv3 as Santos mentioned.

Same, though my point wasn't about generic API integrations. For my point you can use any of the other integrations.

What I mean is that with any of these options we/users gain little from us dictating which integrations they can create. I say we allow creation so we aren't making a UI that is sometimes useful and sometimes useless.

ericholscher:
What are the next steps here? Seems like this is also needed or useful for the GHA?

agjohnson:
My opinion here hasn't changed much, I don't think we should make our UIs conditional like this. There isn't a reason described above the makes me want to limit this UI for users.

I'd still like to see GHA reflected in this list even if just mocked. That seems more actionable than turning the UI off for users.

PR #599, opened 2025-05-23
Have an explicit title on Community/Business

ericholscher:


agjohnson:
Title change is fine, but I'm still -1 on favicon and logo color differentiation. It's mainly a change for core team. Normal users aren't using both platforms and they won't understand what different colors means. One set of icons and logos keeps branding understandable.

agjohnson:
The login and search result example is exactly what I've been describing here though, and we've had recent discussions on this even:

- https://github.com/readthedocs/website/issues/354

Focus on branding colors and icons is not quite identifying what the user needs. Users don't understand what a red vs black icon means in search results or links, _we've had differing icons before and it didn't help the issue then_.

The problem is lower level than visual distinction. _Users don't understand that there are differing Business and Community platforms_, or what that even means. We can't communicate that the user might want to switch platforms through branding and icons though, our UI needs more content and needs to adapt to these cases.

I've described in discussions and the issue above that our login view and/or website should adapt to issues like search engine referrals and show similar UI to point the user to the correct site:

image

Forcing users through this UI has provided the strongest reduction in bad logins that we've seen. This is the level of fix I think we should be focused on, including evaluating what we even need indexed at search engines to begin with.

I've been for greatly simplifying our branding because this is already fairly chaotic and we'd benefit from a singular brand.

ericholscher:
@agjohnson I agree those are all nice to have things, but in general we're going to keep hitting instances where the difference between the sites needs to be visually shown in app. We just hit this yesterday with "We shouldn't have two things that say GitHub but mean different things" discussion around the app. I feel like the same arguments apply here.

I think we definitely need _something_ on the site that says "Community" and "Business". I don't care if it's a logo addition, or something in the header/footer, but I think we need to be explaining to people the difference between those sites _in the applications_, and not just on a screen before they land on the site. If the sites themselves don't brand themselves in that way, users will continue to be forever confused, since there isn't any distinction between them when using them. People will always end up where we don't want them to, and we need to build affordances into the design to accomodate that.

agjohnson:
My point on this has been that branding distinction doesn't actually solve users being confused about which site they are on. It serves a purpose, but it's not a solution to this problem.

The user _has to know that there are two sites and that the two sites function differently_ for this to matter. Most users and certainly most average users on business simply don't have this knowledge.

From our bias we can say users should see "Read the Docs Community" and know not to log in, but the users getting tripped up here will not have the knowledge to even pick up that this is significant in any way.

_Relying on branding changes alone doesn't communicate that there is another site and that these users might be about to use the wrong site_. 

> I think we need to be explaining to people the difference between those sites in the applications, and not just on a screen before they land on the site

Agreed we should be doing more in the application to help problems like this.

But I'm saying we need more _active UI and UX_, like instead of relying on passive clues like branding changes we give the user active UI telling them that there are two sites and they might be about to use the wrong one.

_Passive UI is no match against a user that doesn't read or doesn't have the right knowledge_.

> we definitely need something on the site that says "Community" and "Business"

Adding this text in the UI has seemed like an okay addition. But this isn't new, I was pretty explicit above saying I'm against altering our logo and favicon specifically.

I want to make sure we're not complicating our branding and styling again for small gains.

So, adding this type of text has not been a priority because of all of the above. This change would share similar complications as editing our logos and it will require more space horizontally and vertically in already limited spaces in the UI. So, this serves a purpose but it's not a slam dunk easy fix to a problem.

agjohnson:
But there haven't been other attempts. We've been fixated on changing the logos and colors and it keeps coming up as the only proposal. This is what I responded to, again.

And I have given multiple alternatives for how to actively guide users to the correct login. These suggestions don't include changes for visual distinction because these changes are a solution for after users have logged into the wrong dashboard.

We gain more from consistently pointing users to the correct application, well before they are on the wrong dashboard. Almost all our users should only ever see a single dashboard. And if almost all users never see both dashboards, visual distinction really ends up being a side step.

But the goalposts on this keep moving. What I've been against is varied branding. Visual distinction is not problematic, but it is low impact work especially with stronger solutions in place.

PR #600, opened 2025-05-26
Override allauth template to style "Set password"

### LINKED ISSUE #591: Set password form is unstyled
When a user doesn't have a password set (signed up with GitHub for example), and tries to set a password, that view is unstyled.

Image

humitos:
Screenshot_2025-05-26_11-34-47

Closes #591


agjohnson (approved):
Looks good, I didn't know about the set password view at all.

PR #603, opened 2025-05-28
GitHub App: open beta

stsewd:
The revoke and disconnect options are not available during the beta, so users can use the old button to log in without any problems. They can still do this from the social accounts list, but if they do so, next time they log in will be redirected to the correct login page, so.

Ref https://github.com/readthedocs/meta/issues/187

### DIFF
--- a/readthedocsext/theme/templates/profiles/partials/github_oauth_disconnect_list.html
+++ b/readthedocsext/theme/templates/profiles/partials/github_oauth_disconnect_list.html
@@ -23,7 +23,7 @@
         action="{% url 'socialaccount_connections' %}">
     {% csrf_token %}
     <input type="hidden" name="account" value="{{ object.id }}" />
-    <button class="ui button" type="submit" disabled>
+    <button class="ui disabled button" type="submit">
       <i class="fa-brands fa-github icon"></i>
       {% trans "Disconnect" %}
     </button>
--- a/readthedocsext/theme/templates/profiles/partials/github_oauth_disconnect_list.html
+++ b/readthedocsext/theme/templates/profiles/partials/github_oauth_disconnect_list.html
@@ -23,7 +23,7 @@
         action="{% url 'socialaccount_connections' %}">
     {% csrf_token %}
     <input type="hidden" name="account" value="{{ object.id }}" />
-    <button class="ui button" type="submit">
+    <button class="ui button" type="submit" disabled>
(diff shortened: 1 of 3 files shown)


readthedocsext/theme/templates/profiles/partials/github_oauth_disconnect_list.html
```diff
@@ -23,7 +23,7 @@
         action="{% url 'socialaccount_connections' %}">
     {% csrf_token %}
     <input type="hidden" name="account" value="{{ object.id }}" />
-    <button class="ui button" type="submit">
+    <button class="ui button" type="submit" disabled>
```
agjohnson, line 26: I don't think there is difference on styles, but I'd still do this in case:

```suggestion
    <button class="ui disabled button" type="submit" disabled>
```

readthedocsext/theme/templates/profiles/partials/github_oauth_revoke_list.html
```diff
@@ -43,7 +43,7 @@
   {% else %}
     <a href="{{ old_application_link }}"
        target="_blank"
-       class="ui button"
+       class="ui button disabled"
```
agjohnson, line 46: Nitpick, but classes are easier to maintain when kept in semantic order. In some cases explicit order is required. If you look at any example in FUI docs, you'll see they all read like sentences in english -- adjectives/modifiers before the noun.

```suggestion
       class="ui disabled button"
```

agjohnson (approved):
Looks good, but what does the UI look like with this change?

stsewd:
Screenshot 2025-05-29 at 12-16-27 Migrate account to GitHub App - Read the Docs Dev
Screenshot 2025-05-29 at 12-16-08 Migrate account to GitHub App - Read the Docs Dev

agjohnson:
Ahh okay, so yeah there's more conflicting content on that page that we should disable. The UX will be confusing with just the buttons disabled and there is a lot of text on the page that will make the warning hard to notice, including conflicting warnings.

I'd wrap everything below the "We don't recommend" warning with a `.ui.disabled.basic.segment` div. This will show all the UI as disabled, so it gives a clue as to what is in the UI but doesn't allow the user to do anything with it yet.

stsewd:
Like this?
Screenshot 2025-05-29 at 12-32-23 Migrate account to GitHub App - Read the Docs Dev

PR #605, opened 2025-05-29
GitHub App migration page: put migrate all first

stsewd:
If a user has lots of projects, it's hard to see this option without having to scroll all the way down. But looks weird having this big section before the listing, maybe something like having a "migrate all" link to the actual section is better?

Screenshot 2025-05-29 at 12-23-48 Migrate account to GitHub App - Read the Docs Dev

### DIFF
--- a/readthedocsext/theme/templates/profiles/private/migrate_to_gh_app.html
+++ b/readthedocsext/theme/templates/profiles/private/migrate_to_gh_app.html
@@ -193,8 +193,7 @@
                   <h3 class="ui small header">{% trans "Migrate all projects" %}</h3>
                   <p>
                     {% blocktrans trimmed %}
-                      You can also migrate all projects at once if you don't need to manually review migration for each project.
-                      Note that projects with warnings will not be migrated.
+                      Not all projects may be migrated automatically, you might need to take additional steps for some projects.
                     {% endblocktrans %}
                   </p>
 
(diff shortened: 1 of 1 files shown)


agjohnson:
I would avoid putting any more text on this UI. There is a lot distracting the user already.

stsewd:
> That project listed with the "Install" button won't migrate automatically, correct? That is, it doesn't have any warnings.

yep

agjohnson:
Okay yeah, so that message also needs updated to not mention warnings too. Just having an "Install" button isn't really a "project with warnings" to the user.

I'd just describe that generically - "Not all projects may automatically migrate, you might need to manually migrate projects afterwards."

stsewd:
So, are we okay moving the "migrate all" option at the top?

readthedocsext/theme/templates/profiles/private/migrate_to_gh_app.html
```diff
@@ -188,15 +188,12 @@ <h2 class="ui medium dividing header">
...
-              {% include "profiles/partials/github_app_project_list.html" with objects=migration_targets skip_pagination=True %}
-
               {% if migration_targets %}
                 <div class="ui segment">
                   <h3 class="ui small header">{% trans "Migrate all projects" %}</h3>
                   <p>
                     {% blocktrans trimmed %}
-                      You can also migrate all projects at once if you don't need to manually review migration for each project.
-                      Note that projects with warnings will not be migrated.
+                      Not all projects may be migrated automatically, you might need to take additional steps for some projects.
```
humitos, line 196:
```suggestion
                      You might need to take additional steps for some projects that can't be migration automatically.
```

humitos (approved):
Just a copy suggestion.

PR #606, opened 2025-05-29
GitHub App: link to app from import page

stsewd:
Screenshot 2025-05-29 at 16-08-41 Add project - Read the Docs Dev
Screenshot 2025-05-29 at 16-08-35 Add project - Read the Docs Dev

### DIFF
--- a/readthedocsext/theme/templates/projects/partials/project_create_automatic.html
+++ b/readthedocsext/theme/templates/projects/partials/project_create_automatic.html
@@ -1,4 +1,5 @@
 {% load trans blocktrans from i18n %}
+{% load user_has_github_app_account from readthedocs.socialaccounts %}
 
 {% comment rst %}
   Project automatic configuration view
@@ -99,15 +100,26 @@
 
     </div>
     <div class="ui small bottom attached center aligned message">
+      <p>{% trans "Can't find the repository you are searching for?" %}</p>
 
-      {% trans "Can't find the repository you are searching for?" %}
+      <p>
+        {% if user|user_has_github_app_account %}
+          <a class="ui mini black basic compact button"
+             href="https://github.com/apps/{{ GITHUB_APP_NAME }}/installations/new/"
+             target="_blank">
+            <i class="fa-brands fa-github icon"></i>
+            {% trans "Install GitHub App" %}
+          </a>
 
-      <button class="ui mini black basic compact right aligned button"
-              data-bind="click: sync_remote_repos, css: {disabled: is_syncing(), loading: is_syncing()}">
-        <i class="fa-duotone fa-refresh icon"></i>
-        {% trans "Refresh your repositories" %}
-      </button>
+          {% trans "or" %}
+        {% endif %}
 
+        <button class="ui mini black basic compact button"
+                data-bind="click: sync_remote_repos, css: {disabled: is_syncing(), loading: is_syncing()}">
+          <i class="fa-duotone fa-refresh icon"></i>
+          {% trans "Refresh your repositories" %}
+        </button>
+      </p>
     </div>
   </div>
 {% endblock project_add_automatic_placeholder %}


readthedocsext/theme/templates/projects/partials/project_create_automatic.html
```diff
@@ -99,15 +100,26 @@
...
+      <p>{% trans "Can't find the repository you are searching for?" %}</p>
 
-      {% trans "Can't find the repository you are searching for?" %}
+      <p>
+        {% if user|user_has_github_app_account %}
+          <a class="ui mini black basic compact button"
+             href="https://github.com/apps/{{ GITHUB_APP_NAME }}/installations/new/"
+             target="_blank">
+            <i class="fa-brands fa-github icon"></i>
+            {% trans "Install GitHub App" %}
```
humitos, line 111:
```suggestion
            {% trans "Install GitHub App (beta)" %}
```

humitos (commented):
I'm in doubt about this. The button is specific for GitHub while we do support other providers here as well. Is this a temporary workaround while we are in the transition?

What would be the next steps after the transition? Will we remove the "Refresh repositories"? Will we keep both buttons?

stsewd:
The GH option is only displayed for users that are using the GH app.

readthedocsext/theme/templates/projects/partials/project_create_automatic.html
agjohnson: We should be consistent with how we are describing this, we don't really describe this as a beta anywhere else except maybe the migration page that blocks the last few steps.

I'm find labeling it as beta, but we should do it everywhere.

agjohnson: Regarding content, this doesn't communicate that this step might be required every single project creation. If I had already connected/installed the GHA, I would assume I didn't need to click this button ever again.

It should specifically call out that the user needs to update the GHA permissions.

humitos:
> If users will need to update the permissions every project creation, we should make the UI for updating permissions a bit more obvious.

I understand the user will need to install the GHA on every project creation if they haven't selected the option to install GHA in _all the projects_ at once and instead selected specific projects.

One doubt that I have is: will those projects that where our GHA is not installed appear on the listing? If not, that's a problem... Can we list all the projects there (even those where our GHA is not installed) and make the installation of the GHA part of the "Add project" workflow?

humitos:
> It seems like we'll need the GHA permission step to be more mandatory.

Yeah. I see the same here.

The workflow to "Add a project" will be different for users using GHA (without granting access to _all repositories_) because they will **first need to grant** our GHA access to the repository and **then select the repository from the list**. Unfortunately, we have no way to list the repository first because as we don't have access to that repository, we don't even know it exists 😞 

This first step (granting access to the RTD GHA using the external GitHub page) is what I'm saying it has to be more integrated into the the "Add project" workflow. That said, I'm fine if this integration is not perfect yet and it's just some copy with a link to the external GitHub page; but we need to communicate that required first step _clearly_.

Note that we were already confused about this and we are the ones creating this feature. So, I would expect users to be pretty confused here as well 😄 . We need to guide them.

stsewd:
> If the GHA is installed across all repositories, we don't need to do anything special.

We can know this, but we are not saving this information in the model.

> If the GHA is installed on specific repositories, we should instruct the user to update their GHA configuration
> If the GHA is installed on specific repositories but the user can't update the GHA configuration, we should instruct the user to find someone that can update the settings?

Don't think we can know this directly, but GH will prompt the user to "request" approval to grant access to additional projects.

agjohnson:
> We can know this, but we are not saving this information in the model.

That might be a great addition for later. We'll need to see exactly what we give users for an intermediate UI first, and then if it makes more sense to improve the repository selection UI

readthedocsext/theme/templates/projects/partials/project_create_automatic.html
agjohnson: I think this is the only real thing stopping this PR from going out. We can make more improvements in a following UI.

```suggestion
            {% trans "Update GitHub App permissions" %}
```

agjohnson (approved):

PR #609, opened 2025-06-09
Allauth: tune login/connect sorting, naming, and add GitHub App/Oauth modal

### LINKED ISSUE #379: Social accounts: login and sign up account sorting
I found myself wanting a different sorting for Social Account providers that the one we are exposing:

Screenshot_2024-06-12_16-19-43

Ideally, I'd put GitHub and Google at the top and then GitLab and Bitbucket which are less used.

1. GitHub
2. Google
3. GitLab
4. Bitbucket

They are currently sorted by name, but I'd change that sorting in a way we can promote more GitHub and Google: https://github.com/pennersr/django-allauth/blob/38eb3114ab6c1b63359e51407a7349606bb50740/allauth/socialaccount/templatetags/socialaccount.py#L86

Bitbucket in particular, it the first one while it's the least used.

### LINKED ISSUE #588: SAML: user connected services page looks bad with SAML
We aren't displaying an icon and the title text messes up the display:

Image

### LINKED ISSUE #608: Login: add support for GHA at login and connect
The work here started as supporting GHA on our login and connect forms, but it's become more clear that our login/connect views need to support both our GitHub App and GitHub OAuth app for some period of time.

Thinking out loud a little, we have the following use cases to support here:

- New users just signing up. They can probably use GHA
- Old users wanting to switch. They can migrate and will likely be happy on GHA.
- Old users that are unaware. Most users fall in this bucket probably. Some might try to migrate, but these users might also have legitimate cases for reconnecting their OAuth app account too.
- Old users that don't care to migrate. These users will probably wait as long as possible to migrate.

The progression through stages would look something like:

- Open beta started. Show OAuth and GHA buttons at login, preference still on OAuth initially. Show both buttons at account connection list but put preference on GHA.
- Open beta progressing. Hide GitHub OAuth on connect screen from new users.
- Oauth deprecation start. Show OAuth and GHA buttons at login, preference on GHA. Hide OAuth button on social account connection for some users -- ie users that have connected GHA fully, users that haven't used GH OAuth recently?
- OAuth deprecated. Show GHA only on login, GHA only on connect.

This won't block anyone from doing what they need up front in this process and we can ease into next step as we hit milestones in migration.

I started using a modal for this, as the connect list already uses a model for provider selection. Eventually we'd not need a modal as we progress through removing buttons in either login/connect UI.

agjohnson:
- Requires https://github.com/readthedocs/readthedocs.org/pull/12242
- Requires https://github.com/readthedocs/readthedocs-corporate/pull/2007
- Fixes #608 
- Fixes #379 
- Fixes #588 

----

Work left:

- [x] Test with SAML providers
- [x] **On or before release, all database SocialApps need settings in JSON field for `{"hidden": true}` and testing on actual auth.**
  - Instead of relying on `hidden`, SAML providers were filtered from `get_providers`

#### Business with Google provider and GitHub modal button

image

#### Business when no modal is needed

image

#### GitHub modal

image

#### Social account list

image

### DIFF
--- a/readthedocsext/theme/templates/socialaccount/snippets/provider_list.html
+++ b/readthedocsext/theme/templates/socialaccount/snippets/provider_list.html
@@ -75,7 +75,7 @@
                           </i>
                         {% else %}
                           {% blocktrans trimmed %}
-                            For users that connected their GitHub accounts before June 15, 2025.
+                            For users that connected their GitHub accounts before June 24, 2025.
                           {% endblocktrans %}
                         {% endif %}
                       {% elif provider.id == "githubapp" %}
@@ -110,6 +110,12 @@
             </div>
           {% endfor %}
         </div>
+      </div>
+
+      <div class="ui info message">
+        <i class="fas fa-info-circle icon"></i>
+        To learn more about our new GitHub App integration,
+        <a href="https://about.readthedocs.com/blog/2025/06/announcing-our-github-app-beta/">read our beta announcement</a>
       </div>
     </div>
     <div class="actions">


readthedocsext/theme/templates/socialaccount/snippets/provider_list.html
```diff
@@ -1,29 +1,102 @@
...
+{% get_providers process=process as providers %}
+{% get_github_providers process=process as providers_github %}
 
-{% for provider in socialaccount_providers %}
-  {% comment %}
-    - OpenID is not implemented.
-    - SAML is handled in another view, we don't want to list all SAML integrations here.
-    - GitHub App is not exposed to users yet.
-  {% endcomment %}
-  {% if provider.id != 'saml' and provider.id != 'githubapp' %}
```
agjohnson, line 12: This `saml` check is one I need to test, in theory we can use model attributes `hidden=True` just the same as settings.

humitos:
I'm a little confused with this screenshot.

It seems that we are mixing the copy for the "Login / Signup" with the "Connect to GitHub". When I'm connecting my account, it shouldn't say "For users that connected their account before June". That doesn't make sense on that screen, but it does on "Login / Signup", tho.

453181577-81f390f1-f251-43c1-bce3-7f274cef8800

agjohnson:
Good point, we could use an extra step in this. The copy here now is for the case of:

- Existing users reconect with OAuth
- New users go to GHA

But I think what you're looking for is some copy for early beta that covers the case of:

- Most users should use OAuth still
- Only adventurous users should use GHA

There's definitely room for that here.

readthedocsext/theme/templates/socialaccount/partials/social_account_list.html
```diff
@@ -73,12 +81,19 @@
...
-    <span class="ui label">
-      {# get_brand.name|lower here because we can't use `id`, `Bitbucket2Provider.id == 'bitbucket_oauth2'` #}
-      <i class="fa-brands fa-{{ provider.get_brand.name|lower }} icon"></i>
-      {{ provider.get_brand.name }}
-    </span>
+  {% with provider=object.get_provider %}
+    {% if provider.app.pk and provider.id == "saml" %}
+      <span class="ui label">
+        <i class="fas fa-users icon"></i>
+        {% trans "SAML" %}
```
stsewd, line 88: This can be changed in the provider itself https://github.com/readthedocs/readthedocs-corporate/pull/1966/files#diff-82e41779be7f7e930af4ef33aabc9fb3968a5cd9518376a197ff2d624608e46c

stsewd:
I thought we decided to start the beta with one button only, and only add two buttons if there is need for it?

readthedocsext/theme/templates/socialaccount/snippets/provider_list.html
agjohnson: Maybe that's too much then, our `get_providers` can also filter out database providers like SAML instead of relying strictly on `hidden: true`

agjohnson: Example of that here: e403d49

This feels the easiest and a better solution than doing it at the template at least.

agjohnson: I've gone ahead and added that commit here, it seems like a good approach at least for now. We don't have many SAML providers at the moment.

agjohnson:
> This is what I wrote just after the meeting https://github.com/readthedocs/meta/issues/187#issuecomment-2917068587.

Not revoking access to OAuth on migration is still on plan, that's fine.

> so they can still use the old GH button (no need to have two login options during the beta).

But here, we still want users trying to log in with GHA so we can test the whole workflow. We haven't tested this ourselves yet and we also want users testing the GHA buttons so we can confirm if we can reduce to a single button.

Just an OAuth button won't push much farther forward on testing and I think we just want to jump all in on testing.

What is the concern with two buttons at this point?

> Just to be clear, having two buttons doesn't allow users to "go back" to the previous state if they don't like something about the new app or if they encounter an error.

Yeah I think at least initially it's more likely the user tries to connect their GitHub to GHA, hits an issue and can't connect and needs to fall back to OAuth, or stick with an existing OAuth connection.

> I'd be still -1 on showing two buttons when the beta ends

This is definitely the hope, it's in the plan at https://github.com/readthedocs/ext-theme/issues/608

Now I also wouldn't say it's impossible that we have two buttons after beta either though. If we find users are absolutely blocked on migrating to GHA, we could still need this. But we deal with that if we deal with that.

stsewd:
> @stsewd do you have any other technical consideration here? 

Nope

readthedocsext/theme/templates/socialaccount/snippets/provider_list.html
```diff
@@ -1,29 +1,136 @@
...
+                  {% endif %}
+                </div>
+                <div class="description">
+                  <p>
+                    {# As we progress in a beta test period, the connect text should be tuned to make GHA sound like less of a beta feature. #}
+                    {% with is_early_beta=False %}
+                      {% if provider.id == "github" %}
+                        {% if process == "connect" and is_early_beta %}
+                          {% blocktrans trimmed %}
+                            For new users and users reconnecting a previously connected GitHub account.
```
humitos, line 65: I thought we wanted new users to use GH App even on early beta so we can start testing this. If we don't do that, what users are we targeting here?

readthedocsext/theme/templates/socialaccount/snippets/provider_list.html
```diff
@@ -1,29 +1,136 @@
...
+                            For users reconnecting a previously connected GitHub account and users unable to switch to our new app.
+                          {% endblocktrans %}
+                          <i>
+                            {% blocktrans trimmed %}
+                              We recommend new users connect their account using our GitHub App.
+                            {% endblocktrans %}
+                          </i>
+                        {% else %}
+                          {% blocktrans trimmed %}
+                            For users that connected their GitHub accounts before June 15, 2025.
```
humitos, line 78: We need to update this date here.

humitos (approved):
I'm fine moving forward with this so we can deploy it tomorrow. We will need to tweak it as we go, so I'm not super concern as long as it works 😄

agjohnson:
Yeah, I still think we should have the date to give a clue to existing users with connected accounts. The point about not showing a date was more about the connect page, on the login page it still makes sense.

agjohnson:
With the blog post link:

image

Don't pay attention to the other text, I don't think I have the underlying PRs with the tuned priority and hidden settings.

readthedocsext/theme/templates/socialaccount/snippets/provider_list.html
ericholscher: Yea, I think incremental is good here, as long as it's moving forward. 👍

PR #612, opened 2025-06-12
Add links to impersonate under "Debug" modal

humitos:
Screenshot_2025-06-12_16-36-44

### DIFF
--- a/readthedocsext/theme/templates/projects/partials/header.html
+++ b/readthedocsext/theme/templates/projects/partials/header.html
@@ -193,57 +193,59 @@ collapsed, to save visual space and can be expanded with the right label.
       </div>
     </div>
 
-    <div class="ui mini modal" data-modal-id="project-debug">
-      <div class="header">{% trans "Debug information" %}</div>
-      {% comment %}
+    {% if user.is_staff %}
+      <div class="ui mini modal" data-modal-id="project-debug">
+        <div class="header">{% trans "Debug information" %}</div>
+        {% comment %}
         Show spam information on community instance.
         Allows us to quickly understand what's the spam score,
         and also know what rules are matching this project.
-      {% endcomment %}
-      <div class="content">
-        <h3>{% trans "Impersonate" %}</h3>
-        <div class="ui sub header">{% trans "Project maintainer" %}</div>
-        <div class="ui vertical list">
-          {% for user in project|admins %}
-            <div class="item">
-              <a href="{{ ADMIN_URL|cut:"/admin" }}/impersonate/{{ user.pk }}/" target="_blank">{{ user.username }}</a>
-            </div>
-          {% endfor %}
-
-        </div>
-
-        {% if USE_ORGANIZATIONS %}
-          <div class="ui sub header">{% trans "Organization owner" %}</div>
+        {% endcomment %}
+        <div class="content">
+          <h3>{% trans "Impersonate" %}</h3>
+          <div class="ui sub header">{% trans "Project maintainer" %}</div>
           <div class="ui vertical list">
-            {% for owner in organization|admins %}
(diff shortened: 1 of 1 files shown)


stsewd:
Do we want to list every maintainer? In the admin we just expose a link to the first maintainer, which makes sense, I don't think we ever had the need to impersonate a specific user, unless we need to debug an issue with that user, in that case, we can just do that from the admin.

humitos:
I impersonate a specific user all the time. In particular to debug permissions issues when they can't add project or similar situations. Having this list will make this workflow a lot easier and quicker to follow.

If we are worried about listing too many owners/maintainers, we can limit the list to "the last 5 active users".

Also, I just wrapped the modal inside a `is_staff` conditional as suggested.

agjohnson:
> I impersonate a specific user all the time.

Yeah pretty sure we all do this. We weren't saying we don't impersonate single users, we were saying when we debug organization or project issues we don't need to impersonate a specific (or multiple) maintainer/owner users.

For impersonating a single user or debugging a project, I use the impersonate links in Front via a support request.

For an organization, if the Front links aren't enough I impersonate through the links in the admin -- so just the first organization owner is enough.

I basically never use my own session to debug user issues though, so I don't use this menu either. So I don't mind what content goes into it. But generally, I would be hesitant to continue putting additional features in this staff user only UI. I'd rather these links and functionality end up in Front links or the admin so it's usable on business too.

humitos:
I want to re-open this and re-consider. It harms nobody and reduce support time.

I found myself again wanting to impersonate a user in the community site. The support request came via GitHub (so we don't have any of the quick links we have in Front).

While in the project page, I want to click on Debug and then click on the Impersonate link for debugging this. It saves a lot of time.

agjohnson:
> I don't understand why this wouldn't work on both sites?

For projects that are set to private privacy level, I'm assuming staff users don't bypass those permission checks.

But like I said, I also never debug user issues through my session either, so I could be wrong on that. I go into the admin to impersonate the org/user in this case.

Reiterating from above, I don't care what goes into this menu but think @stsewd is correct on not listing every org owner.

readthedocsext/theme/templates/projects/partials/header.html
```diff
@@ -193,35 +193,59 @@
...
-          <div class="ui sub header">{% trans "Spam score" %}</div>
-          <span>{{ project|get_spam_score }}</span>
-          {% if project.spam_rules.exists %}
-            <div class="ui sub header">{% trans "Spam matching rules" %}</div>
+        {% endcomment %}
+        <div class="content">
+          <h3>{% trans "Impersonate" %}</h3>
+          <div class="ui sub header">{% trans "Project maintainer" %}</div>
+          <div class="ui vertical list">
+            {% for user in project|admins %}
```
agjohnson, line 208: To @stsewd point, we really only need one here. If there is a case for more than one, it seems reasonable to slice down to several results. There are projects/orgs with dozens of users.

You could also just link to the admin view of the filtered list of all org owners or project owners instead of duplicating this information in the application.

readthedocsext/theme/templates/projects/partials/header.html
```diff
@@ -193,35 +193,59 @@
...
+              <div class="item">
+                <a href="{{ ADMIN_URL|cut:"/admin" }}/impersonate/{{ user.pk }}/" target="_blank">{{ user.username }}</a>
+              </div>
```
agjohnson, line 211: No need to nest these elements, links can be items too:

```suggestion
              <a class="item" href="{{ ADMIN_URL|cut:"/admin" }}/impersonate/{{ user.pk }}/" target="_blank">{{ user.username }}</a>
```

PR #614, opened 2025-06-17
Organizations: allow uploading avatar

stsewd:
https://github.com/user-attachments/assets/37882662-81cf-4f95-a870-6c96d9085bfc

Looks like the error message from file fields is duplicated, I think we are overriding how we render fields in the new theme somewhere...

Matches https://github.com/readthedocs/readthedocs.org/pull/12254

### DIFF
--- a/readthedocsext/theme/templates/organizations/partials/header.html
+++ b/readthedocsext/theme/templates/organizations/partials/header.html
@@ -29,8 +29,6 @@ collapsed, to save visual space and can be expanded with the right label.
           <div class="eight wide computer eight wide tablet sixteen wide mobile middle aligned column">
             <img class="ui micro rounded right spaced inline image"
                  src="{{ organization.get_avatar_url }}"
-                 height="48"
-                 width="48"
                  alt="{% blocktrans trimmed with organization_name=organization.name %}{{ organization_name }} organization{% endblocktrans %}" />
             <a class="ui medium text" href="{{ organization.get_absolute_url }}">
               {{ organization.name }}
--- a/readthedocsext/theme/templates/organizations/partials/organization_list.html
+++ b/readthedocsext/theme/templates/organizations/partials/organization_list.html
@@ -67,8 +67,6 @@
 {% block list_item_image %}
   <img class="ui image"
        src="{{ object.get_avatar_url }}"
-       height="48"
-       width="48"
        alt="{% blocktrans trimmed with organization_name=object.name %}{{ organization_name }} organization{% endblocktrans %}" />
 {% endblock list_item_image %}
 


readthedocsext/theme/templates/organizations/partials/organization_list.html
```diff
@@ -66,7 +65,11 @@
 {% endblock list_item_right_buttons %}
 
 {% block list_item_image %}
-  <img class="ui image" src="{% gravatar_url object.email 48 %}" />
+  <img class="ui image"
```
agjohnson, line 68: I think you might want the same `micro rounded` classes here:

```suggestion
  <img class="ui micro rounded image"
```

I'd check the social account listing template and maybe try what is there too.

readthedocsext/theme/templates/organizations/partials/organization_list.html
```diff
@@ -66,7 +65,11 @@
...
+       src="{{ object.get_avatar_url }}"
+       height="48"
+       width="48"
```
agjohnson, line 71: Same here on attributes, you might be able to remove this with CSS classes added above

agjohnson (commented):
This is looking pretty close, though the display of the form field looks a bit janky with the text above. We should tune that a bit more.

readthedocsext/theme/templates/organizations/partials/organization_list.html
stsewd: Looks like we aren't using rounded images on projects listings

https://github.com/readthedocs/ext-theme/blob/c91ab2b841b822ef4313b6ee4ccf67f17dee343f/readthedocsext/theme/templates/projects/includes/project_image.html#L12-L12

default_if_none only works if the value is none, in this case is undefined, so the default isn't applied.

readthedocsext/theme/templates/organizations/partials/header.html
```diff
@@ -28,7 +28,9 @@
         {% block organization_header_title %}
           <div class="eight wide computer eight wide tablet sixteen wide mobile middle aligned column">
             <img class="ui micro rounded right spaced inline image"
-                 src="{% gravatar_url organization.email 48 %}"
+                 src="{{ organization.get_avatar_url }}"
+                 height="48"
+                 width="48"
```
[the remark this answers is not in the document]
agjohnson: Gotcha. Even for non-square images, controlling the image size with width/height isn't the best option. This scales both and forces a square image and will cause one dimension to be out of ratio from the source on non-square images.

We should use CSS image fit attributes instead of HTML to control this. Image fit contain/cover would be better, but I think it's fine to start without this.

readthedocsext/theme/templates/organizations/partials/organization_list.html
agjohnson: Yup, project and org listing views are square images, it's only the details/header image that _are_ rounded.

readthedocsext/theme/templates/organizations/partials/header.html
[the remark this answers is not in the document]
stsewd: Do I leave the size attributes in, then?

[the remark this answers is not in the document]
agjohnson: I'd remove the HTML attributes and let the images be non-square for now. The CSS changes should be minor but we can do that separate.

PR #615, opened 2025-06-19
Update build_name to make it a bit more explicit

ericholscher:
This also refactors the logic to have 2 blocks instead of a ton of interspersed if statements,
I find this much easier to read,
though TBH each of these includes is only used once,
so could probably be better off inline in the template instead of as an include.

### DIFF
--- a/readthedocsext/theme/templates/builds/includes/build_name.html
+++ b/readthedocsext/theme/templates/builds/includes/build_name.html
@@ -33,51 +33,47 @@
 
 {% endcomment %}
 
-{% load i18n %}
+{% from i18n load trans blocktrans %}
 
-{# When not a page title, treat the entire breadcrumb as a link to the build #}
-{% if not is_page_title %}<a href="{{ build.get_absolute_url }}">{% endif %}
-<span class="ui {% if is_page_title %}large{% endif %} breadcrumb">
-  <span class="section">
-      {# When used a a page title, link the sections independently #}
-    {% if is_page_title %}
+{% if is_page_title %}
+  <span class="ui large breadcrumb">
+    <span class="section">
       <a href="{% url "projects_detail" build.project.slug %}?slug={{ build.get_version_slug }}">
-    {% endif %}
-
-    {% if build.is_external %}
-      {{ build.external_version_name | lower | capfirst }}
-      {{ build.get_version_name }}
-    {% else %}
-      {# Translators: this renders with the version name, example "Version latest" #}
-      {% blocktrans trimmed with version_name=build.get_version_name %}
-        Version {{ version_name }}
-      {% endblocktrans %}
-    {% endif %}
-
-    {% if is_page_title %}
+        {% if build.is_external %}
+          {{ build.external_version_name | lower | capfirst }}
+          {{ build.get_version_name }}
+        {% else %}
+          {# Translators: this renders with the version name, example "Version latest" #}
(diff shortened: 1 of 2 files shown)


agjohnson:
I've been hoping for improvement to this as well, I don't love what is in the listing/build detail. I noted this in chat, but for context here too, the current display tried to move away from what we had before. This was on the listing and detail showing a prominent "Build #28123123".

The reason for this is because the shared id sequence makes this number to big to be useful and project build ids are not sequential either. This makes it really difficult to scan the build ids or to use them at all -- for example, it takes work to tell they are even in sequential order by id:

image

So I started drifting the build id to be more visually demoted, hoping we could remove it or find another solution. So I'd like to not go backwards here, but I would be :+1: on finding a different solution. There does still need to be difference in the display listing at least.

Better forward options could be:

- Build sequence number per version would be absolute best. This is too difficult of a migration though, maybe there is a way to hack this in the queryset though?
- "Build for {version} at {datetime}". Explicit but I don't love it.
- I think even a 4-5 character hash could be an improvement to a 28M+ sequence id. It would feel just as random and would be short enough. It would obfuscate having a shared build sequence too.

readthedocsext/theme/templates/builds/includes/build_name.html
```diff
@@ -33,51 +33,47 @@
...
+      <span class="active section">
+        <span class="ui grey text">
+          {# Translators: this shows the build number, example "Build #1234" #}
+          {% blocktrans trimmed with build_id=build.pk %}
+            Build #{{ build_id }}
+          {% endblocktrans %}
+        </span>
+      </span>
+    </span>
+  </a>
```
agjohnson, line 78: So yeah as you were saying it would be better to just skip the include. With this change you aren't showing any other breadcrumbs, so we don't want to use that element anymore. And we don't want to show the text as grey anymore either, in line with all of the other listing views. It was only grey to visually demote the id.

So this becomes:

```suggestion
  <a href="{{ build.get_absolute_url }}">
    {# Translators: this shows the build number, example "Build #1234" #}
    {% blocktrans trimmed with build_id=build.pk %}
      Build #{{ build_id }}
    {% endblocktrans %}
  </a>
```

readthedocsext/theme/templates/builds/partials/build_list.html
```diff
@@ -64,7 +64,7 @@
          data-bind="semanticui: { popup: { content: '{{ object.date }}', position: 'top center', delay: { show: 500 }, variation: 'small'}}">
       {# Translators: this will read like "Started 1 month, 3 days ago" #}
       {% blocktrans with object.date|naturaltime as date trimmed %}
-        Started {{ date }}
+        {{ date }}
```
agjohnson, line 67: Oh and not sure on the reason for removal here, but this feels too terse without prefix text. There should be some context around this string instead of just a "5 minutes ago"

ericholscher: It just feels like a weird use of UI space to have the word `Started` like 10x on the same page, but :shrug:

agjohnson: We have spare room on that line on the listing, so I don't think the UI benefits much from being terse. Many other listing UIs have similar verbose descriptions, so it's consistent with those.

GitHub does do this too, which never really stuck out to me there:

image

PR #622, opened 2025-07-02
Show "Last login" label next to the login method

### LINKED ISSUE #421: Login: show "Last used" small icon when showing all the options to login
Basically, it would be good to implement something like this:   https://x.com/dannypostmaa/status/1810864175503339774

Image

humitos:
This is just a demo of what we can do and/or how to use the `last_login_method` template variable. This needs more discussion and work, but opening as a draft to have something to talk over.

### Email

Screenshot_2025-07-02_17-33-48

### GitHub App

Screenshot_2025-07-02_17-34-33

Requires: https://github.com/readthedocs/readthedocs.org/pull/12286
Closes #421


agjohnson:
> Do you refer to something different here?

Yes, specifically "do we need to put a label on the menu"? If we're defaulting is it necessary?

humitos:
OK, this is moving in a good direction 👍🏼 . I only need to attach the buttons to a specific callback to set the cookie, but all the other login to handle the variables in the template are in place:

### VCS - GitHub App

[Peek 2025-07-14 14-48.webm](https://github.com/user-attachments/assets/473ff99b-b317-47ee-ad44-9efdba5024fb)

### VCS - GitLab


Screenshot_2025-07-14_14-51-38

### Email

Screenshot_2025-07-14_14-52-50

### SSO

Screenshot_2025-07-14_14-54-05

readthedocsext/theme/templates/socialaccount/snippets/provider_list_item.html
```diff
@@ -7,14 +7,26 @@
         action="{% provider_login_url provider.id process=process scope=scope auth_params=auth_params %}">
     {% csrf_token %}
 
+    {# djlint: off #}
```
agjohnson, line 10: You should use something less generic than `off`. If there is a rule you're trying to avoid, limit it to that rule.

readthedocsext/theme/templates/socialaccount/snippets/provider_list_item.html
```diff
@@ -7,14 +7,26 @@
...
     <button class="ui {{ button_classes }} button"
+            data-bind="click: $root.save_login_method"
```
agjohnson, line 12: Another `$root`

src/js/application/views.js
```diff
@@ -65,6 +65,16 @@ export class ApplicationView {
...
+
+      // Remove "visible" variation property to remove "Last used" tooltip
+      const buttons = document.querySelectorAll("button.ui");
+      for (const button of buttons) {
+        if (button.dataset.variation) {
+          const variation = button.dataset.variation.split(" ");
+          variation.pop("visible");
+          button.dataset.variation = variation.join(" ");
+        }
+      }
```
agjohnson, line 77: You should always use KO view code to manipulate the DOM. Using JS queries or jQuery to manipulate the DOM in addition is mixing patterns and should be always avoided.

I'm a bit confused what this is trying to do though. What is the intention here? It also seems like this would affect all modals, which we don't want.

src/js/application/views.js
```diff
@@ -94,4 +104,30 @@ export class ApplicationView {
...
+   * @returns {knockout_click}
+   */
+  save_login_method(data, event) {
+    const elem = event.currentTarget;
+    if (window.isSecureContext) {
+      console.debug("Setting last login method: ", elem.dataset.provider);
+      cookieStore.set("last-login-method", elem.dataset.provider);
+    }
+    return true;
+  }
```
agjohnson, line 132: This should be implemented on a dedicated view class, not on the main application view. The main application view is strictly a router to these dedicated views and a couple utility methods.

You should look at any of the other templates that use a model view and work from those templates, they all follow the same pattern with a dedicated view class for each view.

humitos:
@copilot can you help to perform the changes required from the reviewer?

readthedocsext/theme/templates/account/login.html
```diff
@@ -68,44 +69,66 @@
...
+                   name="{{ redirect_field_name }}"
+                   value="{{ redirect_field_value }}" />
+          {% endif %}
+
+          <div class="ui stackable relaxed text menu">
+            <a class="item" href="{% url 'account_reset_password' %}">{% trans "Forgot your password?" %}</a>
+            <div class="right menu">
+              <button class="ui primary button"
+                      data-bind="click: save_login_method"
+                      data-provider="email"
```
agjohnson: This is still working against KO a bit. We should really not need to do Event or Element manipulation to use data with KO. The two patterns that we already have available are:

`data-bind="click: () => { save_login_method('email') }"`

Or better if you need to do something like ensure the click event is handled normally (`true` is returned from the function):

`data-bind="click: save_login_method.bind('email', $data)"`

Both from: https://knockoutjs.com/documentation/click-binding.html

readthedocsext/theme/templates/account/login.html
```diff
@@ -68,44 +69,66 @@
...
+
+    {% if USE_ORGANIZATIONS %}
+      <div class="ui basic center aligned tab segment" data-tab="sso">
+        <a href="{% url "saml_resolve_login" %}{% if redirect_field_value %}?{{ redirect_field_name }}={{ redirect_field_value }}{% endif %}">
+
+          <button class="ui button"
+                  data-bind="click: save_login_method"
+                  data-provider="sso"
+                  type="submit"
+                  {% if last_login_method == "sso" %} data-tooltip="Last used" data-variation="teal visible" data-position="right center" {% endif %}
```
agjohnson: These all need translated:

```suggestion
                  {% if last_login_method == "sso" %} data-tooltip="{% trans "Last used" %}" data-variation="teal visible" data-position="right center" {% endif %}
```

src/js/account/index.js
```diff
@@ -0,0 +1,51 @@
...
+ * LoginView manages login-related functionality including saving the
+ * last used login method to a cookie for session persistence.
+ *
+ * Usage in a binding context:
+ *
+ * .. code:: html
+ *
+ *   <div data-bind="using: LoginView()">
+ *     <form method="post" action="...">
+ *       <button data-bind="click: save_login_method" data-provider="github">
```
agjohnson: Noted above how this should probably use KO patterns instead of mixing element attribute and KO patterns. I don't see where else we're using `data-provider`

src/js/application/views.js
agjohnson: Okay yeah, there are better ways to achieve that.

I think part of what is making this complicated is mixing non-Knockout JS patterns, template code, and SUI HTML-only tooltips together. Instead the solution here would be more consistent if it leaned into using Knockout more and used JS driven tooltips instead of HTML.

agjohnson:
Sorry, I wasn't clear here. I think you were on the right track still, I thought HTML only popups would indeed be okay. But they turned out to be a bit of a liability in UX and the code ended up less consistent once we got to implementing these. I think in the future HTML only popups are still fine for very simple cases though.

PR #623, opened 2025-07-03
Organization: auto populate slug on creation

### LINKED ISSUE #25: Support manual slugs and slug verification on organization creation
The form should:

* Auto slug the name field into an editable slug field
* Throw some errors when slug doesn't happen
* For nicest UX, slug field should have live verification via and API call, but this can also be a form level error to start

Front logo [Front conversations](https://app.frontapp.com/open/top_1xfwsn)

humitos:
Show the `slug` field when creating the organization and auto-populate it based
on the name of the organization.

[Peek 2025-07-03 17-43.webm](https://github.com/user-attachments/assets/1addb2d3-056f-48db-89d0-19dbe6a4510b)

Requires readthedocs/readthedocs.org#12297
Closes #25

### DIFF
--- a/readthedocsext/theme/templates/organizations/organization_form.html
+++ b/readthedocsext/theme/templates/organizations/organization_form.html
@@ -3,6 +3,32 @@
 {% load i18n %}
 {% load crispy_forms_tags %}
 
+{% block head_scripts %}
+{{ block.super }}
+
+<script>
+ // TODO: where this code should live?
+ document.addEventListener("DOMContentLoaded", (event) => {
+     const titleInput = document.querySelector('input[name=name');
+     const slugInput = document.querySelector('input[name=slug]');
+
+     const slugify = (val) => {
+         return val.toString()
+                   .toLowerCase()
+                   .trim()
+                   // replace spaces, non-word chars, underscores and dashes with a single '-'
+                   .replace(/[^\w\s-_]+/g, '-')
+                   .replace(/[-\s]+/g, '-')
+     };
+
+     titleInput.addEventListener('keyup', (e) => {
+         slugInput.setAttribute('value', slugify(titleInput.value));
+     });
+ });
+
+</script>
+{% endblock head_scripts %}
+
 {% block title %}
   {% blocktrans %}Create a new organization{% endblocktrans %}
 {% endblock %}
--- a/readthedocsext/theme/templates/organizations/organization_form.html
+++ b/readthedocsext/theme/templates/organizations/organization_form.html
@@ -16,7 +16,9 @@
          return val.toString()
                    .toLowerCase()
                    .trim()
-                   // replace spaces, non-word chars, underscores and dashes with a single '-'
+                   // Replace spaces, non-word chars, underscores and dashes with a single '-'
+                   // Copied from Django, which is what we are using under the hood
+                   // https://github.com/django/django/blob/1e9db35/django/utils/text.py#L469-L470
                    .replace(/[^\w\s-_]+/g, '-')
                    .replace(/[-\s]+/g, '-')
      };
--- a/readthedocsext/theme/templates/organizations/organization_form.html
+++ b/readthedocsext/theme/templates/organizations/organization_form.html
@@ -1,12 +1,12 @@
 {% extends "organizations/base.html" %}
 
-{% load i18n %}
-{% load crispy_forms_tags %}
+{% load trans blocktrans from i18n %}
+{% load crispy from crispy_forms_tags %}
 
 {% block head_scripts %}
-{{ block.super }}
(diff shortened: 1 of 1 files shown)


src/js/organization/index.js
```diff
@@ -39,3 +39,34 @@ export class OrganizationSettingsAuthorizationView {
...
+          // Copied from Django, which is what we are using under the hood
+          // https://github.com/django/django/blob/1e9db35/django/utils/text.py#L469-L470
+          .replace(/[^\w\s-_]+/g, "-")
+          .replace(/[-\s]+/g, "-")
+      );
+    };
+
+    nameInput.addEventListener("keyup", (e) => {
+      slugInput.setAttribute("value", slugify(nameInput.value));
+    });
```
humitos, line 69: For some reason this stopped working once I moved it inside this JS file. It was working properly when it was inside the HTML 🤷🏼

humitos: I had to use `slugInput.value = ` here.

src/js/organization/index.js
```diff
@@ -39,3 +39,34 @@ export class OrganizationSettingsAuthorizationView {
...
+          .replace(/[-\s]+/g, "-")
+      );
+    };
+
+    nameInput.addEventListener("keyup", (e) => {
+      slugInput.value = slugify(nameInput.value);
+    });
+  }
+}
+Registry.add_view(OrganizationCreateView);
```
agjohnson, line 72: So, this is mixing a few patterns -- Knockout and direct DOM manipulation. I noted this type of pattern here (though it still notes jQuery, DOM manipulation is the important part).

https://docs.ops.verbthenouns.com/projects/ext-theme/en/latest/javascript.html

Knockout is the pattern we use for data binding views, which is what `using: View()` and the view class pattern use. You shouldn't use direct DOM manipulation in this pattern because Knockout gives us all of this already and direct DOM manipulation is redundant to all of this. If the view looks like a Knockout model view, it should be a model view underneath.

To see how to write a Knockout view, the redirects UI is comparable. Here's the template side, there is a JS view too:

https://github.com/readthedocs/ext-theme/blob/eae6e706dea7ce2b78d0e8ebe2a7e45014d77902/readthedocsext/theme/templates/projects/redirect_form.html#L16-L40

It's also valid to write a web component though, which could look something closer to this. Even then I'd rely on web component natives more though. You probably want this route though, not data binding.

Web components are still not widely used yet though, so there's not strong direction I can give. Here are some WCs we do use:

https://github.com/readthedocs/ext-theme/blob/eae6e706dea7ce2b78d0e8ebe2a7e45014d77902/src/js/modules/notifications.js

I'd suggest wrapping the entire form in a web component probably.

agjohnson:
On the UX side of things, I think the user needs to see a live update field with something like:

"Example project default URL: https://org-slug-example-project.readthedocs-hosted.com"

Users won't understand from this that the organization slug is going to affect project URLs. Largely, they won't know what the organization slug does at all, really.

So on the implementation side, exposing the slug field as a visible input is maybe not what we want. We probably want a separate UI with a bit more explanation of the slug. This might also make my suggestion on using a web component more applicable -- instead of altering a form field the web component would render a segment with a verbose explanation of project URLs/etc.

src/js/organization/index.js
agjohnson: Cool, yeah this is totally fine. I'm still just starting to switch KO out where we can, so WC usage still isn't well defined yet or easier to use compared to KO.

agjohnson:
You might have missed push up the code here, I don't see any new changes.

humitos:
Ha, indeed, you are correct. I just pushed them 😄

agjohnson (approved):
Hiding the block when it's empty works, but makes the UI jump around. I think defaulting to a placeholder when the slug is empty is probably better and would always show this block with something useful.

readthedocsext/theme/templates/organizations/organization_form.html
```diff
@@ -87,8 +88,23 @@ <h2 class="ui medium heading">{% trans "Join or create an organization" %}</h2>
...
           <form class="ui form" action="" method="post">
             {% csrf_token %}
+            {% alter_field form.name data_bind="valueInit: name, textInput: name" %}
+            {% alter_field form.slug data_bind="valueInit:slug, text: slug, textInput: slug" %}
             {{ form|crispy }}
 
+            {# djlint: off H021 #}
+            <div class="ui segment"
+                 data-bind="visible: example()"
+                 style="display: none">
```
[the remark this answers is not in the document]
humitos: OK. I'm going to use `organization` and `project` as slugs for the example and I will always show the UI element:


Screenshot_2025-07-08_10-04-50

PR #625, opened 2025-07-07
Build: highlight the name of the version

humitos:
I find pretty hard to know what's the version of the current build on the build details' page, so I want to make it more visible.

I'm following the same style/pattern we are following for the build id and I like it.

Screenshot_2025-07-07_12-50-18

### DIFF
--- a/readthedocsext/theme/templates/builds/includes/build_name.html
+++ b/readthedocsext/theme/templates/builds/includes/build_name.html
@@ -53,14 +53,14 @@
     {% if build.is_external %}
       <span class="section">
         <span class="divider">/</span>
-        <a href="{% url "projects_detail" build.project.slug %}?slug={{ build.get_version_slug }}">
+        {% if is_page_title %}
+          <a href="{% url "projects_detail" build.project.slug %}?slug={{ build.get_version_slug }}">
+          {% endif %}
           {{ build.external_version_name | lower | capfirst }}
           {{ build.get_version_name }}
-        </a>
+          {% if is_page_title %}</a>{% endif %}
       </span>
-    {% endif %}
-
-    {% if not build.is_external %}
+    {% else %}
       <span class="section">
         <span class="divider">/</span>
         {% if is_page_title %}
--- a/readthedocsext/theme/templates/builds/includes/build_name.html
+++ b/readthedocsext/theme/templates/builds/includes/build_name.html
@@ -38,26 +38,34 @@
 {# When not a page title, treat the entire breadcrumb as a link to the build #}
 {% if not is_page_title %}<a href="{{ build.get_absolute_url }}">{% endif %}
 <span class="ui {% if is_page_title %}large{% endif %} breadcrumb">
+    <span class="section">
+        <span class="ui grey text">
+        {# When used a a page title, link the sections independently #}
+        {% if build.is_external %}
+          {{ build.external_version_name | lower | capfirst }}
+          {{ build.get_version_name }}
+        {% else %}
+        {# Translators: this renders with the version name, example "Version latest" #}
+          {% blocktrans trimmed %}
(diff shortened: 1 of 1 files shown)


readthedocsext/theme/templates/builds/includes/build_name.html
```diff
@@ -50,7 +50,9 @@
     {% else %}
       {# Translators: this renders with the version name, example "Version latest" #}
       {% blocktrans trimmed with version_name=build.get_version_name %}
-        Version {{ version_name }}
+        Versions
+        <span class="divider">/</span>
+        <code class="ui grey text">{{ version_name }}</code>
```
agjohnson, line 55: I think this works broken out into a separate breadcrumb, but:

- Don't use `.ui.text` to change link colors, it only affects the inactive link color. If you hover that link it will turn the hover blue that links normally do. I would use the default styling instead of trying to force the section to stand out.
- You should only translate the bare minimum, don't write HTML blocks in `blocktrans`
- `divider` elements should be sibling to `section` elements, not nested. I'm guessing there might be some display issues with this.
- I wouldn't use `code` here, we don't do that for anything else in the breadcrumb

humitos: Cool. I will follow these rules and update the PR. I should probably have marked this PR as Draft and request for early feedback. I wanted to know if we were happy with idea more than the technical side.

agjohnson:
I didn't think about the listing view, I don't think this quite works there. As a breadcrumb, `Version` has gone from a descriptor of the version type to just a navigation item.

The reason the verbose name of the version type is in there is to differentiate between versions and pull request builds:

image

What you have here is going to make one style for versions (`Versions / {name} / #1234`) and a separate style for pull requests (`Pull request #999 / #1235`). I think we should stick with consistency here and not use the extra breadcrumb section in the listing view -- that is, keep what we have now on the listing view.

The header change mostly works on the build detail view though, but I'd suggest we always have the extra `Versions` breadcrumb for navigation:

- `Versions > {name} > Builds > #1234`
- `Versions > Pull request 999 > Builds > #1235`

readthedocsext/theme/templates/builds/includes/build_name.html
```diff
@@ -33,51 +33,51 @@
...
-        Version {{ version_name }}
-      {% endblocktrans %}
-    {% endif %}
+    <span class="section">
+      <span class="divider">/</span>
+      {% if is_page_title %}
+        <a href="{% url "projects_detail" build.project.slug %}?slug={{ build.get_version_slug }}">
+        {% endif %}
+
+        {{ build.get_version_name }}
```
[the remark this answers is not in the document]
humitos: Updated 👍🏼

agjohnson:
image

The build listing is still off. I don't think we should put a breadcrumb like this in the build list, from my comment here:

- https://github.com/readthedocs/ext-theme/pull/625#issuecomment-3049917495

For consistency that should still be `Version latest` and `Pull request 999`. It's too noisy with the `Versions` breadcrumb in every line.

agjohnson (commented):
And some small adjustments, my main feedback is still at:

- https://github.com/readthedocs/ext-theme/pull/625#issuecomment-3053937982

humitos:
> For consistency that should still be `Version latest` and `Pull request 999`. It's too noisy with the `Versions` breadcrumb in every line.

I tend to agree here. However, the motivation of this PR was to visually distinct the word "Version" from the name of the version itself. So, if we are removing the breadcrumb here I would like to use a different color at least for the name of the version.

An idea:

Screenshot_2025-07-10_11-52-51

* It uses `code` for the version name
* It uses grey color for it

I want to highlight the name of the version somehow over the other elements of that row.

PR #632, opened 2025-07-10
Link to redirect creation page when editing a version

humitos:
This is implements the idea 4) from
https://github.com/readthedocs/readthedocs.org/issues/11978 as a first step.

However, we can get fancier here and automate this creation somehow if we want. There are more ideas for UX in that issue.

### DIFF
--- a/readthedocsext/theme/templates/projects/version_form.html
+++ b/readthedocsext/theme/templates/projects/version_form.html
@@ -53,7 +53,11 @@
           <div class="content">
             {% if version.active %}
               <div class="ui warning message">
-                {% trans "Deactivating a version or changing its slug will result in the deletion of its documentation." %}
+                {% url "projects_redirects_create" project_slug=project.slug as url_redirects %}
+                {% blocktrans trimmed with url=url_redirects %}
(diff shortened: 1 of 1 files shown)


readthedocsext/theme/templates/projects/version_form.html
```diff
@@ -53,7 +53,11 @@ <h2 class="ui medium dividing header">
           <div class="content">
             {% if version.active %}
               <div class="ui warning message">
-                {% trans "Deactivating a version or changing its slug will result in the deletion of its documentation." %}
+                {% url "projects_redirects_create" project_slug=project.slug as url_redirects %}
+                {% blocktrans trimmed with url=url_redirects %}
+                  Deactivating a version or changing its slug will result in the deletion of its documentation.
+                  Make sure to <a href="{{ url }}" target="_blank">create a redirect</a> to avoid not found pages.
```
agjohnson, line 59: "not found pages" could be clearer language.

```suggestion
                  Make sure to <a href="{{ url }}" target="_blank">create a redirect</a> for these URLs to ensure old links continue working.
```

humitos:
Merging it so it goes out today.

PR #634, opened 2025-07-16
Redirect: syntax highlight for `$variable`

### LINKED ISSUE #82: Add redirect template highlighting on create form

humitos:
Screenshot_2025-07-16_16-42-23

Closes #82


readthedocsext/theme/templates/projects/redirect_form.html
```diff
@@ -1,44 +1,44 @@
...
+            <div class="sub header">{% trans "Redirect from" %}</div>
+            <code data-bind="html: redirect_from"></code>
           </div>
           <div class="item">
-            <div class="sub header">
-              {% trans "Redirect to" %}
-            </div>
-            <code data-bind="text: redirect_to"></code>
+            <div class="sub header">{% trans "Redirect to" %}</div>
+            <code data-bind="html: redirect_to"></code>
```
stsewd, line 36: This is likely vulnerable to XSS, we are dealing with user input here.

stsewd: Don't think knockout has a feature like that. This is probably better as a web component, so you don't need to build the HTML manually.

humitos: I used `DOMPurify.sanitize()` which is the same we are already using for the notifications.

stsewd: What you need here is to escape the text, not sanitize it.

stsewd: Sanitizing will still allow inserting HTML

humitos: Yeah, that's not a problem. The problem was XSS, which sanitize solves.

stsewd: Injecting HTML is still dangerous, it can be used to leak information from the page, steal credentials, etc. DOMPurify is being used as an additional layer for notifications, but all values are escaped in notifications.

agjohnson: I think the approach here could be easily tweaked slightly. It is a bit janky you can input HTML into the field and it renders:

image

It's not necessary to sanitize the user input if we just treat it as text instead of HTML, so splitting up our HTML and the user input text would solve this.

Ideally, I wanted this to be a web component so that the same display could be used in the listing view as well. That can be a next step, though I had a POC kicking around in my stash for this too.

src/js/project/admin.js
```diff
@@ -48,34 +49,40 @@ export class ProjectRedirectView {
     this.is_to_url_visible = ko.observable();
 
     this.redirect_from = ko.computed(() => {
-      var from_url = this.from_url();
+      var from_url = DOMPurify.sanitize(this.from_url());
```
agjohnson, line 52: Note, const here should be used instead of var, the value is immutable.

readthedocsext/theme/templates/projects/redirect_form.html
agjohnson: I pushed up a commit that just splits the field up into a prefix and the user content. It's not a great pattern, we should probably just replace this all with a web component that renders everything consistently.

humitos:
Cool! Are we ready to merge this PR?

PR #639, opened 2025-08-13
Rename get_latest_build -> latest_internal_build

stsewd:
Matches https://github.com/readthedocs/readthedocs.org/pull/12400

### DIFF
--- a/readthedocsext/theme/templates/core/project_list_detailed.html
+++ b/readthedocsext/theme/templates/core/project_list_detailed.html
@@ -1,4 +1,4 @@
-{% load i18n %}
+{% load trans blocktrans from i18n %}
 
 {% for project in project_list %}
   <li class="item">
@@ -41,8 +41,6 @@
         {% endif %}
       </li>
     </ul>
-
-
   </li>
 {% empty %}
   <li class="item">{% trans "No projects found" %}</li>


readthedocsext/theme/templates/core/project_list_detailed.html
```diff
@@ -1,29 +1,33 @@
...
+          {% with project.users.all.0 as user %}
+            (<a href="{{ user.get_absolute_url }}">{{ user }}</a>)
+          {% endwith %}
         {% endif %}
       {% endif %}
     {% endwith %}
 
     {% if project.has_good_build %}
-      <span>{% blocktrans with date=project.get_latest_build.date|timesince %}{{ date }} ago{% endblocktrans %}</span>
+      <span>{% blocktrans with date=project.latest_internal_build.date|timesince %}{{ date }} ago{% endblocktrans %}</span>
```
humitos, line 30: This used to be called with `finished=True` by default. Is that change in behavior expected/wanted here?

stsewd: While it was called with finished=True, the real object we were getting was from the prefetch, which returned the latest internal build, either finished or not.

https://github.com/readthedocs/readthedocs.org/blob/bf74954c4e9f70ef75d04867ebbe0a40bccec815/readthedocs/projects/filters.py#L141

stsewd:
Going to merge this so tests pass on .org

PR #640, opened 2025-08-14
Autopaginate fewer objects

ericholscher:
This makes pagination not break on mobile,
and 15 is too many items for something that most users will never use?


agjohnson:
Oh wait. Yeah @humitos is correct. `{% autopaginate 5 %}` isn't 5 pages of autopagination, it's autopaginate by 5 items each time. We don't want that.

agjohnson:
I'll comment there, but I don't think stackable is the answer either.

PR #648, opened 2025-09-03
Add project create modal for repairing the repository listing

### LINKED ISSUE #620: Project: add GHA directions to project creation
#606 adds a button to help cover basic gaps in UX for GHA users, but we are seeing the need for a bit more guidance during this process. There are some cases described here we might want more opinionated UI for:

- https://github.com/readthedocs/ext-theme/pull/606#issuecomment-3009158796

agjohnson:
This is a stronger UI that adds a little more content around why the
user might need to resync/reconfigure permissions. This isn't yet as
aggressive as it might need to be, it's just the scaffolding for the UI
and experimenting a bit more with the UX and feedback to the user.

- Fixes #620


agjohnson:
image
image

agjohnson:
There are some UX issues that were distracting me, they are fairly significant but I am putting these off until later:

- https://github.com/readthedocs/readthedocs-corporate/issues/2046

readthedocsext/theme/templates/projects/partials/project_create_automatic.html
```diff
@@ -77,6 +76,121 @@
...
+
+    This is triggered manually from the ``search_popup_config()`` observable
+    and is set up using the ``semanticui.popup`` binding on the sibling
+    element above.
+  {% endcomment %}
+  <div class="ui small wide primary popup">
+    <div class="header">{% trans "Not finding a repository?" %}</div>
+    <div class="description">
+      <p>
+        This list might be out of date or reconfiguration might be necessary to update repositories.
```
humitos, line 91:
```suggestion
        {% trans "This list might be out of date or reconfiguration might be necessary to update repositories." %}
```

readthedocsext/theme/templates/projects/partials/project_create_automatic.html
```diff
@@ -77,6 +76,121 @@
...
+                <div class="description">
+                  {% blocktrans trimmed %}
+                    If any of your GitHub organizations have not yet approved our GitHub application,
+                    repositories from the organization won't be in your repository list.
+                    In this case, you may need to request approval of our GitHub application from the organization.
+                  {% endblocktrans %}
+                </div>
+                <div class="extra">
+                  <a class="ui right floated button"
+                     href="https://docs.readthedocs.com/platform/stable/reference/git-integration.html#github-permission-troubleshooting">
```
humitos, line 178: This link talks about the OAuth App, not the GitHub App (the new one) -- we should not link to here, or add the documentation for GitHub App.

src/js/application/plugins.js
```diff
@@ -361,6 +354,26 @@ export const semanticui = {
           };
           value(callback);
         } else {
+          if (key === "modal") {
+            // We do something a fun here and move the element into ``body``
```
humitos, line 358: "a fun" is correct or is a typo?

```suggestion
            // We do something fun here and move the element into ``body``
```

humitos: More than fun is a hack to me 😅 

```suggestion
            // HACK: We do something fun here and move the element into ``body``
```

src/js/project/create.js
```diff
@@ -128,6 +136,32 @@ export class ProjectCreateView extends ResponsiveView {
...
+      on: "manual",
+      position: "top right",
+      hoverable: true,
+      closable: false,
+      preserve: true,
+      onHidden: () => {
+        // If the user did something to hide the popup, like click outside the
+        // popup, don't try to show it again.
+        canPopupShow = false;
+      },
```
humitos, line 154: I would remove this, it's pretty confusing having a UI element that doesn't appear anymore even when following the exact same steps. I'm having issues locally to re-trigger the popup after doing a hover and missing clicking on the "Repair" button.

I don't know how to re-trigger the popup to click on "Repair" and see the modal now. If I'm having this issue locally while testing this PR I'm sure that users will hit this a lot.

Refreshing the page re-triggers the popup, but the current behavior is pretty confusing to me.

I prefer to always show the popup on missing repositories.

humitos (commented):
This is going in a good direction.

The modal is still not clear about what is the action I should take. As a user it's hard to know if I need the first or second option from the modal --but I think it's what we have for now until we know how to do detect the GH's user configuration.

I'm not sold on the UX here. The popup looks pretty prominent and I think that's OK. However, I don't like the logic to avoid showing the popup once it was hidden. It's pretty easy to miss the popup by doing hover on it. I would start by removing that logic first.

[Peek 2025-10-06 12-58.webm](https://github.com/user-attachments/assets/1f4eeeab-9a43-4af6-9a6c-0dd7ad17d7d2)


Also, I'm not sure if a popup is the right UI element here. Can't we just add the popup content as a result in the list? I mean, when we show _No matching repositories found_ we can add the button to show the modal there.

Screenshot_2025-10-06_12-55-31

ericholscher:
> The modal is still not clear about what is the action I should take. As a user it's hard to know if I need the first or second option from the modal --but I think it's what we have for now until we know how to do detect the GH's user configuration.

I think this is the big thing. We should probably try to have some clarifying text on which button will actually solve the issue, if we can. When I see those 3 options, I'm mostly just confused about what will actually solve my problem. Perhaps I should just press all 3? 🤷 

> I'm not sold on the UX here. The popup looks pretty prominent and I think that's OK. However, I don't like the logic to avoid showing the popup once it was hidden. It's pretty easy to miss the popup by doing hover on it. I would start by removing that logic first.

Agreed on the popup going away being bad. We should at least have some kind of button that is always available once it's been triggered, because hovering and then losing it is super frustrating. 

> Also, I'm not sure if a popup is the right UI element here. Can't we just add the popup content as a result in the list? I mean, when we show No matching repositories found we can add the button to show the modal there.

I think having it in the list when no matches makes sense, but there's likely also cases where there are _some_ matches, but not the one you want. So we need to account for that as well.

readthedocsext/theme/templates/projects/partials/project_create_automatic.html
agjohnson: Yes and no. We will eventually need to cover this case in UI and in documentation. This issue is the one noted in https://github.com/readthedocs/readthedocs-corporate/issues/2046, where it's not clear how to connect a new GHA organization currently.

But this particular block is still an issue for users of the OAuth app, so we still need to link to this.

The docs around integrations quite jumbled right now. I am going to improve these in a separate PR and might return to this UI afterwards. This is just meant to be a first step.

readthedocsext/theme/templates/projects/partials/project_create_automatic.html
```diff
@@ -77,6 +76,121 @@
...
+            {# This doesn't cover GitHub App providers, the fix for GHA is different #}
+            <div class="item">
+              <div class="content">
+                <div class="description">
+                  {% blocktrans trimmed %}
+                    If any of your GitHub organizations have not yet approved our GitHub application,
+                    repositories from the organization won't be in your repository list.
+                    In this case, you may need to request approval of our GitHub application from the organization.
+                  {% endblocktrans %}
+                </div>
```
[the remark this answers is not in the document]
agjohnson: I did have something like this originally but it was getting very wordy. I'll see if there is a balance here though.

humitos:
Looks better. Can we avoid hiding the popup on hover? It's pretty frustrating UX to me. It's pretty easy to hover the popup and "un-hover" it by mistake while trying to click on "Repair" button.

[Peek 2025-10-07 15-43.webm](https://github.com/user-attachments/assets/db33f157-ebd8-47f2-a8b3-370c4e76f3b7)

agjohnson:
> The modal is still not clear about what is the action I should take

I'd like to get here ,but I don't think this is possible now. We can't suggest a fix because we can't detect what the problem is -- oauth approval in a GH organization can't be detected at all even. I'd rather give the user most options in one place for now.

Also: this UI looks different for users that have just a GHA account attached. Having both GHA and OAuth account attached the modal will have more troubleshooting options.

> but I think it's what we have for now until we know how to do detect the GH's user configuration.

Yup, this is what I'm waiting for. This UI is a first step to improving the troubleshooting steps.

> However, I don't like the logic to avoid showing the popup once it was hidden. It's pretty easy to miss the popup by doing hover on it. I would start by removing that logic first.

There isn't a configuration option for avoiding hover out on popups, but I might be able to intercept the event. I want to retain the logic to show and reshow the popup though, I feel it's bad UX without this.

> We should probably try to have some clarifying text on which button will actually solve the issue, if we can.

This UI is already getting text heavy. I'd like to get this in front of users before optimizing copy much more.

Overall I'd rather we are able to detect what to point the user towards. But also, note above re GHA and GH Oauth accounts -- there are more options in your list because you're connected to both GHA/Oauth, we expect most users to not have both connected.

>> Can't we just add the popup content as a result in the list?
>I think having it in the list when no matches makes sense

I already tried this. It worked okay for when there were no matches and didn't work well at all when there were matches. I've been -1 on this for that and prefer the popup. There might be something better than the popup, but I'd like to just get this in front of users.

agjohnson:
> There isn't a configuration option for avoiding hover out on popups

When I added the popup, I tried a couple things to avoid this but nothing quite worked. The `onHide` event is there, but it breaks hiding on body click events or calling `popup('hide')`.

In 4ae8cb8 I found I could add a long hide delay to mostly mimic this though. This only effects hover focus out, the popup can still be closed by the `popup('hide')` behavior and on body click close without introducing a delay:

https://github.com/user-attachments/assets/cb413c90-ddcc-4b8d-9624-8ed79cf3a335

I feel this is enough to get started with and an improvement to what's there now. Next up would be tuning documentation. I feel our docs should probably be where we provide in depth troubleshooting to supplement this UI being lightweight, instead of bloating this UI much more.

PR #654, opened 2025-09-25
Add title to the status icon

ericholscher:
Add a tooltip to the build status icon so users can understand what each icon state means.

Uses the `whitespaceless` template tag to produce clean tooltip text, and Fomantic UI's `data-tooltip` on a wrapping `<span>` for styled popups consistent with the rest of the UI.

*Generated by Copilot*


readthedocsext/theme/templates/builds/includes/status_icon.html
```diff
@@ -113,5 +113,19 @@
...
+           {% trans "Build finished with errors" %}
+         {% endif %}
+       {% elif build.state == 'cancelled' %}
+         {% trans "Build cancelled" %}
+       {% elif build.is_stale %}
+         {% trans "Build queued" %}
+       {% else %}
+         {% trans "Build in progress" %}
+       {% endif %}">
+</i>
```
agjohnson, line 130: To avoid a multiline attribute, this should all be wrapped with `spaceless` similar to the attributes above this.

ericholscher: I'm not sure spaceless is working..

Screenshot 2025-09-30 at 12 42 38 PM

ericholscher: Spaceless only removes spaces between HTML tags, not inside of text content .

ericholscher:
I ran into this again today and wanted it. I'd love to get something like this in if possible.

ericholscher:
New PR has much nicer native SUI styling:

Screenshot 2026-04-16 at 4 53 03 PM
Screenshot 2026-04-16 at 4 53 00 PM

readthedocsext/theme/templates/builds/includes/status_icon.html
```diff
@@ -38,7 +39,16 @@
 {% endcomment %}
 
 {# fmt:off #}
-<i class="{% spaceless %}
+{% whitespaceless as build_title %}
+  {% if build.state == 'finished' and build.success %}
+    {% trans "Build succeeded" %}
+  {% elif build.state == 'finished' %}
+    {% trans "Build failed" %}
```
humitos: This is the other state we are managing manually below, so I think we should match it here.

```suggestion
  {% elif build.state == 'finished' %}
    {% trans "Build failed" %}
  {% elif build.state == 'cancelled' %}
    {% trans "Build cancelled" %}
```

readthedocsext/theme/templates/builds/includes/status_icon.html
```diff
@@ -1,4 +1,5 @@
 {% load i18n %}
+{% load whitespaceless from ext_theme_tags %}
```
humitos: Does this tag already exist in our code or did you forget to push it?

humitos (approved):
Makes sense to me. I think it could trigger a lot of tooltips when you hover the mouse quickly over these items, but I suppose we can live with that. Maybe we can add a delay or something to avoid that behavior. IIRC, `data-tooltip` has an option for that.

readthedocsext/theme/templates/builds/includes/status_icon.html
ericholscher: Already there.

readthedocsext/theme/templates/builds/includes/status_icon.html
```diff
@@ -38,7 +39,16 @@
...
+{% whitespaceless as build_title %}
+  {% if build.state == 'finished' and build.success %}
+    {% trans "Build succeeded" %}
+  {% elif build.state == 'finished' %}
+    {% trans "Build failed" %}
+  {% else %}
+    {{ build.get_state_display }}
+  {% endif %}
+{% endwhitespaceless %}
+<span data-tooltip="{{ build_title }}"><i class="{% spaceless %}
```
agjohnson: Other elements use `data-content`, which is configured a bit more that `data-tooltip` driven popups. Namely, the delay is tuned to be less aggressive.

https://github.com/readthedocs/ext-theme/blob/e5fb422ac82bfe0baead77bcd90be17ab6c1aa0b/src/js/application/index.js#L134-L140

Another nested span here likely introduces some style issues, there were some rules assuming this structure is `span.image > i.icon`.

The block to override is the wrapping image block:

https://github.com/readthedocs/ext-theme/blob/e5fb422ac82bfe0baead77bcd90be17ab6c1aa0b/readthedocsext/theme/templates/includes/crud/table_list.html#L333-L347

And from there you can add the popup right on the span.

readthedocsext/theme/templates/builds/includes/status_icon.html
agjohnson: Yeah, still need this here too.

agjohnson:
Yeah, tooltips pop up really quick by default. We should avoid this, we've had this bug on listing displays in the past and it's annoying. Other examples already in the templates use `data-content`, which is tuned a bit more and has a longer delay.

https://github.com/readthedocs/ext-theme/pull/654#discussion_r3176231193

ericholscher:
Just going to close this. It makes me sad that we still don't have this feature though, but the review here for something incredibly simple makes me sad.

PR #658, opened 2025-10-10
Add Crispy template for image/file fields

agjohnson:
This fixes UI bugs introduced with

- #614

The native ImageField type doesn't display well but we decided to deal with the
bugs on a later PR. The issues visible there:

- Error fields display twice
- Clear checkbox renders inline and is easy to miss
- The image doesn't display, just the filename

image

image

This PR adds:

- Adds image/file field template to render the file input and uploaded image
- Updates the base field dispatcher template to be cleaner and to support file fields
- Linting and add whitespaceless block we talked about already

### DIFF
--- a/readthedocsext/theme/templates/semantic-ui/layout/field_file.html
+++ b/readthedocsext/theme/templates/semantic-ui/layout/field_file.html
@@ -44,7 +44,7 @@
         <div class="ui basic center aligned segment">
           <img class="ui small bordered centered image"
                src="{{ field.value.url }}"
-               alt="{{ organization.name }} avatar" />
+               alt="{{ field.label }}" />
           <span>{{ field.value.width }}x{{ field.value.height }}px</span>
         </div>
       {% else %}


readthedocsext/theme/templates/semantic-ui/layout/field_file.html
```diff
@@ -0,0 +1,91 @@
...
+
+  {# Confusingly named, but ``is_initial`` signals that there is a current value/file for the file field #}
+  {% if widget.data.is_initial %}
+    <div class="ui top attached segment">
+      {# Is the field for image MIME types? Show the image if so, otherwise basic file info. #}
+      {% if widget.data.attrs.accept == "image/*" %}
+        <div class="ui basic center aligned segment">
+          <img class="ui small bordered centered image"
+               src="{{ field.value.url }}"
+               alt="{{ organization.name }} avatar" />
```
stsewd: Doesn't the field has an attribute specific for this? We are harcoding that this field is for an organization otherwise.

humitos (approved):

Issue #468, opened 2024-08-28
Add project: use richer list display for organization/team


humitos:
We want to use a [richer list display](https://fomantic-ui.com/modules/search.html#category) in the first step on Read the Docs for Business when adding a new project. There are some extra information we want to display in the list in a nicer way:

- status of the organization (e.g. _disabled_)
- Authentication backend (e.g. GitHub SSO, Google SSO, SAML, etc)
- ... maybe more

There are some extra context about this issue in other conversations:
- #466 
- #467

humitos:
I'm adding this issue to my sprint to give it a try. I don't think I will succeed, but at least I want to explore if I'm able to propose something.

PR #660, opened 2025-10-17
Add crispy template for rich choice select field widget

agjohnson:
This adds a field display type for complex dropdown choices and will be 
 used by the project create team and organization selection field.

The only relevant commit is 8b79bd5

 - Refs #468
 - Requires readthedocs/readthedocs.org#12524
 - Based on #658

### DIFF
--- a/readthedocsext/theme/templates/includes/elements/link.html
+++ b/readthedocsext/theme/templates/includes/elements/link.html
@@ -31,8 +31,8 @@
 
 <a href="{{ url }}"
    class="{{ class }}"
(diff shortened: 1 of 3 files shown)


agjohnson:
An example of complex dropdown content is:

image

agjohnson:
> Is that a real example? If so, why the team/organization would be unavailable?

Yes, from 

- https://github.com/readthedocs/readthedocs-corporate/pull/2050

Reasons for disabling the field are from:

- https://github.com/readthedocs/ext-theme/issues/468

PR #662, opened 2025-10-30
Add links to Addons documentation

ericholscher:
This also makes the Analytics pages not show download buttons or graphs with no data.
This feels much cleaner.

Old

Screenshot 2025-10-30 at 8 12 28 AM

New

Screenshot 2025-10-30 at 8 11 46 AM


stsewd:
I think we should also show this message even if there is data but the addon isn't enabled, otherwise the user won't notice that page views aren't being tracked until 30 days have passed.

PR #663, opened 2025-11-05
Don't allow adding domains on subprojects

stsewd:
This action isn't allowed at the form level, it gives a generic error, as it was expecting the UI to block this action, so users won't get this error.

I also noticed that we are blocking the whole list, but we only want to block the "add" action, as users could have had domains already created, we still want users to be able to delete those.

Screenshot 2025-11-05 at 11-58-42 subproject - Domains - Read the Docs Dev

closes https://github.com/readthedocs/ext-theme/issues/666

### DIFF
--- a/readthedocsext/theme/templates/projects/domain_list.html
+++ b/readthedocsext/theme/templates/projects/domain_list.html
@@ -1,6 +1,6 @@
 {% extends "projects/project_edit_base.html" %}
 
-{% load trans from i18n %}
+{% load trans blocktrans from i18n %}
 
 {% block title %}
   {{ project.name }} - {% trans "Domains" %}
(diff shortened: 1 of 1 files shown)


readthedocsext/theme/templates/projects/domain_list.html
```diff
@@ -23,8 +23,24 @@
...
+            <a href="{{ superproject_url }}">{{ superproject_name }}</a>.
+            Custom domains must be <a href="{{ superproject_domains_url }}">added on the superproject</a>.
+          {% endblocktrans %}
+        </p>
+      </div>
+    </div>
   {% endif %}
-  <div class="{% if not enabled %}ui basic fitted disabled segment{% endif %}">
+  <div class="{% if not enabled or project.superproject %}ui basic fitted disabled segment{% endif %}">
     {% include "projects/partials/edit/domain_list.html" with objects=object_list %}
```
agjohnson: In the subproject listing UI, we only display the listing if it's possible to use the listing UI and create subprojects. We should probably do that here instead of just disabling the segment visually.

https://github.com/readthedocs/ext-theme/blob/1831b186331c12700b720e090811cf27e30eaadc/readthedocsext/theme/templates/projects/projectrelationship_list.html#L10-L36

agjohnson: That seems like it should be a validation error on adding a subproject, we shouldn't allow a subproject to be added to a superproject if it has a custom domain. I doubt there would be a lot of projects in this edge case though.

If we're concerned about projects in this state, we should probably still show the domain list. We probably want the user to be able to remove the domain from the project, a disabled list won't allow that. The create button should be disabled either way though.

stsewd: > That seems like it should be a validation error on adding a subproject, we shouldn't allow a subproject to be added to a superproject if it has a custom domain. I doubt there would be a lot of projects in this edge case though.

We already do this validation. We used to block the button in the frontend as well, but this was lost with the migration. We shouldn't allow that action if we know isn't allowed. There are several ways to end up in this state anyway, like when an existing project is converted to be a subproject.

PR #669, opened 2025-11-18
Block GHA migration view for SSO organization users

agjohnson:
This shows the user a verbose error message and blocks the view from
being used for now, and points the user to contact support. From there,
we can either help the user manually migrate or we can at least explain
the limitation and follow up when we are ready to test support for
dual-integration permission checks.

### DIFF
--- a/readthedocsext/theme/templates/profiles/private/migrate_to_gh_app.html
+++ b/readthedocsext/theme/templates/profiles/private/migrate_to_gh_app.html
@@ -2,6 +2,7 @@
 
 {% load provider_login_url from socialaccount %}
 {% load trans blocktrans from i18n %}
+{% load has_sso_enabled from organizations %}
 
 {% block title %}
   {% trans "Migrate account to GitHub App" %}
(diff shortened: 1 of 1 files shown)


agjohnson:
It took a bit to set up to test this, but this is what this looks like locally from a user in a GitHub SSO organization. The steps are disabled individually and can't be navigated to, though the view on the application side could still be used in theory.

image

stsewd:
The changes in the querysets from https://github.com/readthedocs/readthedocs-corporate/pull/2057 worked. The user that was having problems with it already migrated 28 projects :D (I told him to just migrate one project at a time)

ericholscher:
Should we move forward here, or close this out?

ericholscher:
Re-pinging here -- next steps?

stsewd:
We already fixed the problem that prevented users using SSO to migrate to the GH App. The problem now is with users that haven't migrated :D

humitos:
Should we close this PR then?

PR #673, opened 2025-12-02
Expand docs on semanticui binding

agjohnson:
Clarify some of the usage and why each pattern might be useful.


humitos (approved):
Looks good to me.

PR #676, opened 2025-12-03
Fixes for vertical items with images in vertical dropdown lists

agjohnson:
Instead of using a horizontal item, which is prone to running out of room, this uses a vertical item style for the rich select and fixes some display issues that vertical items had when used in combination with an image. This uses a grid layout to make positioning easier. 

image

image


humitos (approved):
Looks good to me. Not sure about the CSS part, though, that I have no idea 🙃 but I trust you 😄

PR #678, opened 2025-12-16
Form improvements for combination repository field

agjohnson:
This work is two things: first a rework of the repo and remote_repository fields, but it also exploring using web components instead of KO. Using KO was turning into a nest of logic and hacks sprinkled through a number of templates and this is a place I'd rather we start using web components.

So there was experimenting on patterns and there are some base classes to help this. The bulk of this UI is in the multifield web component. Much of this change is explained in comments.

Here's what the UI looks like (as of 5b6add4) for first a maintainer with a connected service account and then second a maintainer without a connected service account.

https://github.com/user-attachments/assets/560b4cbc-ad4e-44ac-a3b4-195d0c782f5a

- Requires https://github.com/readthedocs/readthedocs.org/pull/12654

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,8 +23,8 @@ commands:
   run-test:
     description: "Run test suite"
     steps:
-      - browser-tools/install-chrome
-      - browser-tools/install-chromedriver
+      - browser-tools/install_chrome
+      - browser-tools/install_chromedriver
       - checkout
       - run: npm ci
       - run: npm test


agjohnson:
This is settled down, I've updated the description with details.

I'm not quite sure what to do with the UX around the `DONT_SYNC_REMOTE_REPO` feature flag, there are a fair number of projects there.

What I did was leave the manual field enabled, but conditionally the web component won't update the repo field on remote repository field changes if this flag is enabled. I'm not really sure if that is required, but it at least allows for the fields to mismatch on submit.

Issue #583, opened 2025-04-17
GitHub app: connected services page


agjohnson:
Follow up to 

- #570

The comment here shows how we can surface the migration state to the user in the connected services page:

https://github.com/readthedocs/ext-theme/pull/570#issuecomment-2811226267

I would use this UI to communicate which accounts need to be migrated. We should eventually have notifications that point the users to this UI if they have any accounts that require an upgrade to the GHA.

humitos:
I'm adding this to the next sprint and assigning to @stsewd since we want this while doing the migration process.

agjohnson:
@stsewd When you tested multiple account connections, were you logged in to GitHub with both accounts simultaneously? I always use this to log in to multiple accounts and when linking into GitHub, you will be prompted which account you are taking actions against.

![Image](https://github.com/user-attachments/assets/7164525a-9a6f-49cb-813c-c5ec6fbfe430)

stsewd:
@agjohnson nope, I didn't know about that option, I had two containers open with each account.

agjohnson:
It would be worth testing again with the GitHub feature, that might remove the need for anything else. When using multiple accounts and connecting services from Read the Docs, the first UI you get is a switcher to select the account you're using. Perhaps the GHA does (or can do) the same.

stsewd:
So, we are showing a notification when a user connects its account to the GH app and has projects to migrate. Do we still want to expose the status of the migration on the services page? If so, would that be in the GH OAuth item or the GH App item?

There is also this PR that adds a manage button https://github.com/readthedocs/ext-theme/pull/655

PR #680, opened 2025-12-18
Connected services: add actions related to GH App

stsewd:
Closes https://github.com/readthedocs/ext-theme/pull/655

Closes https://github.com/readthedocs/ext-theme/issues/583

I'm not sure if this is the best way to show users that they need to migrate or complete the migration. I think it's better if we just show an info alert at the top? A button in the integration listing feels weird. The manage action looks okay.

Screenshot 2025-12-17 at 22-16-16 Connected services - Read the Docs Dev

### DIFF
--- a/readthedocsext/theme/templates/socialaccount/partials/social_account_list.html
+++ b/readthedocsext/theme/templates/socialaccount/partials/social_account_list.html
@@ -44,21 +44,23 @@
 {% endblock list_placeholder_text %}
 
 {% block list_item_right_menu %}
-  <div class="ui small icon buttons">
-    {% if object.get_provider.id == "github" %}
+  {% if object.get_provider.id == "github" %}
+    <div class="ui small buttons">
       {% if object|has_github_app_account %}
-        <a class="ui icon button"
+        <a class="ui secondary button"
            data-content="{% trans "Complete migration to GitHub App" %}"
            href="{% url "migrate_to_github_app" %}"
-           target="_blank">Migrate</a>
+           target="_blank">{% trans "Migrate" %}</a>
       {% else %}
-        <a class="ui icon button"
+        <a class="ui secondary button"
            data-content="{% trans "Migrate to GitHub App" %}"
            href="{% url "migrate_to_github_app" %}"
-           target="_blank">Migrate</a>
+           target="_blank">{% trans "Migrate" %}</a>
       {% endif %}
-    {% endif %}
+    </div>
+  {% endif %}
 
+  <div class="ui small icon buttons">
     {% if object.get_provider.id == "githubapp" %}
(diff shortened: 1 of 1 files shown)


humitos:
I think we don't have this pattern of a button with text at the right part in any other list. It seems the pattern we need here is probably the three dots with more options inside it, like

Screenshot_2025-12-18_12-09-05

The one that says "Complete migration", I would probably use a ℹ️ or ⚠️ icon in the row with a popup with text explaining the situation. Similar to what we do in the build listing page:

Screenshot_2025-12-18_12-12-31

readthedocsext/theme/templates/socialaccount/partials/social_account_list.html
```diff
@@ -43,16 +44,50 @@
...
+          <i class="fad fa-rocket icon"></i>
+          Migrate to GitHub App
+        </a>
+      {% endif %}
+    </div>
+  {% endif %}
+
+  {% if object.get_provider.id == "githubapp" %}
+    <div class="ui small icon buttons">
+      <a class="ui icon button"
```
agjohnson: You can this button text but this is an icon button, which does not have spacing for text. This button is okay here and not in a context menu if the text is removed.

readthedocsext/theme/templates/socialaccount/partials/social_account_list.html
```diff
@@ -43,16 +44,50 @@
...
+         data-content="{% trans "Manage GitHub App on GitHub" %}"
+         href="https://github.com/apps/{{ GITHUB_APP_NAME }}/installations/new/"
+         target="_blank">
+        Manage
+        <i class="fad fa-external-link icon"></i>
```
agjohnson: This should use a better icon. Other list displays use this for admin/configuration:


```suggestion
        <i class="fad fa-wrench icon"></i>
```

agjohnson:
> @agjohnson what about my suggestion about putting the call to migration at the top instead of inside the listing? That's more prominent and doesn't take space in the listing.

I do like it in the listing, it directly attaches the user action with the integration that is going away. That's helpful context to give the user.

I'm -1 or -2 on relying on notifications and admonitions/prose for UI and UX. Both should be the last resort, always.

In the listing, we just need to be mindful of space considerations. It's also just temporary during the migration, so it's okay if the pattern isn't perfect.

> Also, each listing will have just one of the options, this is, or a migrate button or the manage button, since they belong to different integrations (GH App/GH OAuth).

That seems fine, the integrations don't all align on features so each integration item in that listing isn't quite the same.

The only thing I'd change about that screenshot is to separate the text button and the `.ui.icon.buttons`. This will make two distinct elements -- and we may even style it differently eventually:

image

PR #694, opened 2026-02-04
Inline project action buttons

ericholscher:
Summary:
- Replace the single-item dropdown with a direct Configure button beside View documentation.
- Add an admin-aware web component and tests.

Testing:
- npm run build
- npm test

Issues:
- GitHub: none
- Sentry: none

_Copilot generated_


src/js/modules/menus.js
```diff
@@ -347,6 +347,52 @@ export class ItemDocsElement extends APIConsumerElement {
...
+        <i class="fa-duotone fa-wrench icon"></i>
+      </a>
+    `;
+  }
+}
+customElements.define(
+  "readthedocs-item-project-admin",
+  ItemProjectAdminElement,
+);
+
```
agjohnson: Is there a reason why you implemented this as a new class instead of reusing the existing class? Pretty sure this is the only view the menu item is used.

agjohnson:
For background here, this has a dropdown context menu because there have been other items in this list previously. The context menu is helpful for items that have bad icon matches and would only be confusing to users in icon form. In the context menu, the text is explicit. We don't use text buttons in the listing because it eats space, so context menu serves these two purposes.

If there are no other items to go in the context menu, it can be squashed. If we want other actions in the context menu, we should add them and keep the menu.

What does this look like when the user is not an admin?

Issue #171, opened 2023-06-15
Build: add more output details to the build detail page


agjohnson:
A secondary discussion that came from https://github.com/readthedocs/readthedocs.org/issues/7005 was around some of the additions lined up for the build detail page, which are roughly tangential to adding a debug flag attribute to the BuildCommand model. These additions would aim to increase the user's understanding of the commands and command steps. There are other additions for the build detail page, but just to focus on improvements to the command listing and/or debug output view:

- Show the step that the command executed under
- Show more debug level commands
- Show command descriptions for some commands

The next step is to experiment more with the display here. All of these changes could negatively affect the build command list, so I'll want to go through a few ideas before settling on what works best here.

A few problems we're solving here:

- We aren't surfacing where in the build process the command is executed, which would make build.jobs easier to follow and understand
- We execute commands that aren't clear, for example the `cat` commands might not make sense if the user isn't familiar with POSIX shells. An annotation or description would help explain what is happening to the user.

agjohnson:
I mocked this up really quick. I shows some options I wanted to play with for showing at least a build step in the command list.

Note this doesn't show descriptions though, which would be another decision to work in here. If we want to show descriptions inline, we'd probably want them to look something like _the commented option below_. This would mean we need a visually distinct way to show the build step maybe, or we just call everything a comment.

Also note, this is in debug mode only. I think that is the right call for all of this information. It doesn't pollute the view for most users and allows us to be thorough with output for debug commands.

Also also note, the build step could also just be a metadata label similar to the command timing, and therefore displayed on the non-debug output.

Also also also note, we forgot about the raw logs so far too. :smile: 

Anyways, this is just quick ideas, and I'm not even particularly happy with any, but this will take a few iterations anyways:

![image](https://github.com/readthedocs/ext-theme/assets/1140183/c13e7d74-20f1-4955-b4a4-eeec679a20f7)

humitos:
re "fantasy commands" from the other issue's conversation: IMO, they are more confusing than helpful. If we think that `cat .readthedocs.yaml` is not clear enough because people don't know POSIX, I don't think that `readthedocs-build --show-config` will be clearer. If we want o be clear here, I prefer to show a description instead like `Content of the Read the Docs configuration file (docs/user/.readthedocs.yaml)` --or similar-- (in gray as shown in your example).

[...]

PR #701, opened 2026-02-19
Build detail: switch polling to API v3

ericholscher:
Related to #171.

This updates the build detail page to poll API v3 (`projects-builds-detail`) instead of API v2, while preserving the current UI behavior through a small response normalization layer in `BuildDetailView`.

The build detail template now requests `expand=config` so the existing debug config output continues to render.

Generated by Copilot.


ericholscher:
This is working well for me locally!

Screenshot 2026-02-20 at 11 28 32 AM

PR #724, opened 2026-04-15
Truncate long header content

agjohnson:
This resolves the rest of the display bugs when long header content is used in the model listings. These bugs were prominent in several UIs when using a long version name. Similar to the fix to chip elements, this uses CSS to truncate content only when it doesn't fit the header or table cell. These fixes will apply to all model listing tables and allow the content truncation to change if the viewport size changes.

## Before

image

image

image

## After

image
image

image

image


readthedocsext/theme/templates/projects/partials/edit/redirect_list.html
```diff
@@ -78,10 +78,6 @@
   <i class="fad fa-diamond-turn-right icon"></i>
 {% endblock list_item_image %}
 
-{% block list_item_header_classes %}
-  {{ block.super }} all breaking
-{% endblock list_item_header_classes %}
-
```
humitos, line 84: Is this change related?

src/sui/themes/rtd-application/elements/header.overrides
```diff
@@ -3,29 +3,70 @@
...
+// Truncate some sections of headers where content is too long. This tries to
+// collapse only the necessary content and allows table cells to help control
+// the maximum width allowed. This supports both breadcrumb headers and link
+// headers and is used from both model listing views and pages like the build
+// detail view.
+//
+// Lots of extra specificity here because there are some default rules that can
+// interfere with the display type.
+.ui.ui.ui.ui.ui.truncating.header {
+  display: grid;
```
humitos: Why do we have `.ui.ui.ui.ui.ui....` here? Isn't is enough with just one `.ui`, like `.ui.truncating.header`? If it's not enough, why now? 🤔

humitos (approved):
Screenshots looks good 😄 

Unfortunately, I can't review the CSS code but I trust on you 👍🏼

readthedocsext/theme/templates/projects/partials/edit/redirect_list.html
agjohnson: Oh yeah, so this PR replaces the old method of wrapping the lines by breaking the header (the redirect URL in this case) at any point. This pattern was only used in this template, and so this isn't needed anymore with truncation.

Separately, it's probably worth figuring out what to show in the redirect listing exactly. When projects have a lot of redirects, the listing looks pretty overwhelming.

humitos: Ah, OK, OK. I miss understood this chunk. `all breaking` are the classes added to an list item element 👍🏼

src/sui/themes/rtd-application/elements/header.overrides
agjohnson: This is method of increasing specificity. SUI uses this method in much of their definitions already and we have it a few times. It looks funny but avoids using `!important`, which should be avoided always because of the ways it breaks rules expecting normal specificity.

This pattern works because repeating a class name in a selector doesn't change what is selected, but it does increase the specificity of the selector. ie:

.ui.header is (0, 2, 0) specificity:

image

.ui.ui.ui.ui.header is (0, 5, 0) specificity:

image

Both select the same elements, but the second rule would override the first rule because it has greater specificity.

In this case, there are some rules for headers in SUI that used a number of `:not()` pseudo selectors and so the specificity was quite high.

readthedocsext/theme/templates/projects/partials/edit/redirect_list.html
agjohnson: image

src/sui/themes/rtd-application/elements/header.overrides
humitos: Wow! Gotcha. It's tricky, and looks like a hack on CSS but I think it makes sense? 😄 Thanks for the explanation.

agjohnson: It is a light hack for sure. Another one is to use any nonexistant id in the selector:

```css
// Specificity (1, 3, 0), which is greater specificity than any rule
// matching just classes (0, x, 0)
.ui.truncating.header:not(#doesntexist),
.ui.truncating.header:not(#_) { ... }
```

I'm not sure if it's more or less hacky, but it's an option. It gets points that it will always override SUI rules though. Repeating a class name could still be overridden with a base SUI rule with 1 additional specificity.

https://specificity.keegan.st/ is a handy tool for understanding this all.

agjohnson: Actually, I found a good answer here. I forget we could use LESS variables in the selector:

```less
@{uiImportant}.truncating.header { ... }
```

So the hack is hidden and does a better job that `.ui.ui.ui` in improving specificity.

agjohnson:
Last couple of commits I found a few edge cases in other UIs, the fixes there are also small.

I went with a little bit cleaner way of overriding base SUI rules via higher specificity. There are a number of rules that use repeating `.ui.ui.ui` as a specificity hack but I won't update those here.

src/sui/themes/rtd-application/elements/header.overrides
humitos: How does LESS calculate the amount of `.ui` it has to use? That is possible if LESS has full control over the CSS files, but since we are using an external framework, how does it now if it has to be `.ui.ui.ui` or `.ui.ui.ui.ui.ui.ui` here? 😄

agjohnson: Ah yes so this is all manual calculation, LESS doesn't do anything special with the repeating `.ui`. I wish I could remember which rule was overriding the styles, but it was something like:

```css
.ui.header:not(.someclass1):not(.someclass2):not(.someclass3):not(.someclass4)
```

So that's (0, 6, 0) specificity.

My rule above is:

```css
.ui.truncating.header
```

So that's only (0, 3, 0) specificity so far. 

So what the calculation looks like is to have enough to override the first base SUI rule this rule would need to be (0, 7, 0). So I added 4 additional repeating `.ui` to get there:

```css
.ui.ui.ui.ui.ui.truncating.header
```

And the point above about the weird `:not(#_)` is that it adds (1, 1, 0) to the selector specificity. This is basically tuple math, so (1, 0, 0) > (0, x, 0) for any value of `x`. With this, there's no need to manually calculate how much specificity to add[^1].

[^1]: The exception is SUI could have selectors with ids as well, but this is very unlikely for a library, so this selector should be safe to rely on.

humitos: Great explanation!

agjohnson: Oh and to clarify, the `uiImportant` is just a simple replacement, that's not anything special in LESS:

https://github.com/readthedocs/ext-theme/pull/724/changes#diff-1491d756adeed968212ef7636e91f4741707affee0ea35036c5c065f374fde4bR32

agjohnson:
And note for release, this will need another deep dive on the UI to see if there are any errors introduced in the listing views. I've tested with testing data, but I suspect there will be some leftovers once we get to see the views with some production data.

PR #726, opened 2026-04-16
Automation rule: use KO for dynamic form

humitos:
I don't understand why the info message is not being shown.

I found there is a CSS property applied to these messages by default from `webpack://readthedocsext-theme/node_modules/fomantic-ui-less/definitions/collections/form.less`

```css
.ui.form .info.message, .ui.form .info.message:empty {
  display: none;
}
```

What I have to do to override it? What's the solution here? cc @agjohnson 

[Peek 2026-04-16 17-54.webm](https://github.com/user-attachments/assets/cd944240-32c9-4793-be51-5cbe4a9cf8d9)


src/js/project/admin.js
```diff
@@ -270,19 +270,20 @@ export class ProjectAutomationRuleView {
   static view_name = "ProjectAutomationRuleView";
 
   constructor(automation_rule) {
-    this.predefined_match_arg = ko.observable();
-    this.is_match_arg_visible = ko.computed(() => {
-      let predefined_match_arg = this.predefined_match_arg();
-      return predefined_match_arg === "";
+    this.version_predefined_match_pattern = ko.observable();
+    this.is_version_match_pattern_visible = ko.computed(() => {
+      let version_predefined_match_pattern =
```
agjohnson: This variable is not written, it can be a const scope.

```suggestion
      const version_predefined_match_pattern =
```

readthedocsext/theme/templates/projects/automationrule_form.html
```diff
@@ -17,9 +18,24 @@
...
+        method="post"
+        action="."
+        data-bind="using: ProjectAutomationRuleView()">
     {% csrf_token %}
+    {% alter_field form.version_predefined_match_pattern data_bind="if: true, value: version_predefined_match_pattern" classes="manual select" %}
+    {% alter_field form.version_match_pattern data_bind="enable: is_version_match_pattern_visible" %}
     {{ form|crispy }}
+    <div class="ui info message"
+         data-bind="visible: is_all_versions() || is_semver_versions()"
+         style="display: none">
```
agjohnson: Use the `ko hidden` class pattern instead, don't use inline styles.

This template has similar field interactions and you can see how `ko hidden` classes are used to hide elements:

https://github.com/readthedocs/ext-theme/blob/main/readthedocsext/theme/templates/projects/redirect_form.html

readthedocsext/theme/templates/projects/automationrule_form.html
```diff
@@ -17,9 +18,24 @@
...
+      <span data-bind="visible: is_all_versions()">
```
agjohnson: There is nothing hiding the element by default. This is where `ko hidden` class is needed, to avoid a flash of this content on page load.

readthedocsext/theme/templates/projects/automationrule_form.html
```diff
@@ -17,9 +18,24 @@
...
-  <form class="ui form" method="post" action=".">
+  <form class="ui form"
+        method="post"
+        action="."
+        data-bind="using: ProjectAutomationRuleView()">
     {% csrf_token %}
+    {% alter_field form.version_predefined_match_pattern data_bind="if: true, value: version_predefined_match_pattern" classes="manual select" %}
+    {% alter_field form.version_match_pattern data_bind="enable: is_version_match_pattern_visible" %}
     {{ form|crispy }}
+    <div class="ui info message"
```
agjohnson: Forms have special handling for messages. This is documented in the forms documentation:

https://fomantic-ui.com/collections/form.html#message

```suggestion
    <div class="ui info visible message"
```

humitos:
@agjohnson thanks for the review! It works great now! 👍🏼

humitos:
There are conflicts because we need to re-merge https://github.com/readthedocs/ext-theme/pull/732 again.

agjohnson (approved):

ericholscher:
Should we merge this?

humitos:
I'm not sure to understand the lint errors. I don't have those issues in my local environment 🤷🏼

agjohnson:
I've accidentally done this in the past. If you have any changes not checked in when you build the bundles, those changes will create a different build locally than the bundle on a fresh checkout.

I get the same hashes as CI when I build your branch, so I'm betting this is likely the case. I'd ensure your checkout is clean.

humitos:
I think it was because I was using a different nodejs version or outdated dependencies 🤷🏼

PR #727, opened 2026-04-21
Add dashboard dark mode

### LINKED ISSUE #216: Dark mode on beta dashboard
Would be nice if I can toggle dark mode on https://beta.readthedocs.org/dashboard/ . @humitos said I could submit this feature request here. Thank you for your consideration and for this wonderful project!

agjohnson:
This implements a dark mode for the dashboard by using a high
specificity selector to override the base module styles with each
module's inverted styles, extended using LESS `extend()`.

This does operate on a _CSS rule basis_, not a _LESS rule basis_. So,
it's not enough to just do:

```less
.ui.button :extends(.ui.inverted.button) {}
```

This does not extend all of the child rules. So the rules for each
module extend from child rules for a number of the nested module rules.

Remaining:

- [x] Automatic dark mode by media query
- [x] Add rules for colors for buttons and labels at very least. We commonly use
  specific colors for these elements.
- [x] Fix pseudo class rules where `::after` and `::before` are used, like
  pointing menus
- [x] Maybe for later but inputs are intentionally not inverted by SUI and it
  looks a bit funny in dark mode

----

- Fixes #216
- Requires readthedocs/readthedocs.org#13002

----

image

image

image


agjohnson:
This should be ready to test out and review. The important parts of review:

- Local testing and seeing if any elements are obviously broken.
- Understand roughly how this approach works, documentation is in src/css/dark.less or https://read-the-docs-docs-landing--727.com.readthedocs.build/projects/ext-theme/en/727/api/stylesheets.html#dark-mode
- Understand the template changes to base.html
- At least be aware of how the webpack configuration changed, it's mostly due to internal workings of webpack though. This has potential for breakage.

The actual LESS/CSS rules are not important to understand. All of this was generated from the selectors found in site.css.

### Local testing

- Pull this PR
- Pull https://github.com/readthedocs/readthedocs.org/pull/13002
- Run migration
- Set theme to dark/system in admin dashboard for core/profile for your user
- Reload the page. Theme `dark` will always be `dark`, theme `system` should follow your system/browser theme preference

### And next up

With this merged and out in production, the next steps after this PR are:

- [ ] Continue iterating on styles. This is a mostly complete pass but I expect more style tweaks.
- [ ] Input boxes have no inverted variant. The top search box is an input and dropdown search, all not inverted.
- [ ] Inverted segments used in a few places like destructive actions could be less prominent
- [ ] I think this all might actually make sense as a standalone NPM package, as a general dark theme solution for FUI. It might disappear from our config entirely.
- [ ] Or if not, there should be a more reproducible version of parsing out all the inverted rules. So far, only the main elements and styles we're using are covered.

stsewd:
Some places that I found with low contrast.

image

image

image

humitos:
> * Understand roughly how this approach works, documentation is in src/css/dark.less or [read-the-docs-docs-landing--727.com.readthedocs.build/projects/ext-theme/en/727/api/stylesheets.html#dark-mode](https://read-the-docs-docs-landing--727.com.readthedocs.build/projects/ext-theme/en/727/api/stylesheets.html#dark-mode)
> * Understand the template changes to base.html
> * At least be aware of how the webpack configuration changed, it's mostly due to internal workings of webpack though. This has potential for breakage.

The offsite could be a good opportunity to talk more about this. Maybe it's easier to present/explain these changes to the whole team so we can ask more specific CSS questions and try to refine our understanding around frontend.

agjohnson:
> The offsite could be a good opportunity to talk more about this.

Yeah, always happy to explain and share more on this stuff. Though the technical parts of the CSS are not super important I'd say, it's more the general concept of extending the styles that's important. There's a lot here, but it's just redundant rules.

agjohnson:
> Some places that I found with low contrast

I put up small fixes with [b140ed8](https://github.com/readthedocs/ext-theme/pull/727/commits/b140ed827ad49f5519f37f0a40f2934f68f65910)

humitos:
OK, I pulled down this PR and browsed around the whole site. Regarding the style and based on my personal preference, I think the background is too dark. It's basically _black_ and I haven't seen any other app that use black, instead of a dark grey. I'd suggest to change that to match other applications.

Here are my comments with screenshots.

|Screenshot|Comment|
|----|----|
|Screenshot_2026-05-07_11-56-22|We have a few inputs that are white. I think they should be grey as well.|
|Screenshot_2026-05-07_11-55-09|This section and also notifications don't look great to me on dark mode. They are too prominent and I don't think we want the user to look at them directly all the time.|
|Screenshot_2026-05-07_11-54-44|There are some black horizontal lines on these lists that I suppose they shouldn't be there.|
|Screenshot_2026-05-07_11-51-54|"Wrap line" text is invisible on the build details page.|
|Screenshot_2026-05-07_11-51-46|Slash separators are low contrast.|
|Screenshot_2026-05-07_11-51-26|Another input that it's white.|
|Screenshot_2026-05-07_11-50-04|QR code from 2FA needs more contrast.|
|Screenshot_2026-05-07_11-49-27|This is what I refer about notifications and danger section that don't look great to me on dark mode.|
|Screenshot_2026-05-07_11-49-17|This green from "Add project" looks has too much light compared with other greens we use in the site. We may want to unify them.|
|Screenshot_2026-05-07_11-48-35|Circle contours are different.|

agjohnson:
Thanks for the thorough test! There are some minor things there, but not much that I feel needs to hold up this PR. I noted above but I'm looking to iterate on the styles with this PR merged, instead of kicking this PR along for too long. Any text that is outright illegible I'd take on immediately still though, there are a couple issues there.

A couple major items:

> I think the background is too dark. It's basically black and I haven't seen any other app that use black, instead of a dark grey. I'd suggest to change that to match other applications.

Github is `rgb(13 17 23)`, ours is `rgb(37 42 49)`, so considerably lighter actually. I don't feel we need a lighter dark grey here, we want the inverted segments to stand out from the background.

> We have a few inputs that are white. I think they should be grey as well.

I already noted above that search input fields have no inverted styles, so that is some of the inputs, and one of the inputs you pointed out. In fact, fields had no inverted styles at all, this is custom.

So, I'll leave this for now unless there is an easy answer, but it's something I'm already tracking for next iterations.

> This section and also notifications don't look great to me on dark mode. They are too prominent and I don't think we want the user to look at them directly all the time.

I'd like to redo these styles to be less prominent, but right now there are no inverted styles that match the non-inverted styles we're using there -- the non inverted styles have just a mild color hint border on one side, the inverted styles are what you see.

https://fomantic-ui.com/elements/segment.html#colored

I'd like to replicate the non inverted colored styles I think. But I'd save this for later, what's there is doing fine, it should be prominent, just not that prominent.

stsewd (approved):

agjohnson:
I fixed some of the issues above, I didn't address the remaining input elements that I noted above or the QR code. The QR code only needs `filter: invert()` but targeting it will need some fiddling more with allauth templates.

This should be good to go out for some deeper testing

PR #730, opened 2026-04-21
projects: add filter to redirect list view

ericholscher:
Companion template changes for readthedocs/readthedocs.org#12972.

- Wire filter controls (redirect type dropdown, URL search) onto the redirect list page.
- Extend `includes/filters/form.html` to render a Semantic UI text input for `CharFilter` fields, alongside the existing hidden/dropdown branches.

Related: readthedocs/readthedocs.org#12972

---

Generated by Copilot.

### DIFF
--- a/readthedocsext/theme/templates/projects/partials/edit/redirect_list.html
+++ b/readthedocsext/theme/templates/projects/partials/edit/redirect_list.html
@@ -3,6 +3,9 @@
 {% load trans blocktrans from i18n %}
 
 {% block top_left_menu_items %}
+  <div data-bind="using: FilterView()">
+    {% include "includes/filters/form.html" with fields=filter.form %}
+  </div>
 {% endblock top_left_menu_items %}
 
 {% block create_button %}


ericholscher:
Current version:

Screenshot 2026-04-21 at 7 39 24 PM

Also did a version that looked like this, but was more complex, so I rolled it back:

Screenshot 2026-04-21 at 7 53 06 PM

readthedocsext/theme/templates/includes/filters/form.html
```diff
@@ -8,6 +8,18 @@
...
                value="{% if field.value.0|length == 1 %}{{ field.value|default:"" }}{% else %}{{ field.value.0|default:"" }}{% endif %}" />
+      {% elif field.field.widget.input_type == "text" %}
+        <div class="item">
+          <div class="content">
+            <div class="ui icon input">
+              <input type="text"
+                     name="{{ field.name }}"
+                     value="{{ field.value|default:"" }}"
+                     placeholder="{{ field.label }}" />
+              <i class="fa-solid fa-magnifying-glass icon"></i>
```
ericholscher: This is the main UI thing I'm not sure about -- there isn't a nice way to clear it other than clearing the text and re-submitting. The UI also looks different. The AI had a go at a more complex version, but didn't love it. 

I think this is way better than not supporting filtering at all. i'd be fine with just a type filter for now, but I think the freeform text is important for usability.

humitos:
> Also did a version that looked like this, but was more complex, so I rolled it back:

I think this complex version (the one you searched for `shad`) is what we have in other places. Look build listing page when filtering for version.

Screenshot_2026-04-21_14-19-17

agjohnson:
Yup, I'd use prior art here. The dropdown with search input is what should be implemented. This mostly all happens on the application side.

ericholscher:
Yea.. do we want to select from a list, or allow freeform? Selecting from a list could be heavy with a lot of redirects, but probably fine? 

I think being able to submit arbitary text or substrings is a really useful feature, but maybe a v2 to get something out?

agjohnson:
> Yea.. do we want to select from a list, or allow freeform?

It's not clear what you're describing, but the project and version dropdowns already use both list and text input selection. This is the UI the redirect filter should aim for.

> I think being able to submit arbitary text or substrings is a really useful feature

This is already supported by the filter elements we're describing:

image

I don't see us needing anything complex here, I'd keep it simple.

agjohnson (approved):
This looks good, it's all that should be needed on the front end side to display the filter form.

humitos:
> I think being able to submit arbitary text or substrings is a really useful feature, but maybe a v2 to get something out?

Yeah, I've been wanting this feature forever. Unfortunately, none of the filters work in this way. I'd say we should stick with the pattern we current have and then change them all to allow submitting substrings 👍🏼

agjohnson:
Well again, the dropdowns already support substring search.

What isn't supported yet is the model listing querysets using these non-exact matches. We've talked how to implement this in the existing filter fields already. I would not use a separate filter field for this.

I've wanted this as well, but agreed this is very much separate work and not redirect specific.

PR #733, opened 2026-04-24
footer: submit language form on select, drop Update button

ericholscher:
## Summary

The `Version` / `Language` / `Update` row in the footer looked awkward because the `Update` submit button sat next to the language dropdown in the `very relaxed horizontal inverted list`.

Rather than styling around it, this drops the button entirely and submits the language form on select, which was already called out as a TODO in the template.

A new `FooterView` exposes a `language_config` FUI dropdown config with an `onChange` handler that submits the parent form, mirroring the existing `FilterView` pattern in `src/js/modules/filter.js`.

## Test plan

- [ ] Visual check: footer metadata row shows `Version` and `Language` cleanly with no stray button.
- [ ] Open the language dropdown and select a different language &mdash; page reloads in the chosen language.
- [ ] Selecting the currently active language does not cause a reload loop.

---

Generated by Copilot.

---
_Generated by [Claude Code](https://claude.ai/code/session_01AXMNkQhUnb4gT9xi1Fsbb5)_


humitos:
I don't remember exactly why, but I think there was a reason to have a button -- @agjohnson do you remember?

src/js/modules/footer.js
```diff
@@ -0,0 +1,27 @@
+import { Registry } from "../application/registry";
+
+/**
+ * View for the site-wide footer.
+ *
+ * Exposes ``language_config``, the FUI dropdown configuration for the language
+ * selector. Selecting a language submits the parent form, so there is no
+ * separate Update button.
+ */
```
agjohnson, line 9: I realized this might not be explained anywhere, no one has written public docs here yet but me.

This comment style (`/** **/`) is used by jsdoc to parse out docs for our hosted docs. We could make it more explicit if we want at some point, it's a sphinx-anysrc thing. The docs here are mostly just internal notes, so I'd do `//` style to avoid bring this content into our docs:

```suggestion
// View for the site-wide footer.
//
// Exposes ``language_config``, the FUI dropdown configuration for the language
// selector. Selecting a language submits the parent form, so there is no
// separate Update button.
```

I'm also fine skipping documentation that just explains what code is doing. We mostly need to capture the high level _why_, etc.

src/js/modules/footer.js
```diff
@@ -0,0 +1,27 @@
...
+  constructor() {
+    this.language_config = {
+      direction: "upward",
+      fullTextSearch: true,
+      cache: false,
+      onChange: (value, text, $elem) => {
+        if (!value) {
+          return;
+        }
+        $elem.closest("form").submit();
```
agjohnson, line 22: Getting the form through jQuery like this is okay as a one off thing, we do something similar in couple other views, and especially where FUI gives us a jQuery object in a callback like this.

Otherwise, using jQuery or DOM manipulation inside KO or FUI should be avoided. Pattern mixing gets confusing and pollutes the patterns more than needed. I note at these points where this pattern creeps in:


```suggestion
        // Note: don't copy this pattern of using jQuery directly. Use FUI or KO. This is an exception to this rule.
        $elem.closest("form").submit();
```

agjohnson (commented):
This looks good, just sharing some notes on the standards here, please look those over.

I think the button was previously required because the view or language change form actually inspected the button input element in processing the form submit. This appears to work okay so that likely changed.

High level, this is totally fine as a KO view. However this is where I've been switching to writing web components. It's a really basic element so would be a good place to chip away at writing less KO and to practice low stakes WC. But either works.

Also high level, I am considering moving this out of the footer entirely. It should be in settings (or duplicated in settings), it's not easily discoverable in the footer. I'm thinking it will end up on a "Appearance" user settings page with the dark mode theme option as well. If we want it in both places that's fine, not much opinion there.

PR #735, opened 2026-04-29
Clarify integration terminology to "Git integrations"

ericholscher:
Matches the docs (https://docs.readthedocs.com/platform/stable/reference/git-integration.html) 

https://claude.ai/code/session_01XkTHGXh9d9m3iiWKR3fktd

### DIFF
--- a/readthedocsext/theme/templates/projects/edit_base.html
+++ b/readthedocsext/theme/templates/projects/edit_base.html
@@ -35,7 +35,7 @@
           <div class="header">{% trans "Building" %}</div>
           <div class="menu">
             <a class="item {% block project_integrations_active %}{% endblock project_integrations_active %}"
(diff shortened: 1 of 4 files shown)


agjohnson:
We support generic API integrations, so I don't think this is something we should change. 

image

We could be more specific about "Git integration" when talking about the other integrations, though I don't love how we use the term already either. For the top level concept, menu, headers, etc, "Integrations" still feels fine to me.

agjohnson (commented):
We could be more specific about "Git integrations" specifically where we mention GitHub/et al integrations, though I think that's mostly at the form.

"Integrations" still fits well given the option for generic API integrations.

ericholscher:
We should improve our docs then, because we don't mention Integrations anywhere that I can tell.

PR #739, opened 2026-05-14
theme: Fix signup page header and copy

ericholscher:
Human summary: The signup page had a header of "Log in" -- fix that, and tidy up the signup copy.

The signup template (`account/signup.html`) extends `account/login.html`. It was overriding a block (`authentication_header`) that the parent had renamed to `content_title_text`, so the override was silently dropped and the signup page rendered the parent's "Log in to continue" title.

**Key changes:**
- Fix the orphaned block override so the signup title renders again.
- Title now reads "Sign up for Read the Docs Business" / "Sign up for Read the Docs Community", chosen via `USE_ORGANIZATIONS` so it works in dev too.
- Remove the redundant "Sign up using:" label above the auth tabs.
- Center the signup title.
- Use specific template tag imports (`crispy from crispy_forms_tags`, `trans blocktrans from i18n`) to clear djlint `T901` warnings.

---
*Generated by AI agent*

https://claude.ai/code/session_01DGqPZ1528y9aXikwnZTsBG


ericholscher:
Screenshot 2026-05-14 at 11 48 39 AM

readthedocsext/theme/templates/account/signup.html
```diff
@@ -19,7 +27,6 @@
   {% trans "Sign up with email" %}
 {% endblock authentication_email_text %}
 {% block authentication_header_text %}
-  {% trans "Sign up using:" %}
 {% endblock authentication_header_text %}
```
humitos: Should we remove this block completely or do we need to keep to override it as blank?

humitos (approved):
:shipit:

readthedocsext/theme/templates/account/login.html
```diff
@@ -11,8 +11,10 @@
 {% block content_title_text %}
   {% if is_from_cas_login %}
     {% trans "Log in to view this documentation" %}
+  {% elif USE_ORGANIZATIONS %}
+    {% trans "Log in to Read the Docs Business" %}
```
humitos: we use "Read the Docs _for_ Business"

readthedocsext/theme/templates/account/signup.html
```diff
@@ -1,26 +1,24 @@
...
+    {% trans "Sign up for Read the Docs Community" %}
+  {% endif %}
+{% endblock content_title_text %}
 
-{% block authentication_vcs_text %}
-  {% trans "Connect account" %}
-{% endblock authentication_vcs_text %}
 {% block authentication_email_text %}
-  {% trans "Sign up with email" %}
+  {% trans "Email" %}
```
agjohnson: This is actually the default text from the underlying block, it can be removed if there isn't a specific reason to have it here.

agjohnson (approved):

PR #745, opened 2026-05-22
builds: auto-expand failed commands on the build detail page

ericholscher:
## Context

On the build detail page, a failed command should expand its output so the error is visible without manually clicking to open it.

This only works for live builds today. The old dashboard expanded failed commands by default — it rebuilt every command object on each poll, so the "expanded if failed" state was always re-evaluated. The new dashboard reimplemented this as a Knockout subscription, which only fires on changes *after* the command is constructed. Opening an already-finished failed build never triggers it, so the failed command stays collapsed. It only works if you watch the build fail live while polling.

This PR makes failed commands expand on initial page load too.

---
*Generated by Claude Code.*


ericholscher:
Not a great video, but I reload the page and it auto-scrolls:

https://github.com/user-attachments/assets/9476fa8b-567c-4738-b314-1425cafd0ca1

src/js/build/detail.js
```diff
@@ -593,6 +600,27 @@ export class BuildDetailView {
...
+    if (!failed_command) {
+      return;
+    }
+    const elem = document.getElementById(
+      "build-command-" + failed_command.id(),
+    );
+    if (elem && elem.scrollIntoView) {
+      elem.scrollIntoView({ behavior: "auto", block: "start" });
+    }
+  }
```
agjohnson: This is a pattern rather separate to what we already do in the models. This should follow what `set_selected_line_by_hash` is doing instead. Then there is no need to replicate behavior that is already in the model view, like expanding the selected command.

agjohnson (commented):
This is close but is mixing in a new pattern that avoids existing logic in the views.

We are using the KO models directly to call `set_selected_line` on the command model to perform targeting of build command and command output line number. This should also call `set_selected_line` in a similar way.

This PR is instead adding a id and breaking out of KO models to target via the DOM, which avoids the existing KO view logic that happens when setting a line as selected -- like expanding the command that is selected, etc.

ericholscher:
Scoped things back to just the auto-expand logic for now. Moved the auto-scroll larger change here: https://github.com/readthedocs/ext-theme/pull/750

src/js/build/detail.js
```diff
@@ -111,8 +111,9 @@ class BuildCommand {
       }
     });
 
-    /** @observable {Boolean} Is this command expanded? */
-    this.is_expanded = ko.observable(false);
+    /** @observable {Boolean} Is this command expanded? Failed commands are
+     * expanded by default. */
+    this.is_expanded = ko.observable(this.exit_code() > 0);
```
agjohnson: Ah so the code I was thinking of is directly below this. But it won't get the instantiation value, only updates to the value through the API. I'm guessing if you sit and watch a build progress, the command will expand automatically, but a finished build it won't.

The only thing on the code here is that accessing an observable should only be done outside the instantiation scope. There are side effects to calling an observable (`this.exit_code()`) in instantiation.

A direct value is better and avoids unnecessary computations and side effects. The exit_code above could be a const too, or just use the same value:


```suggestion
    this.is_expanded = ko.observable((build_command.exit_code || 0) > 0);
```

agjohnson (approved):

humitos:
Simple but super useful 💯

PR #746, opened 2026-05-26
builds: Remove debug mode toggle functionality

ericholscher:
Remove the debug mode toggle feature from the build detail view. This simplifies the UI by eliminating the ability to hide/show debug commands and removes associated observable state management.

All debug commands are now always visible in the build output, eliminating the need for the toggle mechanism.

---
*Generated by AI agent*

https://claude.ai/code/session_012d3umT3c4chkcXvQmNwcRQ


ericholscher:
@agjohnson We talked about removing the user facing debug mode at the offsite. This PR just always shows the comments, but they are greyed out -- seems like a nice simplification, but not sure if it's a priority?

ericholscher:
We talked about this in a call, so going to hold off on this for now, and just focus on adding the Bill detail files changed view.

agjohnson:
Yeah I think hiding these commands is still important. `cat .readthedocs.yml` is pretty hacky UX and if the users doesn't know what `cat` means they are never going to click on that.

If these were descriptive I wouldn't have a problem showing them on builds. The first debug command was an example of using a more descriptive (and fake) command. It still looks like a command, so doesn't break out of the UX for the command list UI, and it is at least descriptive about what is happening.

For now, the debug menu option seems okay. A few options to avoid work here for now:

- The menu could just wrap for some view ports sizes for now.
- The debug and wrap lines option could become a context menu.
- I've mentioned a lot about how there should not be two different locations for a view docs button. The per-build docs could move up and combine with the per-project view docs button. This should really be more complex UI, allowing for version selection in both cases.

PR #749, opened 2026-05-26
Add profile preferences view and reorganize user settings menu

### LINKED ISSUE #743: Add user profile admin for display settings
To allow user configuration of the dashboard theme, we need to expose an option that allows for theme selection. Most services I've seen use a user admin panel with a list of the options, I'm going to follow this pattern.

I'm also thinking it makes sense to include other options on the page in the future, perhaps language locale would be a next option to move/copy into the admin view.

- [x] Add new user profile admin view for "Display" options with additional form handling
- [ ] Consider default for theme more, maybe drop undefined theme option

agjohnson:
Just a simple form for now.

Also updates the menu to reorganize for a "preferences" concept. We'll have more submenus there at some point.

- Fixes #743
- Requires https://github.com/readthedocs/readthedocs.org/pull/13074

image

### DIFF
--- a/readthedocsext/theme/templates/profiles/base_edit.html
+++ b/readthedocsext/theme/templates/profiles/base_edit.html
@@ -22,13 +22,15 @@
             <div class="header">{% trans "Account" %}</div>
             <div class="menu">
               <a class="item {% block profile_admin_details %}{% endblock %}"
-                 href="{% url 'profiles_profile_edit' %}">{% trans "Account" %}</a>
+                 href="{% url 'profiles_profile_edit' %}">{% trans "Profile" %}</a>
               <a class="item {% block profile_admin_change_email %}{% endblock %}"
                  href="{% url 'account_email' %}">{% trans "Email addresses" %}</a>
               <a class="item {% block profile_admin_social_accounts %}{% endblock %}"
                  href="{% url 'socialaccount_connections' %}">{% trans "Connected services" %}</a>
-              <a class="item {% block profile_admin_preferences %}{% endblock %}"
-                 href="{% url 'profiles_preferences_edit' %}">{% trans "Preferences" %}</a>
+              {% if USE_PROMOS %}
+                <a class="item {% block profile_admin_gold_edit %}{% endblock %}"
+                   href="{% url 'gold_detail' %}">{% trans "Gold membership" %}</a>
+              {% endif %}
             </div>
           </div>
           <div class="item">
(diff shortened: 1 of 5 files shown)


readthedocsext/theme/templates/profiles/base_edit.html
```diff
@@ -27,6 +27,8 @@
                  href="{% url 'account_email' %}">{% trans "Email addresses" %}</a>
               <a class="item {% block profile_admin_social_accounts %}{% endblock %}"
                  href="{% url 'socialaccount_connections' %}">{% trans "Connected services" %}</a>
+              <a class="item {% block profile_admin_preferences %}{% endblock %}"
+                 href="{% url 'profiles_preferences_edit' %}">{% trans "Preferences" %}</a>
```
ericholscher: This is a very ambiguous title -- Should we call it `Display preference` or similar unless we have more to add to the form? I wouldn't know to click on this.

ericholscher (approved):

readthedocsext/theme/templates/profiles/base_edit.html
stsewd: GitHub, and Instagram call this "appearance"

agjohnson: I was leaning towards "Display" or "Appearance" originally, but we don't have more to go into this form if it's just appearance configuration. It might be hard for us to even fill out an admin page that is general dashboard preferences, really.

I don't expect users to naturally find this page either way, so I'm more just thinking we just have a form that's easy to maintain.

Renaming and grouping preferences in the sidebar helps:

image

Changes here are:

- This new menu is Preferences:Dashboard for now, but could be separate menus for Appearance, Notifications, Accesibility, Privacy, etc later.
- Account:Account menu option is now Account:Profile to make this menu make more sense (this matches closer)

Not pictured here but also:

- Advertising:Advertising is now Preferences:Advertising
- Advertising:Gold membership is now Account:Gold membership (this matches closer too though I dislike the Gold membership framing pretty generally)

ericholscher (approved):
Yea, this seems way nicer and cleaner. I'd love to see a few of these brought up to the user menu dropdown as discussed as well.

Though I think the Dark Mode stuff is still quite hidden. I think if we don't tell users about it loudly (or set `system` as default) 99% of users won't know about it.

I'm fine shipping this though, along with some docs explaining how to enable it, so users/agents can find it.

 **I think this would also be a great call topic (eg. design the user preferences menus, and the user dropdown in the dashboard). Would spark some good feature ideas and things we wish were there I bet, and verbiage like this is great to go back and forth on in real time. **

readthedocsext/theme/templates/profiles/base_edit.html
ericholscher: Yea, I'd be fine scaling back or removing the whole `Gold` concept at this point, tbh. Though it's probably more trouble than it's worth.

agjohnson: Yeah same. An improvement here would be to reframe gold membership as a _subscription_ instead. The UI would be a bit cleaner at least, even if we didn't do any major code changes.

It looks okay and makes sense as "Gold membership" under "Account" menu at least for now.

agjohnson:
Yeah I'm just aiming for something incremental to get it in front of users soon. The next batch of work is still more onboarding, I'm just going to keep chunks digestible.

I'm planning on onboarding UI in the dashboard, but outside the user menu so far. The user menu UI is also hidden, so doesn't quite help in broadcasting the feature, and it's a setting you should only need to interact with once, so a quick change UI in the user menu feels not super userful. There are other features I wanted to add to that dropdown though, so I'm interested in planning more.

ericholscher:
Cool, then let's ship it!

humitos (approved):

PR #750, opened 2026-05-27
builds: jump to the failed command via #error hash or button

ericholscher:
## Context

When a build fails, give the user a way to jump to the failed command without scrolling through the log — but only when they ask for it.

[The earlier version of this PR](https://github.com/readthedocs/ext-theme/pull/750/commits) scrolled unconditionally when polling finished, which @agjohnson pointed out was jarring: loading any failed build dragged you to the bottom of a potentially very long page whether or not that's what you came for. The trigger now has to come from the user:

- **`#error` URL hash** — opening `/.../builds/123/#error` resolves to the first failed command, expands it, and scrolls. Intended for deep-linking from GitHub PR comments.
- **Jump to failed command** button on the old-style build-error banner (the one shown when `build.error` is set).

Both routes go through `set_selected_line` so the existing KO view logic runs (per @agjohnson's earlier review on #745): expand the command, mark the line, scroll into view, update the URL hash.

A synthetic build-error notification (for the case where the build failed but no notifications were attached) is a possible follow-up.

---
*Generated by Claude Code.*


src/js/build/detail.js
```diff
@@ -344,6 +344,10 @@ export class BuildDetailView {
     this.builder = ko.observable(build.builder);
     /** @observable {Array.<Object>} Build command objects as an array */
     this.commands = ko.observableArray(build.commands);
+    /** @computed {Boolean} Does the build have at least one failed command? */
+    this.has_failed_command = ko.computed(() => {
+      return this.commands().some((command) => command.exit_code() > 0);
```
humitos, line 349:
```suggestion
      return this.commands().some((command) => command.exit_code() !== 0);
```

humitos: Also, should we consider 183 for skip builds as a special case here?

src/js/build/detail.js
```diff
@@ -593,6 +604,22 @@ export class BuildDetailView {
...
+  /**
+   * Select the first output line of the first failed command.
+   *
+   * Reuses :meth:`set_selected_line` so the existing view logic runs:
+   * expanding the command, marking the line, scrolling to it, and updating
+   * the URL hash.
+   */
+  set_selected_line_from_first_failed_command() {
+    const failed_command = ko.utils.arrayFirst(this.commands(), (command) => {
+      return command.exit_code() > 0;
```
humitos, line 616:
```suggestion
      return command.exit_code() !== 0;
```

readthedocsext/theme/templates/builds/build_detail.html
```diff
@@ -215,6 +215,11 @@
               {% trans "There was an error with this build" %}
             </div>
             <p data-bind="text: error"></p>
+            <a class="ko hidden ui inverted basic small button"
+               data-bind="css: { hidden: !has_failed_command() }, click: set_selected_line_from_first_failed_command">
+              <i class="fad fa-arrow-down icon"></i>
+              {% trans "Jump to failed command" %}
+            </a>
```
humitos, line 222: Can you add a screenshot showing how this looks?

humitos (commented):
I think this behavior makes sense to me.

PR #753, opened 2026-05-30
Configure package for NPM minimum age

agjohnson:
Uses npm 11.10 which has the configuration for package minimum age.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,2 @@
+engine-strict=true
+min-release-age=14400
--- a/.npmrc
+++ b/.npmrc
@@ -1,2 +1,3 @@
 engine-strict=true
-min-release-age=14400
+# Don't install packages younger than 14d, also set in .ncurc.mjs
+min-release-age=14
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -6,6 +6,9 @@ build:
     # Python 3.9 is required here since we are using an old version of Sphinx
     python: "3.9"
     nodejs: "22"
+  jobs:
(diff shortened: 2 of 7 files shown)


.npmrc
```diff
@@ -0,0 +1,2 @@
+engine-strict=true
+min-release-age=14400
```
humitos: How much time is this? We could add a comment here.

humitos (approved):
I suppose this is about security concerns, right? I think I probably missed a conversation here. If you have a link to the conversation, can you paste it here?

.npmrc
agjohnson: The unit here is seconds. The config wants days either way though.

agjohnson:
It's linked above

- https://github.com/readthedocs/meta/issues/225

agjohnson:
Build failure is jsdoc, probably stemming from 24.16 upgrade. I'll try fixing this quickly but we can also address that separately if needed.

Edit: Yup, minor upgrade needed.

.readthedocs.yaml
```diff
@@ -1,11 +1,16 @@
...
 
 build:
-  os: ubuntu-22.04
+  os: ubuntu-26.04
   tools:
     # Python 3.9 is required here since we are using an old version of Sphinx
     python: "3.9"
-    nodejs: "16"
+    # TODO use 24.16 when we update our build tools
+    nodejs: "24"
```
humitos, line 9: We can't specify the minor version on nodejs. Once we upgrade build tools, the minor version will be updated automatically.

PR #755, opened 2026-06-02
theme: don't open the GitHub App skip-step link in a new tab

ericholscher:
The "Skip this step" link in the GitHub App installation-target empty state used `target="_blank"`, but it navigates within the same migration flow (`?step=revoke`). Opening it in a new tab leaves the original tab on the previous step. Removed `target="_blank"` so it navigates in place.

---
🤖 Generated with [Claude Code](https://claude.ai/code)

---
_Generated by [Claude Code](https://claude.ai/code/session_01Tb9Ea7oNTCDqt3huHWu6Pa)_

### DIFF
--- a/readthedocsext/theme/templates/profiles/partials/github_app_target_list.html
+++ b/readthedocsext/theme/templates/profiles/partials/github_app_target_list.html
@@ -26,8 +26,7 @@
 {% block list_placeholder_text %}
   <a class="ui button"
      href="?step=revoke"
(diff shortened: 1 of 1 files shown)


agjohnson (approved):
Yeah don't see any reason why this should be a new tab. If this was on purpose, @stsewd would know why.

stsewd:
It was probably a copypasta from another place.

PR #760, opened 2026-06-10
User: Remove homepage field from UserProfile

stsewd:
Ref https://github.com/readthedocs/readthedocs.org/pull/13125


humitos:
What's the reason to remove this field?

stsewd:
> What's the reason to remove this field?

I just updated the description in https://github.com/readthedocs/readthedocs.org/pull/13125.

PR #766, opened 2026-07-08
Add CF custom error pages

### LINKED ISSUE #469: Errors: use application errors from CF and ALBs
Just noting here before I forget, I'd like to make sure our 5xx error pages are used from our various services, and likely need to configure bold CF and ALBs to use these error pages, or copy and hardcode these pages in the services.

- [ ] Cloudflare use 502/504 unavailable error pages
- [ ] Cloudflare general 5xx response error page?
- [ ] ALB 502/504 unavailable error pages? Is this necessary or does CF cover this?

We shouldn't show nginx or the default ALB error pages with nicer error pages.

- Refs #326

davidfischer:
- Excluded these from pre-commit as they intentionally have inline styles
- Punted for now on adapting a subset of our styles for inlining
- Partially resolves https://github.com/readthedocs/ext-theme/issues/469 (the CF part but not the ALB part)


readthedocsext/theme/static/readthedocsext/theme/error-pages/managed_challenge.html
```diff
@@ -0,0 +1,189 @@
...
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>403 - Request Challenge</title>
+    <style>
+    :root {
+      --bg-color: #f4f4f5;
+      --text-color: #18181b;
+      --box-bg: #ffffff;
+      --border-color: #e4e4e7;
+      --accent-color: #dc2626; /* Red for blocking */
```
agjohnson: We should match the already used site colors here, especially for the accents though.

readthedocsext/theme/static/readthedocsext/theme/error-pages/managed_challenge.html
agjohnson, line 1: We already have the template path `readthedocsext/theme/errors/`, let's use `errors` in the path instead of `error-pages` to reduce confusion between the two.

agjohnson:
> Punted for now on adapting a subset of our styles for inlining

Do we need to use a subset? We hit this pretty quickly the last time we tried this but maybe the limit was lower.

> When you provide a URL for a custom error asset, Cloudflare fetches the page and inlines all referenced resources into the HTML. Images and other binary resources (including those referenced from CSS via url(...)) are inlined as base64-encoded data URLs. CSS and JavaScript files are inlined as plain text inside <style> and <script> tags. The processed page must not exceed approximately 1.5 MB.

I'm guessing what is/was happening is that our CSS is only 800kB on disk but then that tries to also inline the fonts/etc, exceeding 1.5MB.

These error messages should match our existing look and feel, especially on customer documentation projects. They're pretty one-off feeling right now. It's better than Cloudflare branding these error pages at least though.

There are some modifications that would make this fit better temporarily, but I also won't want to get too far away from finishing this work either way. We'll want to follow up with pages that use our existing styles and structure so these don't feel so out of place.

readthedocsext/theme/static/readthedocsext/theme/error-pages/managed_challenge.html
```diff
@@ -0,0 +1,189 @@
...
+      :root {
+        --bg-color: #09090b;
+        --text-color: #f4f4f5;
+        --box-bg: #18181b;
+        --border-color: #27272a;
+        --accent-color: #ef4444;
+        --light-svg-display: none;
+        --dark-svg-display: block;
+      }
+    }
```
agjohnson: We don't support dark themes on the current error pages and this is probably most of what feels so out of place currently. I might say drop dark mode from this pass.

readthedocsext/theme/static/readthedocsext/theme/error-pages/managed_challenge.html
```diff
@@ -0,0 +1,189 @@
...
+      margin-left: auto;
+      margin-right: auto;
+      margin-bottom: 2rem;
+    }
+    svg.light {
+      display: var(--light-svg-display);
+    }
+    svg.dark {
+      display: var(--dark-svg-display);
+    }
```
agjohnson: No need for multiple SVGs, we can use CSS to invert:

```suggestion
    svg.dark {
      filter: invert(100) brightness(110);
    }
```

That assumes the dark color text SVG for use on white background, inverted to be (brighter) white.

readthedocsext/theme/static/readthedocsext/theme/error-pages/managed_challenge.html
```diff
@@ -0,0 +1,189 @@
...
+    <div class="container">
+
+      <svg class="dark"
+           version="1.1"
+           id="svg"
+           xmlns="http://www.w3.org/2000/svg"
+           xmlns:xlink="http://www.w3.org/1999/xlink"
+           x="0px"
+           y="0px"
+           viewBox="694 197 2000 400">
```
agjohnson: I don't think we need to hardcode this. If you reference the hosted SVG, Cloudflare will inline it for us. Then we don't need to keep multiple hardcoded copies around.

https://app-assets.readthedocs.org/readthedocsext/theme/images/logo-wordmark-dark.svg

readthedocsext/theme/static/readthedocsext/theme/error-pages/managed_challenge.html
```diff
@@ -0,0 +1,189 @@
...
+        <path style="fill:#32322A" d="M2592.9,376.5c-4.3-1.6-9.6-2.4-15.7-2.4c-6.1,0-11.1,1.4-14.9,4.3c-3.6,2.7-5.4,6.1-5.4,10.3   c0,4,0.6,7.1,1.9,9.5c1.4,2.2,3.6,4.1,6.5,5.7c4.5,2.3,9.9,4.4,16.2,6.2c6.3,1.6,11,3,14.1,4.1c3.1,0.9,6.8,2.5,11.4,4.9   c4.7,2.3,8.2,4.9,10.5,7.6c6.3,6.7,9.5,15.2,9.5,25.7c0,13.5-5,24.1-14.9,31.9c-9.7,7.6-22.2,11.4-37.3,11.4   c-22,0-38.6-2.8-49.7-8.4v-37.6l24.3-1.9v13c0,7.9,7.6,11.9,22.7,11.9s22.7-5.5,22.7-16.5c0-4-1.4-7.2-4.1-9.7   c-2.5-2.5-5-4.2-7.6-5.1c-2.5-0.9-5.6-1.8-9.2-2.7c-3.4-0.9-6.8-1.8-10.3-2.7c-3.2-0.9-6.8-2.1-10.8-3.5c-3.8-1.6-8-3.9-12.7-6.8   c-9.2-5.9-13.8-15.9-13.8-29.7c0-14.1,5-24.9,14.9-32.4c9.9-7.6,22.3-11.4,37.3-11.4c15.1,0,30.1,3.6,44.9,10.8v32.4l-24.3,1.9   v-11.4C2599.1,381.2,2597,378.1,2592.9,376.5" />
+        </g>
+      </svg>
+
+      <h1>403 - Request Challenge</h1>
```
agjohnson: We use sentence case for headings

```suggestion
      <h1>403 - Request challenge</h1>
```

agjohnson (commented):
Assuming reusing our styles is a harder task, this looks okay temporarily. I commented on the first file, but most of that applies to all the files. A second pass at these can explore trimming our existing CSS down and perhaps consider a dark theme, though I wouldn't support that unless we decide to support our dark theme without user authentication.

readthedocsext/theme/static/readthedocsext/theme/error-pages/managed_challenge.html
davidfischer: There's a lot of advantages to having the page served by others (Cloudflare) being a single self-contained page. We *can* rely on Cloudflare to handle serving other content correctly, but it'd be a lot simpler and a lot harder to get wrong if the error page was entirely self-contained.

davidfischer:
> Do we need to use a subset? We hit this pretty quickly the last time we tried this but maybe the limit was lower.

We don't strictly need to but there's a lot of nice aspects of having the entire error page being self-contained. We can rely on Cloudflare to "just handle it" but it'd be a lot harder to get wrong and a lot more portable if we had self-contained error pages.

I can attempt to match the styles/colors a little closer and possibly inline a small subset, but unless you are dead-set on it, I'd advise against using our full styles. If we do go that route, I'd probably just make this a template that uses other styles rather than a static file collected with `collectstatic`.

readthedocsext/theme/static/readthedocsext/theme/error-pages/managed_challenge.html
agjohnson: It would still be self-contained, Cloudflare automatically does all of the inlining instead of referencing a URL[^1]:

> When you provide a URL for a custom error asset, Cloudflare fetches the page and inlines all referenced resources into the HTML. Images and other binary resources (including those referenced from CSS via url(...)) are inlined as base64-encoded data URLs.

[^1]: https://developers.cloudflare.com/rules/custom-errors/#custom-error-assets

davidfischer: No I get that. I just think there are additional advantages to having it actually be self-contained by design rather than relying on Cloudflare to magically make it self-contained.

agjohnson:
Well, I'd say just the changes above are enough for a first pass -- skipping dark mode and using colors from our styles.

Long term, these error pages should definitely match our dashboard errors but for a first pass the basic style is okay. They just should stick to a look and feel that isn't too far removed from our current, it's a bit confusing when we give varying UX/styles.

When we polish these up to match our current styles, I wouldn't use the full stylesheet here due to the size but I would use just the rules we need to make it look like the other error pages.

And yeah, the idea was to make this a template long term too. But similarly, we can keep this simple by just leaving them as static files to start. Eventually we'd author them alongside all of our other error templates and feed them to Cloudflare through [our existing ErrorView](https://github.com/readthedocs/readthedocs.org/blob/496c525faa7b65824d9d2faa4c72bd8566e08323/readthedocs/core/views/__init__.py#L93). In this world, our templates handle the static files so we wouldn't have hardcoded inlines for images/etc.

readthedocsext/theme/static/readthedocsext/theme/error-pages/managed_challenge.html
agjohnson: Yeah I get that it's more explicit, but if we aren't hitting any issues with Cloudflare inlining of these files then it seems fine to lean on Cloudflare and keep our maintenance simple. The long term goal would be to feed Cloudflare these pages through our templates, so would end up relying directly on static file collection.

davidfischer:
Ok, I updated the styles on these pages to match the dashboard styles. They're still static, self-contained HTML rather than templates, but this should give us a first pass.

PR #767, opened 2026-07-09
Fix remaining dark theme issues and abstract dark theme creation

### LINKED ISSUE #747: Dark mode display issues
Some new issues and leftovers from

- #727 

----

- [x] #748 
- [x] Input boxes have no inverted variant. The top search box is an input and dropdown search, all not inverted.
- [x] Inverted segments used in a few places like destructive actions could be less prominent
- [x] QR code image from Allauth isn't inverted

agjohnson:
This solves a large number of display issues with the dark theme and
moves the plan towards breaking the dark theme creation out of this
repository.

Issues addressed with this:

- Input fields have dark styles
- Basic search module element dark styles
- Many minor CSS rules and pseudo selectors are now supported in dark
  theme transform. Much wider coverage of all the elements missed on manual
  pass.
- Secondary and primary color style issues. This also affected light
  theme but is more noticable on the dark theme where some different
  colors were used, like bright teal from the base FUI colors.
- Fixes for segment styles, including some of the danger zone colored segments
  we were using as `ui inverted red segment`
- Lots of little polish like dropdown and search hover colors, this was
  missing because of the `:extends()` pattern used previously.
- QR code background now

The approach here is:

- I broke out the transform step to a general PostCSS library:
  - https://github.com/readthedocs/postcss-fomanticui-dark and now also a package
- Webpack was reconfigured to output the two files using the plugin
  instead of a separate file.
- FUI fixes on top of the base styles are a mix:
  - Mostly changing variables used by the elements
  - Some instances of explicit one-off dark theme rules that load when
    the theme is active
  - The FUI search module is the largest one-off chunk, it was just a
    simple find and copy to create an inverted style. FUI doesn't have
    this currently. It's not a common element for us either way but
    we've talked about improving it.

- Fixes #747

----

image

image

image

image

image

image

image


package.json
```diff
@@ -58,12 +58,17 @@
     "lit": "^3.1.2",
     "matchdep": "~2.0.0",
     "mini-css-extract-plugin": "^2.7.6",
+    "path": "^0.12.7",
     "plausible-tracker": "^0.3.8",
+    "postcss": "^8.5.15",
+    "postcss-fomanticui-dark": "github:agjohnson/postcss-fomanticui-dark#4140232",
```
agjohnson: Note, this is just pointing to GH for now but should be a full release instead at least. This could be moved to readthedocs org too, but I might also see what interest there is to keep this going from FUI folks too.

humitos:
I tested this locally and it worked fine. Here is the list of things I found could be polished a little more. Nothing blocking, tho:

|Image|Description|
|----|----|
|Screenshot_2026-08-03_12-11-19|The hidden tag/label contrast isn't too high and it's hard to read it|
|Screenshot_2026-08-03_12-13-15|Same here for these labels|
|Screenshot_2026-08-03_12-11-59|There is no visual separation between the popup and the `use-apt` label from another row|
|Screenshot_2026-08-03_12-12-31|Hovering this buttons makes the icons to disappear|
|Screenshot_2026-08-03_12-14-42|Confirmation dialog is "too white/shiny" for a dark mode. I wasn't expected so much light after clicking on "Delete"|


Note I had to rebuild the webpack Docker image so the new postcss dependency was installed.

agjohnson:
All of these were just the base inverted styles. I added a few tweaks to these in 4795cbb. None of these issues were really required, but I think this should settle any changes here for a while.

image

image

image

image

humitos (approved):

humitos:
Looks good to me. I think we can just deploy this as-is and tweak anything we may find in the future.

Nit: the "Add a connection" modal screenshot you shared hides the GitLab icon since you have that option selected.

PR #768, opened 2026-07-29
organizations: fix empty team list for SSO organizations

ericholscher:
For organizations with SSO enabled, the Teams page always rendered an empty list: the SSO branch of `organizations/team_list.html` included its list partial without passing `objects`, so the shared table template had nothing to iterate. The non-SSO branch — and the sibling SSO includes for members and projects — all pass `objects` correctly; this was found in passing while auditing dashboard pagination for readthedocs/readthedocs.org#13210.

The touched file also carried four pre-existing template-lint findings (wildcard `{% load %}`s and unnamed `{% endblock %}`s) that CI's changed-files pre-commit run would now flag, so they are fixed here as well.

---

https://claude.ai/code/session_01BB6HHBjvZJsSgYC1aL1N2p

---
_Generated by [Claude Code](https://claude.ai/code/session_01BB6HHBjvZJsSgYC1aL1N2p)_

### DIFF
--- a/readthedocsext/theme/templates/organizations/team_list.html
+++ b/readthedocsext/theme/templates/organizations/team_list.html
@@ -1,18 +1,18 @@
 {% extends "organizations/base.html" %}
 
-{% load i18n %}
+{% load trans from i18n %}
 
-{% load organizations %}
+{% load has_sso_enabled from organizations %}
 
-{% block title %}{{ organization.name }} - {% trans "Teams" %}{% endblock %}
+{% block title %}{{ organization.name }} - {% trans "Teams" %}{% endblock title %}
 
 {% block content-header %}
   {% include "organizations/partials/header.html" with organization=organization teams_active="active" %}
-{% endblock %}
+{% endblock content-header %}
 
 {% block content %}
   {% if organization|has_sso_enabled:"allauth" %}
-    {% include "organizations/partials/has_sso/team_list.html" %}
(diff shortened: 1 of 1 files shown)


agjohnson:
> the SSO branch of organizations/team_list.html included its list partial without passing objects, so the shared table template had nothing to iterate

This seems to be missing the intention of the included template. The template isn't meant to list teams, this view has always intentionally disabled the team listing specifically:

https://github.com/readthedocs/ext-theme/blob/05c10e8ac6c248e5ab4ac7623da153911a2f1507/readthedocsext/theme/templates/organizations/partials/has_sso/team_list.html#L21-L34

We note the same in our docs:

https://docs.readthedocs.com/platform/stable/commercial/single-sign-on.html#:~:text=Once,support

ericholscher:
We should probably rename the template then to make it more explicit?

humitos:
I'm not sure I follow. What name would you suggest?

ericholscher:
`no_team_warning.html`? :) 

Not really worth fixing tbh -- but definitely looks like a bug at first peek.

humitos:
Ah, you mean we can use a completely different file for organizations with SSO enabled instead of reusing the same file with a special case? That could work

agjohnson:
I prefer to always stick to django naming conventions for template naming, including partials and includes, so just sticking to model/verb naming as much as possible. The agent should be inspecting and understanding the included file anyway, which it doesn't seem it did. More commenting maybe helps here, not sure.

PR #771, opened 2026-08-18
projects: show a browsable repository list on the add project page

ericholscher:
The repository picker only renders results in response to a search, so a fully synced account looks exactly like an empty one until the user types something — and many users never do. This loads the first page of the user's repositories on view load and shows them as a selectable list, with search covering the rest and the previous placeholder kept as the loading/fallback state.

Now based directly on `main` and mergeable on its own: it needs no backend changes — the list comes from the existing APIv3 remote repositories endpoint. Two optional companions, in either order: readthedocs/readthedocs.org#13245 orders the endpoint so the first page is the most likely import target (without it the list is alphabetical), and #770 (stacked on this PR) adds empty states for accounts where the list can't have results.

---

https://claude.ai/code/session_019yjYstmBb2HiscZUc5eKH4


humitos:
Can you post a few screenshots here? I like the idea, but I'm not sure to understand how it looks like.

> This loads the first page of the user's repositories on view load and shows them as a selectable list, with search covering the rest and the previous placeholder kept as the loading/fallback state.

Is is possible to use the "latest modified repositories" instead just the first page? I think it will be more likely to succeed in that case.

ericholscher:
@humitos Basiclaly something like this. It's really silly that we make users search before adding anything, it's quite a big step for onboarding to not list something. 

image-1789975617878


> Is is possible to use the "latest modified repositories" instead just the first page? I think it will be more likely to succeed in that case.

Yea, I think even doing something smarter and filtering by things with `docs` in them, or similar. We should be making it as easy as possible for someone to import the right repo with as few clicks as possible. It's a huge miss in our current onboarding. 

https://github.com/readthedocs/readthedocs.org/pull/13245

humitos:
What happened here? This got merged while we were talking about it 🤔

ericholscher:
It's merged into another branch. But I ended up moving the discussion to a design doc instead.

PR #774, opened 2026-09-15
socialaccount: trigger modal with dropdown instead of primary action

ericholscher:
On Community we want the GitHub App to be the default way to log in, sign up, and connect, without asking every user to choose between it and the GitHub OAuth app first. The OAuth app still needs to exist for people who haven't migrated, just somewhere less prominent. This is the "hide the popup on Community" part of readthedocs/meta#201 (see [this comment](https://github.com/readthedocs/meta/issues/201#issuecomment-5666585642)).

The modal already keeps both GitHub providers out of the list below it, so the only thing that needed to change is what the modal hangs off. The recommended provider, the same one the modal gives its primary button to, now gets a button of its own, with a second button beside it that opens the modal. Nothing is forced on arrival, and the choice is still one click away with a description of each option. Deployments opt in with a `default` app setting, so nothing changes where it isn't set, including Business.

This went through two other shapes first, and the reasons for dropping them are worth knowing:

- A "More options" accordion below the provider list. Too disconnected from the GitHub button to be discoverable, per review.
- A dropdown menu hanging off the button, closer to the CircleCI reference. It looked right, but SemanticUI's dropdown can't activate its items from the keyboard. Selecting an item with the arrow keys and pressing Enter does nothing, and that holds for `a` items, `button` items, and a custom `action` handler, so it isn't something this PR could fix without writing our own key handling. A menu would have made the OAuth app reachable by mouse only, on the one page where people log in. It also needed form submit buttons styled to look like menu items, which meant a new theme rule fighting the browser's native button chrome.

Reusing the modal avoids that: tab to the second button, Enter, tab, Enter submits. It also costs far less code, since the provider list, the provider tags and the built assets are all untouched.

The two "Last used" popups sit on opposite sides of the split button, so someone who last logged in with the OAuth app is pointed at the second button rather than at the GitHub App.

The settings side is readthedocs/readthedocs.org#13306. This should land first, since the setting is inert without it.

https://claude.ai/code/session_01WTGoz49zyDE7SwdqHqt2PK


ericholscher:
Guessing the main UI question here is how we actually hide it (cc @agjohnson).

stsewd:
I was thinking of something like circle

image

but not sure how easy is that to implement

agjohnson:
Yeah, I don't love the disconnected accordion. Circleci at least hints that there are more than one Github login options. Users won't know to look in the accordion without a better hint.

This would involve updating the button that triggers the modal in the provider list snippet but is possible because we are already customizing the GitHub button there

ericholscher:
New approach is a dropdown button that pulls up the modal. I think that's clean since we already have the modal, so minimal code churn. 

I might argue we should ship that on both sites? Make it obvious what the default should be.

ericholscher:
Example:

Screenshot 2026-09-22 at 4 31 54 PM

then it just opened the modal on the dropdown.

PR #776, opened 2026-09-16
projects: stop posting the default branch into the import form

ericholscher:
The repository picker posted the selected repository's default branch into the "Add project" form, which froze a copy of it on the project. Read the Docs already resolves the default branch from the connected repository, so the copy only went stale when someone renamed their branch.

Companion to readthedocs/readthedocs.org#13314, which removes the field this input was filling. That one should merge first — until it does, the extra POST key is simply ignored, so the two can land in either order without breaking the form.

https://claude.ai/code/session_01CT1CbriDPnfk2tz6FwUMm3

---
_Generated by [Claude Code](https://claude.ai/code/session_01CT1CbriDPnfk2tz6FwUMm3)_


humitos:
I think the question here is _Why do we had this field in the first place if it's not needed?_ and _What happens when the repository is added manually?_

stsewd (approved):

stsewd:
> I think the question here is _Why do we had this field in the first place if it's not needed?_ and _What happens when the repository is added manually?_

Our code used to require it for everything, but now when an empty value is given, we default to the value from the remote repository, or from the default branch when a repo is first cloned.

ericholscher:
Yea, the description here breaks down how we're currently replacing the usage: https://github.com/readthedocs/readthedocs.org/pull/13314

Issue #286, opened 2024-03-01
Search: project/version create search requires typing to load initial list


agjohnson:
This is probably a small fix on the semanticui search field configuration. These fields are initialized with some custom settings, there is likely a setting to handle an initial, empty search request.

This is unclear UX as when the field gets focus, it doesn't give any hints at _what you type_. If there were repo names in the search list, this would be more obvious.

agjohnson:
This is no longer an issue with the move to standard filters, all the options are prepopulated in the search menu. Users will get an unfiltered list on click of these filters.

PR #777, opened 2026-09-21
Projects: show search results on focus for add project and add version

ericholscher:
The repository search on the add project page stayed empty until you typed something, so it gave no hint of what was available. It now loads the first 10 repositories on focus, like the version search already does (#286).

The version search now lists only inactive versions, since active ones are already on the versions page and can't be activated again.

The `semanticui` binding change is needed because `autofocus` focuses the input before Knockout sets up the search, so search on focus never ran on page load. The add project search also starts out `disabled`, and it drops queries until its `css` binding removes that.

Loading on focus counts as a search attempt, so the "Not finding a repository?" popup now shows one keystroke earlier.

https://claude.ai/code/session_01Wrp1gcYfhYAUgRH4VXPVpp

---
_Generated by [Claude Code](https://claude.ai/code/session_01Wrp1gcYfhYAUgRH4VXPVpp)_


src/js/application/plugins.js
```diff
@@ -450,6 +450,14 @@ export const semanticui = {
...
           );
           jq_element[key](value);
+
+          // ``autofocus`` fires before the module exists. Deferred as the module
+          // drops queries until a ``css`` binding removes ``disabled``.
+          if (key === "search" && element.contains(document.activeElement)) {
+            queueMicrotask(() => {
+              jq_element.search("event focus");
+            });
+          }
```
humitos, line 460: I'm not sure to understand this correctly. Why `event focus` will be available when the microtask is executed? Shouldn't we based this on a particular event instead of just queueing it?

humitos (commented):
I tested it locally and it worked fine:

https://github.com/user-attachments/assets/cb2abe9c-8f2f-4369-ae5f-b8d84a660239

I'm not 100% sure about the JS implementation, but it looks simple enough.

As a future reference, it would be a lot easier to understand the PR and review it if you attach a GIF or a screenshot showing what behavior is changing and how it looks like. It's hard to get that from the description of the PR most of the time when it's a UX/UI change.

src/js/application/plugins.js
ericholscher: This is basically just firing the signal on page load, and it's placed after the loading of the other JS, so it's done when it's ready.

agjohnson: Is there a strong reason why you used microtasks here? This is a pretty specialized part of JS code timing and one that interferes with async, evented, and promise code patterns.

Execution order may not be predictable and we already have patterns for triggering behaviors on elements in other UIs that we should follow.

However overall I'd say we should skip autofocusing without user interaction entirely, it simplifies and makes for fewer edge cases.

agjohnson:
This feels like there are going to be negative edge cases from using autofocus. What does this look like if the user doesn't have any synced repos initially -- is it an error in the search drop down about no results? I think requiring user interaction for focusing is best.

PR #780, opened 2026-09-22
Projects: list repositories on the add project page

ericholscher:
> **Experimental.** This is a first cut of one option from the design doc in readthedocs/readthedocs.org#13324 and is not ready for human review until we decide on that doc.

The "Add project" page now renders the user's repositories as a standard CRUD list, the same shape as the versions, builds, and domains pages, instead of a search dropdown that only showed results while focused.

### Before

The page was a bare search input above a placeholder that pointed back at it. Results only existed while the dropdown was open.

Before: empty search page

Before: search dropdown open

### After

Repositories are listed on load with a name filter and an organization dropdown. Each row shows the repository name with its owner underneath, icon-only labels for state (lock for private, hand for "needs manual configuration", check for "already imported", each with a popup), and an arrow to continue. The arrow posts the same hidden fields the old selected-repository card did, so the import wizard is unchanged.

After: repository list

Filtering by name:

After: filtered by name

The refresh and repair actions only appear when the list is empty. When the filters match nothing, the placeholder replaces the popup that used to appear after failed searches:

After: no matching repositories

When the user has no repositories at all, the placeholder also links to connected accounts. Repair opens the existing modal with the GitHub App and resync guidance.

After: no repositories

### Notes for whenever this is reviewed

- The filter form include gained support for free text filters, rendered as a small icon input that submits on enter. Every existing filter was a dropdown.
- A `before_last` template filter splits the owner from the repository's full name, since user-owned repositories have no organization object.
- The search JS and the Knockout result template are removed rather than adapted, so this conflicts with #777 in `create.js`. Whichever merges second needs a trivial rebase.
- The manual import path is still the sidebar tab, and URL paste is not in this PR.
- Screenshots were taken locally with Font Awesome Free standing in for the Pro kit, so a couple of icons differ slightly from production.

Requires the backend change in [readthedocs/readthedocs.org#13331](https://github.com/readthedocs/readthedocs.org/pull/13331) for the `remote_repositories` and `filter` context.

https://claude.ai/code/session_01Wrp1gcYfhYAUgRH4VXPVpp


humitos:
Small quick notes:
- Overall, I like it. I think we are going back to what we had in the old dashboard, tho 🙃 -- but the main issue there was there was no way to filter repositories. I think the "presentation" of the data was correct, tho.
- The screenshots already show an "horizontal space issue". There is no too much space for all the information we want to display here and we get the `...` (epsilon) immediately. If the name of your organization is a little more longer than a few characters, all the repositories will end up _being named equal_: full name + tags + Continue button, seems like too much for this presentation. I think we could find a way to make this work, somehow.
- We can remove the "Not finding a repository?" message when the empty list appears.
- The initial list should show some copy like "Suggested repositories" or similar before the list. Otherwise, it's hard understand if these are all the available repositories or what I'm seeing this unsorted/weird list. Basically, we need to communicate that the input filter will be applied over _more repositories_ than the ones I'm currently seeing here.
- Do we plan to have pages here? I guess "No" is the answer and in case the repository is not listed there by default, the user _has to_ use the filter/search input and can't click on "Next page" or similar.

humitos:
The new iteration looks a lot better to me:

after-list


The only information we would be missing here is _knowing where/how many project from this particular repository are already imported_. The plus item in this screenshot:

Screenshot_2026-09-23_13-15-21

- Automatically configured, in the new proposal, is whether the continue arrow is enabled/disabled 👍🏼 
- Repository is public, I understand is defined by the lock in the proposal?

Now, regarding to the "already configured repositories". I understand this is the check label (✔️) which makes sense. However, it's not telling us _which ones are those repositories_ and link to them. I don't think it's _super relevant_ information, but worth to show it somehow. We could either show them when hovering the icon, or in the next page when you click on the arrow to continue if we want it to be bigger with links and more. Not a blocker, tho, I'd say.

It still needs copy or something to communicate the list shown is not complete and there are more repositories you can find by start typing the name. The "Not finding a repository?" box at the bottom will guide me incorrectly in that case. I just need to type the name, there is nothing to refresh or repair.

Oh, there is something else, not super important, tho --but we are not communicating _where_ the repository is hosted. Is it on GitHub, GitLab, etc? Very unlikely, but the user could have the same named repository on both platforms. I'm not concern about this use case, tho.

